PR #36, opened 2014-12-11
Replace Map<String,String[]> with serializable Metadata

jnioche:
Having a wrapper for Metadata would be a bit more elegant than moving the Map<String,String[]> around.
This initial implementation adds Metadata and removes the KeyValues utility. Later on we can replace all occurrences of Map<String,String[]> and systematically use Metadata instead.


jakekdodd:
Now that we've released 0.4, should we make this priority 1? I think several other issues are affected by the new metadata scheme

GuiForget:
Yes yes yes!!! :dancer: 

My only concern is should we just go further and replace the use of Map<String, String[]> with Metadata class right away?

jnioche:
+1 to make it top priority. I will take care of this.

> replace the use of Map with Metadata class right away

that's what the old PR was attempting to do, might as well do it all in one go. 

Can't output it from a bolt and not get the following bolt to handle them, can't we?

jnioche:
Opened a new PR #78

Issue #38, opened 2014-12-11
Sitemap parser


jnioche:
It would be good to reuse the parsers from [crawler-commons](https://code.google.com/p/crawler-commons/) to handle sitemaps. Probably simpler to have a separate parser from the Tika-based one and trigger the use of the Sitemap parser based on the presence of an arbitrary metadata like 'parse.sitemap'=true.

The triaging of the tuples based on the metadata could be done in a meta parser wrapping both the tika or sitemap one but probably simpler if that could be done somehow in the topology class. Alternatively the sitemap parser could also wrap the tika parser. 

The output of this parser would not be the same as the Tika based one as it would not generate any textual content or call the parsefilters but consists of URL,Metadata pairs.  Similarly to what I suggested for #37, this bolt could generate a 'status' stream that would be constumed by a persistence bolt. It would also use the default stream to pass on URLs that are not sitemaps.

Question : what happens if we send things down a stream with nothing to consume them?

jnioche:
Implemented in #42

PR #42, opened 2014-12-18
Sitemaps parser

jnioche:
See #38 

This adds a SiteMapParserBolt and a test class. It also reorganizes the test resources and makes the ParsingTester more generic.

As discussed, this Bolt uses a non-default stream 'status' to output the newly discovered URLs and the default stream for documents that are not marked as being sitemaps.

Could you please review this? Thanks!

### DIFF
--- a/src/main/java/com/digitalpebble/storm/crawler/bolt/SiteMapParserBolt.java
+++ b/src/main/java/com/digitalpebble/storm/crawler/bolt/SiteMapParserBolt.java
@@ -36,6 +36,8 @@ import backtype.storm.tuple.Fields;
 import backtype.storm.tuple.Tuple;
 import backtype.storm.tuple.Values;
 
+import com.digitalpebble.storm.crawler.Constants;
+import com.digitalpebble.storm.crawler.persistence.Status;
 import com.digitalpebble.storm.crawler.protocol.HttpHeaders;
 import com.digitalpebble.storm.crawler.util.KeyValues;
 
@@ -55,7 +57,6 @@ public class SiteMapParserBolt extends BaseRichBolt {
 
     private OutputCollector collector;
     public static final String isSitemapKey = "isSitemap";
-    public static final String statusStreamName = "status";
 
     private boolean strictMode = false;
 
@@ -66,6 +67,7 @@ public class SiteMapParserBolt extends BaseRichBolt {
     public void execute(Tuple tuple) {
         HashMap<String, String[]> metadata = (HashMap<String, String[]>) tuple
                 .getValueByField("metadata");
+
         // TODO check that we have the right number of fields ?
         String isBoolean = KeyValues.getValue(isSitemapKey, metadata);
         if (StringUtils.isBlank(isBoolean)
@@ -75,6 +77,7 @@ public class SiteMapParserBolt extends BaseRichBolt {
             this.collector.ack(tuple);
             return;
         }
+
         // it does have the right key/value
         byte[] content = tuple.getBinaryByField("content");
         String url = tuple.getStringByField("url");
@@ -89,8 +92,13 @@ public class SiteMapParserBolt extends BaseRichBolt {
 
         // send to status stream
         for (Values ol : outlinks) {
-            collector.emit(statusStreamName, new Values(ol));
+            collector.emit(Constants.StatusStreamName, new Values(ol));
         }
+
+        // marking the main URL as successfully fetched
+        collector.emit(Constants.StatusStreamName, new Values(url, metadata,
+                Status.FETCHED));
+
         this.collector.ack(tuple);
     }
 
@@ -122,7 +130,7 @@ public class SiteMapParserBolt extends BaseRichBolt {
                 // TODO configure which metadata gets inherited from parent
                 HashMap<String, String[]> metadata = KeyValues.newInstance();
                 KeyValues.setValue(isSitemapKey, metadata, "true");
-                Values ol = new Values(s, metadata);
+                Values ol = new Values(s, metadata, Status.DISCOVERED);
                 links.add(ol);
                 LOG.debug(url + " : [sitemap] " + s);
             }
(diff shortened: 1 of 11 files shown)


jnioche:
Have rebased with changes from master (status stream). Now generates an update about the status of the main URL 

@DigitalPebble/committers-crawler any objections to committing this?

GuiForget:
Can't be merged yet

jnioche:
Care to elaborate a bit ? ;-)

GuiForget:
There are conflicts preventing this to be merged

jnioche:
I see. I probably did something wrong with the rebase. Will have a look on Wednesday

jnioche:
Have merged the problematic file. @GuiForget could you please have another look?

src/main/java/com/digitalpebble/storm/crawler/bolt/SiteMapParserBolt.java
```diff
@@ -0,0 +1,178 @@
...
+    private static final org.slf4j.Logger LOG = LoggerFactory
+            .getLogger(SiteMapParserBolt.class);
+
+    @Override
+    public void execute(Tuple tuple) {
+        HashMap<String, String[]> metadata = (HashMap<String, String[]>) tuple
+                .getValueByField("metadata");
+
+        // TODO check that we have the right number of fields ?
+        String isBoolean = KeyValues.getValue(isSitemapKey, metadata);
```
GuiForget: Did you mean to name it isBoolean?

src/main/java/com/digitalpebble/storm/crawler/bolt/SiteMapParserBolt.java
```diff
@@ -0,0 +1,178 @@
...
+        if (StringUtils.isBlank(isBoolean)
+                || !isBoolean.equalsIgnoreCase("true")) {
```
GuiForget: Looks like this could simply be `if (!Boolean.valueOf(isBoolean))`

src/main/java/com/digitalpebble/storm/crawler/bolt/SiteMapParserBolt.java
```diff
@@ -0,0 +1,178 @@
...
+        if (outlinks == null) {
+            // likely to happen ad lib - just ack
+            // error has been logged
+            this.collector.ack(tuple);
+            return;
+        }
+
+        // send to status stream
+        for (Values ol : outlinks) {
+            collector.emit(Constants.StatusStreamName, new Values(ol));
```
GuiForget: Isn't this odd? `new Values(ol)` doesn't create a copy of the Values. Shouldn't it be `collector.emit(Constants.StatusStreamName, ol);`

src/main/java/com/digitalpebble/storm/crawler/bolt/SiteMapParserBolt.java
```diff
@@ -0,0 +1,178 @@
...
+                // TODO convert the frequency into a numerical value and handle
+                // it in metadata
+                // TODO configure which metadata gets inherited from parent
+                String s = smurl.getUrl().toExternalForm();
+                // TODO apply filtering to outlinks
+                HashMap<String, String[]> metadata = KeyValues.newInstance();
+                KeyValues.setValue(isSitemapKey, metadata, "false");
+                Values ol = new Values(s, metadata, Status.DISCOVERED);
+                links.add(ol);
+                LOG.debug(url + " : [sitemap] " + s);
```
GuiForget: You should use `{}` place holders

src/main/java/com/digitalpebble/storm/crawler/bolt/SiteMapParserBolt.java
```diff
@@ -0,0 +1,178 @@
...
+
+        crawlercommons.sitemaps.SiteMapParser parser = new crawlercommons.sitemaps.SiteMapParser(
+                strictMode);
+
+        AbstractSiteMap siteMap = null;
+        try {
+            siteMap = parser.parseSiteMap(contentType, content, new URL(url));
+        } catch (Exception e) {
+            LOG.error("Exception while parsing sitemap", e);
+            return null;
```
GuiForget: Why not return an empty list. That way no need to check for null later

src/main/java/com/digitalpebble/storm/crawler/bolt/SiteMapParserBolt.java
jnioche: another variable with existential issues ;-) fixed

src/main/java/com/digitalpebble/storm/crawler/bolt/SiteMapParserBolt.java
jnioche: definitely more elegant. fixed

src/main/java/com/digitalpebble/storm/crawler/bolt/SiteMapParserBolt.java
jnioche: fixed

src/main/java/com/digitalpebble/storm/crawler/bolt/SiteMapParserBolt.java
jnioche: good idea. I discovered this only recently and keep forgetting to use it. I will open a new issue to do the same everywhere else in the project

src/main/java/com/digitalpebble/storm/crawler/bolt/SiteMapParserBolt.java
jnioche: even more so now that we also send a tuple on the status stream for the main URL. Fixed

jnioche:
thanks for the comments and discussion @DigitalPebble/committers-crawler

Issue #43, opened 2014-12-19
Parser to send tuple to 'status' stream instead of failing


jnioche:
The parser currently fails tuples when an exception is caught. This is probably not the right behaviour as the same URL will be refetched and refail again later (unless of course). What we could do instead would be to use the `status` stream to mark the URL as failed and keep track of the reason why it did so. Whichever component is in charge of persisting the URL status can then decide on what to do with it.

This is related to the discussion in #42 

The same logic could be applied to fetch failures as well. Instead of failing them and let the spout handle the logic of keeping track of the number of errors we'd send to the status stream. The advantage of doing this is that the spout wouldn't have to update anything and would just read.

jnioche:
Implemented in #44

Issue #37, opened 2014-12-11
Handling of redirections


jnioche:
The FetcherBolt nor the ParserBolt currently do anything specific with the HTTP status returned for a given page and as a result redirections are currently not handled.

One option would be to let whichever component is in charge of the persistence of the URL status i.e. post parsing. Another approach would be to have the parser (and maybe also the fetcher) declare a non-default output (e.g. 'status') and plug that as input to the persistence bolt. 

any thoughts on this?

jnioche:
Part of it is implemented in #44. Still need to implement case where redirected URLs are sent straight into the internal queues

jnioche:
Will close for now. Redirections can be handled within the protocol implementations if necessary. Better just to treat them with the default behaviour and go through the status stream anyway as the targets might already be known. Keeps the code simpler as well.

PR #44, opened 2014-12-19
Fetcher and Parser use the status stream to report errors instead of fail

jnioche:
also used by fetcher to report redirections
related to issues  #37 and #38 #42 
implements #43

### DIFF
--- a/src/main/java/com/digitalpebble/storm/crawler/persistence/Status.java
+++ b/src/main/java/com/digitalpebble/storm/crawler/persistence/Status.java
@@ -18,15 +18,15 @@
 package com.digitalpebble.storm.crawler.persistence;
 
 public enum Status {
-    DISCOVERED, FETCHED, FETCH_ERROR, REDIR, ERROR;
+    DISCOVERED, FETCHED, FETCH_ERROR, REDIRECTION, ERROR;
 
-    /** Maps the HTTP Code to FETCHED, FETCH_ERROR or REDIR **/
+    /** Maps the HTTP Code to FETCHED, FETCH_ERROR or REDIRECTION **/
     public static Status fromHTTPCode(int code) {
         if (code == 200)
             return Status.FETCHED;
         // REDIRS?
         if (code >= 300 && code <= 400)
-            return Status.REDIR;
+            return Status.REDIRECTION;
         // error otherwise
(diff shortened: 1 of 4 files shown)


jakekdodd:
No shouts from me, although I suggest that we update SimpleFetcherBolt with the new changes as well. If you want to commit this as-is, I can add the new features to SimpleFetcherBolt in another commit

src/main/java/com/digitalpebble/storm/crawler/persistence/Status.java
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.digitalpebble.storm.crawler.persistence;
+
+public enum Status {
+    DISCOVERED, FETCHED, FETCH_ERROR, REDIR, ERROR;
```
GuiForget: I would prefer a full name rather than REDIR. Also do you need to differentiate between ERROR and FETCH_ERROR. Maybe the tuple emitted could have the name of the component who emitted it instead if one needs to differentiate between a fetch error and a parsing error

jnioche: Will rename REDIR to REDIRECTION. As for ERROR vs FETCH_ERROR : the distinction is required so that the users can implement a different logic if they want to. Typically a FETCH ERROR could be due to a temporary problem with the network or the target server whereas ERROR would be used for any other case e.g. parser fails on a given document. With the former you want to reschedule the URL and give it another try later (up to a configurable number of tries), whereas with the latter you'd probably want to give it up straight away. That type of error could happen within the Fetcher as well and is not only for the parser.
Differentiating between the 2 gives users the option of implementing whichever logic the see fit.
Passing the name of the component would be an overkill IMHO + does not really address the fact that the Fetcher could generate both types of status.

jnioche:
@jakekdodd sure, let's port this to the SimpleFetcherBolt in a separate issue after I've merged this one

jnioche:
Thanks for your comments

Issue #24, opened 2014-11-14
RobotsURLFilter


jnioche:
The filtering of URLs based on the robots.txt directives is done within the Fetcher for an incoming URL. It would be efficient to be able to filter on the outlinks so that URLs do not get added to the queues  (or any other form of persistence) if they can't be fetched anyway.

We should provide a RobotsURLFilter for this where we'd refetch the robots.txt file for a given URL and store it in a cache. The additional cost of pulling the robots.txt would be outweighed by the benefits of not adding unnecessary URLs to the queues.

jnioche:
Have implemented the simple version I originally had in mind. See #178
We can of course add a more sophisticated one based on the discussions we had here later on; for now this does exactly the job for small vertical crawls.

PR #46, opened 2014-12-19
Improved robots.txt handling

jakekdodd:
This PR adds a new mechanism for robots.txt caching and filtering.

Among the changes are:

- A new thread-safe, in-memory singleton cache for robots rules
- A RobotsURLFilter
- Moves the getCacheKey() method to the cache implementation because it's used in several locations
- The ParserBolt ensures that the robots rules are loaded into the shared cache, which means conditionally refetching and parsing the domain's robots.txt file

### DIFF
--- a/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
+++ b/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
@@ -27,6 +27,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import backtype.storm.Config;
+import com.digitalpebble.storm.crawler.protocol.Protocol;
+import com.digitalpebble.storm.crawler.protocol.ProtocolFactory;
 import org.apache.commons.lang.StringUtils;
 import org.apache.html.dom.HTMLDocumentImpl;
 import org.apache.tika.Tika;
@@ -89,8 +92,13 @@ public class ParserBolt extends BaseRichBolt {
     private boolean upperCaseElementNames = true;
     private Class HTMLMapperClass = IdentityHtmlMapper.class;
 
+    private ProtocolFactory protocolFactory;
+
     public void prepare(Map conf, TopologyContext context,
             OutputCollector collector) {
+
+        Config config = new Config();
+        config.putAll(conf);
 
         String urlconfigfile = ConfUtils.getString(conf,
                 "urlfilters.config.file", "urlfilters.json");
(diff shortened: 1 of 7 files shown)


jnioche:
It would make sense to pass the source URL as metadata anyway (great for debugging) and that could indeed be used to determine whether or not to do the sitemap checks. 

I am planning to add a utility class that would help generate metadata for outlinks given the metadata for the source page. This would be configurable of course and could be used to add the 'origin' metadata.

> The logic would call protocol.getRobotRules() for links with the same hostname as the source URL, and a direct cache query for all others.

I presume that 'direct cache query' means  hitting the same cache instance as used by the Fetcher, which in the case of the memory based one would work if it is located in the same JVM. If so, why not do that when the hostname is the same as the source as well? it should definitely be there.

jakekdodd:
> I presume that 'direct cache query' means hitting the same cache instance as used by the Fetcher, which in the case of the memory based one would work if it is located in the same JVM. If so, why not do that when the hostname is the same as the source as well? it should definitely be there.

The assumption is that many or most outlinks will be internal, so we would *guarantee* that the robots rules would be available for outlinks with the same hostname as the source URL; for external outlinks, the robots rules *might* be available for filtering, depending on a number of factors. 

To guarantee that robots rules for the hostname are in the JVM-local cache, we need to either fetch or re-parse the robots.txt file within the same process as the filters. For external outlinks, the rules *might* be present in the cache; if they are, then this is a secondary benefit.

Keeping the filter straightforward (give it a cache, feed it URLs to filter) leaves the cache-filling strategy outside of the filter, which is where I think it should be.

jnioche:
I agree with your definition of what the filter does (i.e. guaranteed for same hostname).

Most of the time the fetcher and parser will be running on the same JVM - if not the performance of the crawler would certainly suffer a lot. I'd be happy with a simpler solution where we expect this to be the case.

If we want a 100% guarantee then indeed we need to check the cache and if not found there, refetch and parse the robots.txt. What I think we disagree on is where it should be done. I don't think it should be done in the parser at all, its purpose is to be generic and moreover, people could have various implementations of parsing bolts (Gui wrote one for instance to deal with HTML exclusively). We'd want them to be able to reuse the parsingfilters as-is without having to write any bespoke code that would be a prerequisite for any of these filters to work. If we need to fetch+parse the robots rules then it should be done within the filter itself.

Again, there would be something wrong in the users' setup if the fetcher and parsers were not running in the same VM. I think it is reasonable to just reuse the cache filled by the Fetcher.

jakekdodd:
Let's close this PR, and revisit when #36 and #53 are closed. We'll keep the related issue (#24) open

PR #57, opened 2015-01-15
Pass in the Storm config when configuring ParseFilters

### LINKED ISSUE #56: Pass the storm config when configuring the ParseFilter
So they can access information shared via the yaml files

GuiForget:
A ParseFilter may want to access the Storm config during its config phase. To
that end the interface is changed to receive both the Storm config as well as
the Filter's parameters.

Also I changed ParseFilters to always call configure on the ParseFilter even
if there are no params defined. I think the original difference was too subtle
and therefore error prone. To make it easy on the ParseFiler, the paramsNode
is set to a NullNode.

Fixes #56


GuiForget:
After propagating all the changes to my own project, I gained confidence that this change is rock solid :)

PR #58, opened 2015-01-15
Fix XPathFilter to properly process XPathExpressions with element names

### LINKED ISSUE #31: Xpath expressions don't match when element name is set
As shown in the test [https://github.com/DigitalPebble/storm-crawler/commit/7cf89baec8297fc4e725d6d06be5f1145d6194c1], the XPathFilter does not extract the data when an element name is specified.

GuiForget:
When parsing the content, Tika creates a properly formatted XHTML document:
all elements are created within the namespace XHTML.

However in XPath 1.0, there's no concept of default namespace so
XPath expressions such as //BODY doesn't match anything. To make this
work we should use //ns1:BODY and define a NamespaceContext which associates
ns1 with "http://www.w3.org/1999/xhtml"

To keep the XPathExpressions simpler, I modified the DOMBuilder which
is our SaxHandler used to convert the SAX Events into a DOM tree to
ignore a "default name space" and the ParserBolt initializes it with
the XHTML namespace. This way //BODY matches.

Fixes #31

### DIFF
--- a/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
+++ b/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
@@ -38,6 +38,7 @@ import org.apache.tika.sax.BodyContentHandler;
 import org.apache.tika.sax.Link;
 import org.apache.tika.sax.LinkContentHandler;
 import org.apache.tika.sax.TeeContentHandler;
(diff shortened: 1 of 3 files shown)


jnioche:
Brilliant! Thanks for investigating and solving this @GuiForget 
Will probably push this back into Nutch as it has the same issue

PR #63, opened 2015-01-16
Add a method to create a mock tuple to TestUtil class

acinader:


jnioche:
looks good thanks Arthur

Issue #64, opened 2015-01-16
Refactoring the metrics in ParserBolt


jnioche:
See discussion in #41 : the metrics package will probably move to 'external' in which case we'll have to refactor ParserBolt so that it uses standard storm metrics instead.

jnioche:
implemented in #65

PR #65, opened 2015-01-16
ParserBolt uses right Status class + relies on standard metrics

jnioche:
Implements #64 
Could you please review?

### DIFF
--- a/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
+++ b/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
@@ -43,6 +43,7 @@ import org.slf4j.LoggerFactory;
 import org.w3c.dom.DocumentFragment;
 import org.xml.sax.ContentHandler;
 
+import backtype.storm.metric.api.MultiCountMetric;
 import backtype.storm.task.OutputCollector;
 import backtype.storm.task.TopologyContext;
 import backtype.storm.topology.OutputFieldsDeclarer;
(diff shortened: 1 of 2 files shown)


src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
```diff
@@ -152,21 +148,15 @@ public void prepare(Map conf, TopologyContext context,
...
 
-        this.eventMeters = context.registerMetric("parser-meter",
-                new MeterMetric(), 5);
-        this.eventTimers = context.registerMetric("parser-timer",
-                new TimerMetric(), 5);
-        this.eventHistograms = context.registerMetric("parser-histograms",
-                new HistogramMetric(), 5);
-
+        this.eventCounter = context.registerMetric(this.getClass()
+                .getSimpleName(), new MultiCountMetric(), 10);
```
GuiForget: How about using the component name instead of the class name?

GuiForget:
Looks good beside my quick comment on the metric's name

src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
jnioche: The component ID is already handled by the metrics mechanism and will be set for the datapoints.  Specifying the class here allows to add more info on which class was actually used, e.g. various flavours of parsers all running as 'parse' component.

Here is what you'd get by using the component name. Bit redundant, isn't it?

```
Mint15-Vostro-410:1027	  6:parse      	parse                  	{outlink_kept=952, tuple_in=14, error_out_link_parsing_MalformedURLException=105, tuple_success=14, outlink_filtered=119, outlink_outsideSourceDomainOrHostname=217}
```

jnioche:
thanks for reviewing it @GuiForget

Issue #66, opened 2015-01-19
ParserBolt to send outlinks to status stream, remove from default one?


jnioche:
The [ParserBolt](https://github.com/DigitalPebble/storm-crawler/blob/master/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java) currently sends the outlinks to the default stream alongside the other fields for a document ("url", "content", "metadata", "text"). It would be better to send them to the status stream instead, as we do for the Sitemaps, redirections etc...

jnioche:
If we decide to send the outlinks through the 'status' stream then it would be good to do it before the 0.4 release
@DigitalPebble/committers-crawler  any thoughts on this?

GuiForget:
I'm ok with that

jnioche:
see #67

PR #67, opened 2015-01-21
ParserBolt sends outlinks via status stream

jnioche:
see #66


jakekdodd:
Looks good to me

src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
```diff
@@ -300,8 +305,17 @@ public void execute(Tuple tuple) {
...
 
-        collector.emit(tuple, new Values(url, content, metadata, text.trim(),
-                slinks));
+        for (String outlink : slinks) {
+            // configure which metadata gets inherited from parent
+            Map<String, String[]> linkMetadata = metadataTransfer
+                    .getMetaForOutlink(metadata);
+            collector
+                    .emit(com.digitalpebble.storm.crawler.Constants.StatusStreamName,
+                            tuple, new Values(outlink, linkMetadata,
```
GuiForget: How about emitting the page URL as well?

jnioche:
> How about emitting the page URL as well?

because there could be other bolts between the parser and the updater, e.g indexing. we don't want to mark the URL as successful until it has finished all the things it has to do and/or make too many assumptions as to what a pipeline should contain.

One way to do would be to connect the status updater bolt to whichever bolt is last in the topology, even if it is on the default stream. We could also have a small Bolt that would generate am entry on the status stream for the main URL. This one would be put last in the topology.

src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
GuiForget: You misunderstood me. When emitting the outlinks, how about emitting within that tuple, the page URL as well. So your code would be 
```
           collector
                   .emit(com.digitalpebble.storm.crawler.Constants.StatusStreamName,
                           tuple, new Values(url, outlink, linkMetadata,
                                   Status.DISCOVERED));
```

GuiForget: Or maybe ensure that the linkMetadata contains the page URL?

jnioche: >  linkMetadata contains the page URL

why not. What do you have in mind?

GuiForget: Well I don't have anything in particular in mind. But I'm wondering how you will implement the click path that we want for our project and maybe the page URL could be handy

jnioche: I'm still not convinced that this is the best way of categorizing the content but that's another matter (e.g. can't use sitemaps for discovery then). It makes sense to be able to track the path indeed, at least for debugging. 

Could do that by modifying [MetadataTransfer](https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/java/com/digitalpebble/storm/crawler/util/MetadataTransfer.java#L50) so that it takes the sourceURL as argument and have the option to add that to a metadatum. It would inherit the ones from the parent metadata.

jnioche:
Have rebased and updated the branch + modified metadataTransfer as per my comment above so that it takes the source URL and optionally tracks the url path in the metadata. For some reason this seems to have caused carnage with the formatting, will reapply the formatting to the whole of the master branch before releasing anyway.

jnioche:
@DigitalPebble/committers-crawler will merge and commit soon unless someone objects

core/src/main/java/com/digitalpebble/storm/crawler/bolt/FetcherBolt.java
```diff
@@ -76,746 +76,746 @@
...
-            LOG.info("[Fetcher #{}] Missing value for field url in tuple {}",
-                    taskIndex, input);
-            // ignore silently
-            _collector.ack(input);
-            return;
-        }
-
-        fetchQueues.addFetchItem(input);
-    }
+	public static final Logger LOG = LoggerFactory.getLogger(FetcherBolt.class);
```
GuiForget: R u switching to tabs?

core/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
```diff
@@ -68,255 +69,267 @@
 @SuppressWarnings("serial")
 public class ParserBolt extends BaseRichBolt {
 
-    private Tika tika;
+	private Tika tika;
```
GuiForget: Same here. Looks like you have some code formatting issue

core/src/main/java/com/digitalpebble/storm/crawler/util/MetadataTransfer.java
```diff
@@ -28,37 +28,68 @@
  * what should be stored back in the persistence layer etc...
  **/
 public class MetadataTransfer {
+	/**
+	 * Parameter name indicating which metadata to transfer to the outlinks.
+	 * Boolean value.
+	 **/
+	public static final String metadataTransferParamName = "metadata.transfer";
```
GuiForget: Should be all capitalized no?

core/src/main/java/com/digitalpebble/storm/crawler/util/MetadataTransfer.java
```diff
@@ -28,37 +28,68 @@
...
-    private static String metadataTransferParamName = "metadata.transfer";
+	/**
+	 * Parameter name indicating whether to track the url path or not. Value is
+	 * either a vector or a single valued String.
+	 **/
+	public static final String trackPathParamName = "metadata.track.path";
 
-    private List<String> mdToKeep = new ArrayList<String>();
+	/** Metadata key name for tracking the source URLs **/
+	public static final String urlPathKeyName = "url.path";
```
GuiForget: So it took me a while to understand what you were doing even though I just requested this. Maybe call this "anchor.path" instead of "url.path" or anchor.text.path

jnioche: except that what we log here is the URL, not the anchor

core/src/main/java/com/digitalpebble/storm/crawler/util/MetadataTransfer.java
jnioche: We don't capitalize our param  names, just camel case them

core/src/main/java/com/digitalpebble/storm/crawler/util/MetadataTransfer.java
GuiForget: Ha ha! See, I definitely didn't understand what you were doing. Sorry

core/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
jnioche: yep, not clear why. Will check that the formatting is applied.

core/src/main/java/com/digitalpebble/storm/crawler/util/MetadataTransfer.java
jnioche: ;-) salut Gui. it's a bit like instant messaging, isn't it?

core/src/main/java/com/digitalpebble/storm/crawler/util/MetadataTransfer.java
GuiForget: I see. Constants has mixed names

core/src/main/java/com/digitalpebble/storm/crawler/util/MetadataTransfer.java
jnioche: could handle the anchors in a similar fashion, maybe in a separate PR

GuiForget: Yes that sounds like a good idea.

jnioche:
thanks for reviewing it

PR #69, opened 2015-01-22
Project reorganization

jakekdodd:
Project has been reorganized with a parent pom and two submodules, `core` and `external`.

The `core` submodule, whose artifact name is `storm-crawler-core`, contains all of the core components and dependencies for storm-crawler.

The `external` submodule, whose artifact name is `storm-crawler-external`, contains components with external dependencies (right now Elasticsearch and Librato), as well as non-core/optional components (such as metrics).


core/pom.xml
```diff
@@ -0,0 +1,157 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.storm</groupId>
+            <artifactId>storm-core</artifactId>
+        </dependency>
```
GuiForget: shouldn't scope be set to provided?

core/pom.xml
```diff
@@ -0,0 +1,157 @@
...
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+    <dependencies>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+        </dependency>
```
GuiForget: Test scope missing

external/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+                            <execution>
+                                <id>attach-test</id>
+                                <phase>none</phase>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
```
GuiForget: I would personally delete the whole `<profiles>` block for external since we don't use it right now

external/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.digitalpebble</groupId>
+            <artifactId>storm-crawler-core</artifactId>
+            <version>${project.version}</version>
+            <scope>provided</scope>
```
GuiForget: I don't think this is a provided scope here. It should be the default compile scope.

external/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+
+        <dependency>
+            <groupId>org.apache.storm</groupId>
+            <artifactId>storm-core</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+        </dependency>
```
GuiForget: scope: test

external/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+            </build>
+        </profile>
+    </profiles>
+
+    <dependencies>
+
+        <dependency>
+            <groupId>org.apache.storm</groupId>
+            <artifactId>storm-core</artifactId>
+        </dependency>
```
GuiForget: scope: provided

external/pom.xml
GuiForget: I take it back. this is already mentioned in the parent POM

external/pom.xml
GuiForget: Already mentioned in the parent POM

core/pom.xml
GuiForget: Already specified in parent POM

core/pom.xml
GuiForget: Already specified in parent POM

GuiForget:
I'm good w/ the proposed changes: I have 2 questions: one about the dependency scope for storm-crawler-core and the profiles block. Regarding the latter it's possible you need that in order to override what you have in the parent POM so ignore my comment if that's the case

jnioche:
Can't see ca06583144652cf4281c7fe9bdefffb2296105e9 in the diff file. Are the most recent changes included or does it need rebasing?

jakekdodd:
> Regarding the latter it's possible you need that in order to override what you have in the parent POM

Yeah that's the intent; I think that block is the proper way to override the test-jar in sc-external for the release profile.

> I don't think this is a provided scope here. It should be the default compile scope

Definitely shouldn't be provided, will have to change that before merging. 

@jnioche good catch--you can see that ca06583144652cf4281c7fe9bdefffb2296105e9 is in the commit log, so it's there, but misplaced in `src/test/...` in the parent project. I must have overlooked the file location when rebasing. Will move the class to the correct package when I fix the `sc-core` dependency scope

jakekdodd:
In fact, the commit history is all messy because I rebased several times after pushing to the remote. Even though we can do a fast-forward commit, the commit log will be convoluted. Will need to submit a new PR with a clean commit history

jakekdodd:
Re-branched and applied diffs to clean up the commit history; this has been superseded by #70.

PR #74, opened 2015-01-25
Implement filtering of URL by host/domain via the regular URLFilter

GuiForget:
- Extended the URLFilter interface to take the source URL and the metadata when filtering
- Also passed the storm config in the configure method, just like I did for the ParseFilter
- Replaced URLFilterUtil with HostURLFilter


core/crawler-conf.yaml
```diff
@@ -43,8 +43,6 @@ parsefilters.config.file: "parsefilters.json"
 urlfilters.config.file: "urlfilters.json"
 
 parser.htmlmapper.classname: "org.apache.tika.parser.html.IdentityHtmlMapper"
-parser.ignore.outlinks.outside.host: true
```
jnioche: These params are easier to change in the general config than buried in the urlfilters.json, since your patch allows the filters to be configured from the global config why not leave them there?

jnioche:
I like the simplification we'd get by having the host/domain filter like any other URL filter but am not so sure about requiring the source URL and metadata. The filter could be used at any stage of the pipeline for instance in the spouts where there would be no parent URL. We'd have to make sure that the mechanism is bullet-proof and can deal with null values elegantly. The alternative would be to deal with that through metadata but as seen in the issue this is not as straightforward.

core/src/main/java/com/digitalpebble/storm/crawler/filtering/URLFilters.java
```diff
@@ -96,43 +90,48 @@ public void configure(JsonNode jsonNode) {
...
+
             // check that it is available and implements the interface URLFilter
             try {
                 Class<?> filterClass = Class.forName(className);
                 boolean interfaceOK = URLFilter.class
                         .isAssignableFrom(filterClass);
                 if (!interfaceOK) {
-                    LOG.error("Class {} does not implement URLFilter",
-                            className);
+                    LOG.error("Filter {} does not implement URLFilter",
```
jnioche: if it does not implement URLFilter all we know about it is that it is a class. It is not incorrect to refer to it as such

core/src/main/java/com/digitalpebble/storm/crawler/filtering/host/HostURLFilter.java
```diff
@@ -0,0 +1,107 @@
...
+            }
+        } else {
+            ignoreOutsideDomain = false;
+        }
+    }
+
+    @Override
+    public String filter(URL sourceUrl, Map<String, String[]> sourceMetadata,
+            String urlToFilter) {
+        if (!ignoreOutsideHost && !ignoreOutsideDomain) {
```
jnioche: sourceUrl could be null e.g. when the filtering is done in a spout. Let's check that it has a non null, non empty value and simply return if it does

core/crawler-conf.yaml
GuiForget: Why the inconsistency if all other filters have their settings set in the JSON file?

core/src/main/java/com/digitalpebble/storm/crawler/filtering/URLFilters.java
GuiForget: Sure. The error message includes both the filter name and the class name. I thought it was easier to also include the filter's name rather than just the classname. But I can revert

GuiForget:
@jnioche updated to revert the log message about the class not implementing the URLFilter interface and to check for null before filtering. I indicated in the Javadoc that sourceUrl can be null

core/crawler-conf.yaml
jnioche: you are probably right, let' do it that way.

jnioche:
Thanks Gui

Issue #68, opened 2015-01-22
Normalise metadata returned by servers 


jnioche:
```
curl -I "http://twitter.com/home?status=Check+this+out%3ahttp%3a%2f%2fwww.dillards.com%2fwebapp%2fwcs%2fstores%2fservlet%2fProductDisplay%3fcatalogId%3d301%26productId"
HTTP/1.1 301 Moved Permanently
content-length: 0
date: Thu, 22 Jan 2015 12:11:11 UTC
location: https://twitter.com/home?status=Check+this+out%3ahttp%3a%2f%2fwww.dillards.com%2fwebapp%2fwcs%2fstores%2fservlet%2fProductDisplay%3fcatalogId%3d301%26productId
server: tsa_b
set-cookie: guest_id=v1%3A142192867165557827; Domain=.twitter.com; Path=/; Expires=Sat, 21-Jan-2017 12:11:11 UTC
x-connection-hash: ca85b658e4cc93174936c73ea46fa39c
x-response-time: 8
```

where *location* is in lowercase and our code expects it to be 'Location' in order to work. We could lowercase ALL keys in our metadata or alternatively normalise the key names at the protocol level.

GuiForget:
Per RFC, the field names are indeed case insensitive. I would lowercase them all in the protocol

jakekdodd:
See 527f93699315600ed2861ad34dc6060617afaa18

I identified side effects for:

* `com.digitalpebble.storm.crawler.protocol.HTTPHeaders`
* `com.digitalpebble.storm.crawler.protocol.http.HttpRobotRulesParser`

CrawlTopology seems to run fine, and inspecting the metadata in the debugger confirmed that it's all lowercase.

Can you guys think of any other side effect that need to be addressed?

GuiForget:
I think that's good. Can't think of anything else

jnioche:
Implemented in #77

PR #77, opened 2015-01-28
Lowercase headers

jakekdodd:
This PR lowercases HTTP header field names. See #68


core/src/main/java/com/digitalpebble/storm/crawler/protocol/http/HttpResponse.java
```diff
@@ -447,7 +447,7 @@ private void processHeaderLine(StringBuffer line) throws HttpException {
             valueStart++;
         }
         String value = line.substring(valueStart);
-        headers.put(key, new String[] { value });
+        headers.put(key.toLowerCase(), new String[] { value });
```
jnioche: toLowerCase(Locale.ROOT)? just so that we don't rely on the default locale

jakekdodd: Fixed in ced932387adb0245a17573bad5c491da6d6eecea

jnioche:
thanks Jake

PR #78, opened 2015-01-29
Replace Map<String,String[]> with serializable Metadata in Fetcher and protocol

jnioche:
The bolts still emit Map<String,String[]> to the subsequent bolts for now.

Replaces #36 

Still need to port to other parts of the code but we can already discuss possible improvements to it now.


jnioche:
Note to self \: read about serialization in Storm and Kryo 
[https://storm.apache.org/documentation/Serialization.html]
[https://github.com/EsotericSoftware/kryo#quickstart]

jnioche:
@DigitalPebble/committers-crawler could you please review this?

core/src/main/java/com/digitalpebble/storm/crawler/ConfigurableTopology.java
```diff
@@ -54,6 +54,8 @@ protected int submit(String name, Config conf, TopologyBuilder builder) {
         }
 
         else {
+            // register Metadata for serialization with FieldsSerializer
+            Config.registerSerialization(conf, Metadata.class);
```
GuiForget: There's also a non static registerSerialization on the Config object

core/src/main/java/com/digitalpebble/storm/crawler/Metadata.java
```diff
@@ -0,0 +1,145 @@
...
+import org.apache.commons.lang.StringUtils;
+
+/** Wrapper around Map <String,String[]> **/
+
+public class Metadata {
+
+    // TODO customize the behaviour of Kryo via annotations
+    // @BindMap(valueSerializer = IntArraySerializer.class, keySerializer =
+    // StringSerializer.class, valueClass = String[].class, keyClass =
+    // String.class, keysCanBeNull = false)
```
GuiForget: Is this comment still relevant>

core/src/main/java/com/digitalpebble/storm/crawler/Metadata.java
```diff
@@ -0,0 +1,145 @@
...
+     * Wraps an existing HashMap into a Metadata object - does not clone the
+     * content
+     **/
+    public Metadata(Map<String, String[]> metadata) {
+        if (metadata == null)
+            throw new NullPointerException();
+        md = metadata;
+    }
+
+    public Map<String, String[]> getMap() {
```
GuiForget: I see this method only used to get the entrySet. Maybe just expose the entrySet instead?

core/src/main/java/com/digitalpebble/storm/crawler/Metadata.java
```diff
@@ -0,0 +1,145 @@
...
+            return null;
+        return values;
+    }
+
+    public void setValue(String key, String value) {
+        md.put(key, new String[] { value });
+    }
+
+    public void setValues(String key, String[] values) {
+        md.put(key, values);
```
GuiForget: SHould we check that values is neither null nor empty?

core/src/main/java/com/digitalpebble/storm/crawler/Metadata.java
```diff
@@ -0,0 +1,145 @@
...
+        String[] values = md.get(key);
+        if (values == null)
+            return null;
+        if (values.length == 0)
+            return null;
+        return values;
+    }
+
+    public void setValue(String key, String value) {
+        md.put(key, new String[] { value });
```
GuiForget: Should we check that value is not blank to be consistent with addValue?

core/src/main/java/com/digitalpebble/storm/crawler/Metadata.java
```diff
@@ -0,0 +1,145 @@
...
+    public void addValues(String key, Collection<String> values) {
+        if (values == null || values.size() == 0)
+            return;
+        String[] existingvals = md.get(key);
+        if (existingvals == null) {
+            md.put(key, values.toArray(new String[values.size()]));
+            return;
+        }
+
+        ArrayList<String> existing = new ArrayList<String>(existingvals.length);
```
GuiForget: Should be existingVals.length + values.size()

core/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
```diff
@@ -169,7 +168,7 @@ public void execute(Tuple tuple) {
         // rely on mime-type provided by server or guess?
 
         ByteArrayInputStream bais = new ByteArrayInputStream(content);
-        Metadata md = new Metadata();
+        org.apache.tika.metadata.Metadata md = new org.apache.tika.metadata.Metadata();
```
GuiForget: Makes me wonder if we should rename our class to CrawlMetadata

core/src/main/java/com/digitalpebble/storm/crawler/bolt/URLPartitionerBolt.java
```diff
@@ -74,9 +77,9 @@ public void execute(Tuple tuple) {
         // IP in metadata?
         if (mode.equalsIgnoreCase(Constants.PARTITION_MODE_IP)
                 && metadata != null) {
```
GuiForget: I think metadata can no longer be null

core/src/main/java/com/digitalpebble/storm/crawler/protocol/http/HttpProtocol.java
```diff
@@ -202,22 +202,23 @@ public Config getConf() {
...
         long startTime = System.currentTimeMillis();
         HttpResponse response = new HttpResponse(this, u); // make a request
-        HashMap<String, String[]> metadata = response.getHeaders();
+        Metadata metadata = response.getHeaders();
 
         if (this.responseTime) {
             int elapsedTime = (int) (System.currentTimeMillis() - startTime);
-            metadata.put("_rst_",
-                    new String[] { Integer.toString(elapsedTime) });
+            metadata.setValue("_rst_", Integer.toString(elapsedTime));
```
GuiForget: Any reason why this metadata key is somewhat obfuscated?

core/src/main/java/com/digitalpebble/storm/crawler/protocol/http/HttpResponse.java
```diff
@@ -139,8 +143,7 @@ public HttpResponse(HttpProtocol http, URL url) throws IOException {
 
             this.conf = http.getConf();
             if (ConfUtils.getBoolean(conf, "store.ip.address", false) == true) {
-                headers.put("_ip_", new String[] { sockAddr.getAddress()
-                        .getHostAddress() });
+                headers.setValue("_ip_", sockAddr.getAddress().getHostAddress());
```
GuiForget: Same here

GuiForget:
Looks good to me. I noticed your code formatter adds a space after the `*` in empty javadoc line.

core/src/main/java/com/digitalpebble/storm/crawler/ConfigurableTopology.java
jnioche: any reason to use one and not the other?

core/src/main/java/com/digitalpebble/storm/crawler/Metadata.java
jnioche: yes, I think this is something we might want to look into later maybe. let's leave it as a comment for now

core/src/main/java/com/digitalpebble/storm/crawler/Metadata.java
jnioche: it was used to facilitate the transition, might have a use for it later? don't know we can add it back if so.
better than adding entrySet I will add a addAll(Metadata md) method as this is basically what we do with the iteration on the entrySet.

core/src/main/java/com/digitalpebble/storm/crawler/Metadata.java
jnioche: I can think of cases where we'd want to store only a key and not be interested in the values. let's leave it as is

core/src/main/java/com/digitalpebble/storm/crawler/Metadata.java
jnioche: maybe we should remove this check from addValue?

core/src/main/java/com/digitalpebble/storm/crawler/Metadata.java
jnioche: would be better indeed

core/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
jnioche: there is nothing about it that is specific to crawling, could be use for storing any other type of metadata. The package name is probably enough to differentiate.

core/src/main/java/com/digitalpebble/storm/crawler/bolt/URLPartitionerBolt.java
jnioche: that's right

core/src/main/java/com/digitalpebble/storm/crawler/protocol/http/HttpProtocol.java
jnioche: that's how we inherited it from Nutch, can think about that one in a separate issue maybe

jnioche:
Thanks for reviewing this Gui. Haven't found how to change the code formatter to do that. Will start working on the merge and commit once I've heard back from you on the points above.

core/src/main/java/com/digitalpebble/storm/crawler/ConfigurableTopology.java
GuiForget: None. Except my preference for non-static method :)

core/src/main/java/com/digitalpebble/storm/crawler/Metadata.java
GuiForget: Not sure why you think addValue should be different but I can live with that :)

GuiForget:
Looks good

Issue #76, opened 2015-01-28
Fix release artefacts


jnioche:
The release is done with the following commands \:

mvn release:clean release:prepare -P release
mvn release:perform -P release

Since refactoring into core and external I am seeing the following 

*uber-jar published for core* 

.../content/com/digitalpebble/storm-crawler-core/0.4/storm-crawler-core-0.4-jar-with-dependencies.jar

which should not have happened since I specify the 'release' profile.

*generated tests for parent pom* 

.../com/digitalpebble/storm-crawler/0.4/storm-crawler-0.4-tests.jar

which should not have happened.

Any ideas? I will release 0.4 with these but it would be good to fix that for the next release

jnioche:
and maybe rename the parent pom into 'storm-crawler-parent' to prevent any confusion as 'storm-crawler' used to contain the code but does not anymore

jakekdodd:
Should be easy fixes, I'll get on that.

jakekdodd:
Addressed in #80

PR #80, opened 2015-01-29
Fix maven release profile

jakekdodd:
No jar-with-dependencies; no test jar for parent.


jakekdodd:
Addresses #76 

@jnioche When you get a chance, can you check that everything is fixed with `mvn release:prepare -DdryRun=true`?

jnioche:
great, thanks Jake.

PR #91, opened 2015-02-11
Update to the ParserBolt

GuiForget:
- Add a boolean to skip the emit of outlinks
- wrap the parseFilters within a try/catch

Note that one subtle change I made is that if we fail to build a URL object
out of the url String, we simply skip the outlinks emitting. The comment
hints toward this not be possible so I don't think that's an issue

### DIFF
--- a/core/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
+++ b/core/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
@@ -17,11 +17,12 @@
 
 package com.digitalpebble.storm.crawler.bolt;
 
+import static com.digitalpebble.storm.crawler.Constants.StatusStreamName;
+
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -51,6 +52,7 @@ import backtype.storm.tuple.Fields;
 import backtype.storm.tuple.Tuple;
 import backtype.storm.tuple.Values;
 
+import com.digitalpebble.storm.crawler.Constants;
 import com.digitalpebble.storm.crawler.Metadata;
 import com.digitalpebble.storm.crawler.filtering.URLFilters;
 import com.digitalpebble.storm.crawler.parse.DOMBuilder;
@@ -83,10 +85,13 @@ public class ParserBolt extends BaseRichBolt {
     private Class<?> HTMLMapperClass = IdentityHtmlMapper.class;
 
     private MetadataTransfer metadataTransfer;
+    private boolean emitOutlinks = true;
 
     @Override
     public void prepare(Map conf, TopologyContext context,
             OutputCollector collector) {
+
+        emitOutlinks = ConfUtils.getBoolean(conf, "parser.emitOutlinks", true);
 
         String urlconfigfile = ConfUtils.getString(conf,
                 "urlfilters.config.file", "urlfilters.json");
@@ -99,6 +104,8 @@ public class ParserBolt extends BaseRichBolt {
                 throw new RuntimeException(
                         "Exception caught while loading the URLFilters", e);
             }
+        } else {
+            urlFilters = URLFilters.emptyURLFilters;
         }
 
         String parseconfigfile = ConfUtils.getString(conf,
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
```diff
@@ -201,22 +204,24 @@ public void execute(Tuple tuple) {
...
-                    com.digitalpebble.storm.crawler.Constants.StatusStreamName,
-                    tuple, new Values(url, metadata, Status.ERROR));
+            // send to status stream in case another component wants to update
+            // its status
+            metadata.setValue(Constants.STATUS_ERROR_SOURCE, "content parsing");
+            metadata.setValue(Constants.STATUS_ERROR_MESSAGE, errorMessage);
+            collector.emit(StatusStreamName, tuple, new Values(url, metadata,
+                    Status.ERROR));
             collector.ack(tuple);
-            eventCounter.scope("parse exception").incrBy(1);
```
jnioche: I presume you removed this because we already emit a more specific counter above. Leaving this one would give a more global view, regardless of the type of error.

core/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
```diff
@@ -201,22 +204,24 @@ public void execute(Tuple tuple) {
...
 
         // parse
+        String text;
         try {
             tika.getParser().parse(bais, teeHandler, md, parseContext);
             text = textHandler.toString();
         } catch (Exception e) {
-            LOG.error("Exception while parsing {}", url, e.getMessage());
+            String errorMessage = "Exception while parsing " + url + ": "
+                    + e.getMessage();
```
jnioche: What if the exception does *not* have a message and returns null?
Use a StringBuffer instead of string concatenation?

core/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
```diff
@@ -238,8 +242,46 @@ public void execute(Tuple tuple) {
         LOG.info("Parsed {} in {} msec", url, duration);
 
         // apply the parse filters if any
-        parseFilters.filter(url, content, root, metadata);
+        try {
+            parseFilters.filter(url, content, root, metadata);
+        } catch (RuntimeException e) {
+            String errorMessage = "Exception while running parse filters on "
```
jnioche: same as above

core/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
```diff
@@ -249,13 +291,10 @@ public void execute(Tuple tuple) {
             LOG.error("MalformedURLException on {}", url);
             eventCounter.scope(
                     "error_outlinks_parsing_" + e1.getClass().getSimpleName())
-                    .incrBy(1);
-            collector.fail(tuple);
-            eventCounter.scope("tuple_fail").incrBy(1);
+                    .incr();
```
jnioche: you're being pedantic here ;-)

core/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
GuiForget: OK I hadn't thought about it. I'll add comments

core/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
GuiForget: Internally the compiler will use a StringBuilder anyway so I don't see the value. I will change it to use e.toString() however which includes the class of the exception

core/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
GuiForget: Because my comment got lost, I'll add it back here: The Java compiler uses a StringBuilder when concatenating multiple Strings. So for readability purposes, I prefer using the `+` sign. As far as the getMessage() being null, this has been addressed by using toString instead (implied by the String concatenation)

core/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
```diff
@@ -269,8 +310,8 @@ public void execute(Tuple tuple) {
                 urlOL = tmpURL.toExternalForm();
             } catch (MalformedURLException e) {
                 LOG.debug("MalformedURLException on {}", l.getUri());
-                eventCounter.scope(
-                        "error_out_link_parsing_"
+                eventCounter
+                        .scope("error_outlink_parsing_"
```
GuiForget: I changed the name of the metric here to have a consistent spelling for "outlink"

GuiForget:
Thanks @jnioche for the feedback. I have updated my PR accordingly

jnioche:
Great stuff, thanks!

PR #95, opened 2015-02-25
Fix compilation error

GuiForget:
Not sure this is right or if I should upgrade to a more recent version of ElasticSearch...


GuiForget:
cc @jnioche

jnioche:
I'll upgrade the version of ES shortly anyway.

PR #98, opened 2015-03-01
Add a normalizer to remove specified query parameters.

acinader:
Easier to reason about than writing regular expressions to normalize a particular query
parameter.  Handles things like remove the ? if there are no more query parameters.

@jnioche not sure if you'd think this generally useful? I can easily pull into our own project, but would like your feedback anyway :) so started here..

@GuiForget any pointers on java idioms, etc always welcome.

### DIFF
--- a/core/src/main/java/com/digitalpebble/storm/crawler/filtering/basic/BasicURLFilter.java
+++ b/core/src/main/java/com/digitalpebble/storm/crawler/filtering/basic/BasicURLFilter.java
@@ -18,16 +18,31 @@
 package com.digitalpebble.storm.crawler.filtering.basic;
 
 import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
 
 import com.digitalpebble.storm.crawler.Metadata;
 import com.digitalpebble.storm.crawler.filtering.URLFilter;
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+
+import org.apache.http.NameValuePair;
+import org.apache.http.client.utils.URIBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class BasicURLFilter implements URLFilter {
 
+    private static final Logger LOG = LoggerFactory.getLogger(BasicURLFilter.class);
+
     boolean removeAnchorPart = true;
+    final Set<String> queryElementsToRemove = new TreeSet<>();
 
     @Override
     public String filter(URL sourceUrl, Metadata sourceMetadata,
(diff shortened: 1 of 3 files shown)


jnioche:
Would be useful indeed. Could be as a separate filter but also part of [https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/java/com/digitalpebble/storm/crawler/filtering/basic/BasicURLFilter.java]. Not sure we really need to have a separate config file for that, the list of filters to remove could be put in the 'normal' config as the list would be used across all the URLs and typically limited to a few values. 

What this filter could do would be to sort the key/values so that if some URLs differ only in the order of the K/V they'd be normalised to the same form.

And yes, this project is Java 7 for now ;-)

core/src/main/java/com/digitalpebble/storm/crawler/filtering/basic/QueryParameterNormalizer.java
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * Basic filter to remove query parameters from urls so parameters that don't change the content of
+ * the page can be removed. An example would be a google analytics query parameter like
+ * "utm_campaign" which might have several different values for a url that points to the same
+ * content.
+ */
+public class QueryParameterNormalizer implements URLFilter {
+
+    private List<String> parametersToRemove;
+    private static final Logger LOG = LoggerFactory.getLogger(QueryParameterNormalizer.class);
```
GuiForget: static normally goes at the top

core/src/main/java/com/digitalpebble/storm/crawler/filtering/basic/QueryParameterNormalizer.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public String filter(URL sourceUrl, Metadata sourceMetadata, String urlToFilter) {
+        URIBuilder uriBuilder;
+        try {
+            // Handle illegal characters by making a url first
+            // this will clean illegal characters like |
+            URL url = new URL(urlToFilter);
+            URI uri =
+                    new URI(url.getProtocol(), url.getAuthority(), url.getPath(), url.getQuery(),
+                            url.getRef());
```
GuiForget: Is there a reason for not using `URL.toURI()` directly?

core/src/main/java/com/digitalpebble/storm/crawler/filtering/basic/QueryParameterNormalizer.java
```diff
@@ -0,0 +1,107 @@
...
+            uriBuilder = new URIBuilder(uri);
+        } catch (URISyntaxException e) {
+            LOG.warn("Invalid urlToFilter {}. {}", urlToFilter, e);
+            return urlToFilter;
+        } catch (MalformedURLException e) {
+            LOG.warn("Innvalid urlToFilter {}. {}", urlToFilter, e);
+            return urlToFilter;
```
GuiForget: Wouldn't you want to filter out the URLs that are not valid?

core/src/main/java/com/digitalpebble/storm/crawler/filtering/basic/QueryParameterNormalizer.java
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Basic filter to remove query parameters from urls so parameters that don't change the content of
+ * the page can be removed. An example would be a google analytics query parameter like
+ * "utm_campaign" which might have several different values for a url that points to the same
+ * content.
+ */
+public class QueryParameterNormalizer implements URLFilter {
+
+    private List<String> parametersToRemove;
```
GuiForget: Should probably be a set rather than a list

core/src/main/java/com/digitalpebble/storm/crawler/filtering/basic/QueryParameterNormalizer.java
```diff
@@ -0,0 +1,107 @@
...
+        List<NameValuePair> paramsToKeep = new ArrayList<>();
+        for (NameValuePair param : params) {
+            if (!parametersToRemove.contains(param.getName())) {
+                paramsToKeep.add(param);
+            }
+        }
+        uriBuilder.removeQuery();
+        // would be nice to use a newer version of URIBuilder, let's you set all params at once
+        for (NameValuePair param : paramsToKeep) {
+            uriBuilder.setParameter(param.getName(), param.getValue());
```
GuiForget: You should use addParameter instead

core/src/main/java/com/digitalpebble/storm/crawler/filtering/basic/QueryParameterNormalizer.java
acinader: i caught that before reading this. no really.  i did :)

core/src/main/java/com/digitalpebble/storm/crawler/filtering/basic/QueryParameterNormalizer.java
acinader: there is!

if you do toUri() directly, it'll throw a ```URISyntaxException```

but the Uri constructor used above will:
```If a scheme-specific part is given then it is appended. Any character that is not a legal URI character is quoted.```

File: core/src/main/java/com/digitalpebble/storm/crawler/filtering/basic/QueryParameterNormalizer.java (hunk 3 above)
Comment (acinader) [inline core/src/main/java/com/digitalpebble/storm/crawler/filtering/basic/QueryParameterNormalizer.java]: i thought i didn't but misunderstood something....

**Comment by acinader** [comment]
@jnioche yes, this is much better.

I don't think we want to sort the params.  I suspect that most competent retailers will be sorting params for caching anyway???

**Comment by GuiForget** [comment]
"competent retailers" :hear_no_evil:

**Comment by kkrugler** [comment]
1. Re sorting the query parameters...yes, sadly that's important.

2. I might be able to generate a candidate list of query parameters from a large set of URLs...does anyone else have a set of these?

3. This part of URL normalization is something that many web crawlers need, and (sadly) everyone implements on their own. E.g. Nutch has BasicURLNormalizer, QuerystringURLNormalizer, RegexURLNormalizer, AjaxURLNormalizer, and HostURLNormalizer. Bixo has its own set. So it would be great to add this to crawler-commons, but that's probably out-of-scope.

**Comment by acinader** [comment]
added sort to the query params.

@kkrugler I'd certainly be interested in looking at the candidate list!

**Comment by acinader** [comment]
note that i re-arranged the order of discovery.urlfilters.json so that the regex normalizers run before the BasicURLFilter.  I did this so that my clean up of urls like http://host&foo=bar to http://host?foo=bar will run before the BasicURLFilter so the urls in proper form.

File: core/src/main/java/com/digitalpebble/storm/crawler/filtering/basic/BasicURLFilter.java [hunk 6]
```diff
@@ -43,14 +60,76 @@ public String filter(URL sourceUrl, Metadata sourceMetadata,
...
+        uriBuilder.removeQuery();
+        Collections.sort(paramsToKeep, comp);
+        for (NameValuePair param : paramsToKeep) {
+            uriBuilder.addParameter(param.getName(), param.getValue());
+        }
+
+        return uriBuilder.toString();
+    }
+
+    Comparator<NameValuePair> comp = new Comparator<NameValuePair>() {
```
acinader: this wont work in java 8.  major facelift for the Comparator!

PR #101, opened 2015-03-03
Fix a bad query string that starts with a '&' instead of a '?'

acinader:
A common error to find is a query string that starts with an & instead of a ? This will fix
that error. So http://foo.com&a=b will be changed to http://foo.com?a=b.

### DIFF
--- a/core/src/main/java/com/digitalpebble/storm/crawler/filtering/basic/BasicURLFilter.java
+++ b/core/src/main/java/com/digitalpebble/storm/crawler/filtering/basic/BasicURLFilter.java
@@ -44,6 +44,7 @@ public class BasicURLFilter implements URLFilter {
     private static final Logger LOG = LoggerFactory.getLogger(BasicURLFilter.class);
 
     boolean removeAnchorPart = true;
+    boolean unmangleQueryString = true;
     final Set<String> queryElementsToRemove = new TreeSet<>();
 
     @Override
@@ -60,6 +61,10 @@ public class BasicURLFilter implements URLFilter {
             }
         }
 
+        if (unmangleQueryString) {
+            urlToFilter = unmangleQueryString(urlToFilter);
+        }
+
         if (!queryElementsToRemove.isEmpty()) {
             urlToFilter = filterQueryElements(urlToFilter);
         }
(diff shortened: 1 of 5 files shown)


GuiForget:
Why should we fix this error and how can we tell apart the URLs that did intend to have an ampersand in their path?

acinader:
well, this should only catch a case where there is an ampersand before a question mark which shouldn't be valid.  it's a common mistake.  In the extremely unlikely case that you've got a site that means to do this, you can turn the filter off.

jakekdodd:
Take a look at the [default regex normalizers](https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/resources/default-regex-normalizers.xml), there's one that cleans `?&var=value` to `?var=value`. I think that's close to, or exactly, what you're looking for. There are several other normalizers in there that fix some common mistakes.

I think generally it's a good idea to group normalizers into normalizing classes, and filters into filtering classes. Both because `BasicUrlFilter` will become too bloated, and because there are performance benefits to filtering before normalizing. 

Having said that, the XML-regex config file is totally nasty. As a user, I'd probably take one look at that thing and say 'no thanks.' Let's make it better, any ideas?

jnioche:
Looking at the code of *BasicURLFilter*, we should rename it as *BasicURLNormalizer* as it is more about rewriting URLs than removing them. What the *BasicURLFilter* could do is e.g. remove URLs that have too long a domain name (I have seen monstrosities in the past, trust me), too long a query part etc... The *BasicURLNormalizer* is still relatively lightweight and would typically be done early in the sequence of filters.

Re-separation between filters and normalizers, I intentionally  wanted to avoid the situation we have in Nutch were we have separate interfaces for them. Instead we have just one here and 'filtering' is a generic operation which covers both removal and rewriting of URLs. This allows us to mix the operations as we see fit, with the general idea that cheapest operations should be done first as you pointed out. Now from an implementation perspective it is probably clearer to name the filters based on whether they rewrite or remove URLs but let's not make it too rigid.

acinader:
I renamed BasicURLFilter to BasicURLNormalizer.  

I have a very slight preference having this in the Basic vs. Regex normalizer.   I def like having the url param remover in Basic since it would otherwise require two regex expressions.  In any event, this should work as advertised ;)

jnioche:
> I def like having the url param remover in Basic since it would otherwise require two regex expressions

and it is something we'd want for any crawl, whereas the regex based stuff is where you'd have user specific normalisations

PR #111, opened 2015-03-19
ParseFilters get list of Outlink from parser

jnioche:
We might want the ParseFilters to have access to the Oulinks extracted by the parsers. This is useful when creating new outlinks programatically e.g. navigational links that can't be found in Javascript but also in order to attach arbitrary metadata for some outlinks. It can also be used to do some filtering from within the parsefilters.

The parserbolt currently filters and generates the metadata for the outlinks before calling the parsefillters. 

@DigitalPebble/committers-crawler   can you please review and comment?


jakekdodd:
Haven't had a chance to test it, but everything looks fine to me!

Quick question...how does LinkContentHandler treat relative path links? It might be useful to put a helper method in Outlink that cals `URLUtil.resolveURL`.

core/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
```diff
@@ -264,9 +270,14 @@ public void execute(Tuple tuple) {
         }
 
         if (emitOutlinks) {
-            List<Link> links = linkHandler.getLinks();
-            if (!links.isEmpty()) {
-                emitOutlinks(tuple, url, metadata, links);
+            if (!outlinks.isEmpty()) {
```
GuiForget: There isn't much value in that if check anymore

jnioche:
thanks guys

> how does LinkContentHandler treat relative path links?

it does not resolve them and keeps them relative

> It might be useful to put a helper method in Outlink that cals URLUtil.resolveURL.

something like `static makeAbsolute(URL parentURL, String targetURL)`

why not. Calling URLUtil.resolveURL is probably just as easy.

core/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
jnioche: fixed in [master 7d70198]

PR #113, opened 2015-04-07
Removed write limit from Tika handler to get full text

dremeika:
Tika shouldn't limit content size. It's already done using http.content.limit property.

### DIFF
--- a/core/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
+++ b/core/src/main/java/com/digitalpebble/storm/crawler/bolt/ParserBolt.java
@@ -177,7 +177,7 @@ public class ParserBolt extends BaseRichBolt {
         org.apache.tika.metadata.Metadata md = new org.apache.tika.metadata.Metadata();
 
         LinkContentHandler linkHandler = new LinkContentHandler();
-        ContentHandler textHandler = new BodyContentHandler();
+        ContentHandler textHandler = new BodyContentHandler(-1);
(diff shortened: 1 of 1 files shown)


jnioche:
Thanks Donatas, I hadn't noticed the implicit limitation.
I would make sense to add something in the config to be able to limit the size of the text sent, regardless of `http.content.limit`. The latter can influence whether the document will be parseable at all (think pdf docs), sometime you need to set it to -1 but still want to limit the amount of text passed to the following bolts.  The limitation on the text length could also be done in [AbstractIndexerBolt](https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/java/com/digitalpebble/storm/crawler/indexing/AbstractIndexerBolt.java).

jnioche:
BTW if you are dealing with HTML docs only I'd recommend that you use the parser in  [https://github.com/PopSugar/storm-crawler-extensions]. Tika filters/normalises the HTML content a lot and if you use XPath expression to extract data from pages then the JSoup based parser should be a better option.

dremeika:
Yes, I am dealing with HTML only and I only need parser to discover outlinks. Tika seems to do the job, but JSoup should perfrom better, I guess. Will try it. Thanks.

jnioche:
@dremeika see [master 6a2d145]. It is now in the core code, thanks to @GuiForget and Shopstyle.com who kindly donated it to Storm Crawler.

dremeika:
Great! Big thanks @GuiForget and @jnioche .

PR #115, opened 2015-04-09
Added support for Json config instead of filename.

dang-sugarinc:
config format:
{
  "urlFilters": [
    "-^http:\/\/dimg.dillards.com",
    "-^http:\/\/www.dillards.com\/webapp"
  ],
  "urlNormalizers": [
    {
        "pattern": "^(http://www.dillards.com/product/.+)(\?.*)$",
        "substitution": "$1"
    },
    {
        "pattern": "somepattern2",
        "substitution": "replacement2"
    }
  ]
}

### DIFF
--- a/core/src/main/java/com/digitalpebble/storm/crawler/filtering/regex/RegexURLFilterBase.java
+++ b/core/src/main/java/com/digitalpebble/storm/crawler/filtering/regex/RegexURLFilterBase.java
@@ -34,6 +34,7 @@ import org.slf4j.LoggerFactory;
 import com.digitalpebble.storm.crawler.Metadata;
 import com.digitalpebble.storm.crawler.filtering.URLFilter;
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
 
 /**
  * An abstract class for implementing Regex URL filtering. Adapted from Apache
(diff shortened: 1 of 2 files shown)


dang-sugarinc:
@GuiForget Please comment.

core/src/main/java/com/digitalpebble/storm/crawler/filtering/regex/RegexURLFilterBase.java
```diff
@@ -49,22 +50,38 @@
...
+                try {
+                    RegexRule rule = createRule(urlFilterNode.asText());
+                    if (rule != null) {
+                        rules.add(rule);
+                    }
+                } catch (IOException e) {
+                    LOG.error("There was an error reading regex filter {}",
+                            urlFilterNode.asText());
+                    e.printStackTrace();
+                }
```
GuiForget: I think to keep the configure method lean, it would be better to have a readRules method which takes an ArrayNode

core/src/main/java/com/digitalpebble/storm/crawler/filtering/regex/RegexURLFilterBase.java
```diff
@@ -49,22 +50,38 @@
...
+            for (JsonNode urlFilterNode : array) {
+                try {
+                    RegexRule rule = createRule(urlFilterNode.asText());
+                    if (rule != null) {
+                        rules.add(rule);
+                    }
+                } catch (IOException e) {
+                    LOG.error("There was an error reading regex filter {}",
+                            urlFilterNode.asText());
+                    e.printStackTrace();
```
GuiForget: Pass `e` to the LOG.error instead

core/src/main/java/com/digitalpebble/storm/crawler/filtering/regex/RegexURLNormalizer.java
```diff
@@ -42,14 +48,10 @@
 import org.w3c.dom.Text;
 import org.xml.sax.InputSource;
 
-import com.digitalpebble.storm.crawler.Metadata;
-import com.digitalpebble.storm.crawler.filtering.URLFilter;
-import com.fasterxml.jackson.databind.JsonNode;
```
GuiForget: Hum... Not sure storm-crawler has the same import ordering rule. You need to override your Eclipse settings so storm-crawler-core project keeps using the original/default formatting rules

core/src/main/java/com/digitalpebble/storm/crawler/filtering/regex/RegexURLNormalizer.java
```diff
@@ -72,15 +74,44 @@
...
+                    substitutionValue = substitutionNode.asText();
+                }
+                if (patternNode != null
+                        && StringUtils.isNotBlank(patternNode.asText())) {
+                    Rule rule = createRule(patternNode.asText(),
+                            substitutionValue);
+                    if (rule != null) {
+                        rules.add(rule);
+                    }
+                }
```
GuiForget: Same comment here. Keep the configure method lean

dang-sugarinc:
@GuiForget Made changes based on comments.

GuiForget:
@jnioche take a look and let us know if you disagree with the changes

jnioche:
The title of this issue is misleading\: you don't do `instead` but `as well as`, which is fine. We still read the default config files.

Maybe add an example of filter or normaliser configuration in the json files? Could you please update the WIKI page [https://github.com/DigitalPebble/storm-crawler/wiki/URLFilters] and give an  example there.

dang-sugarinc:
WIKI page updated.

jnioche:
Great! Thanks

Issue #114, opened 2015-04-08
Use JSoupParserBolt in example topologies and move tika parser to external


jnioche:
[master 6a2d145] Added JSoupParserBolt (donated by Shopstyle)

which was the one in [https://github.com/PopSugar/storm-crawler-extensions/tree/master/jsoup-parser]

This should be used instead of the Tika-based one which is not a great option for HTML content.

The Tika parser should probably be moved to external \: it has tons of dependencies and this would make the generated jars a lot lighter. It will still be available from external if people need to process non HTML documents.

@DigitalPebble/committers-crawler any thoughts on this?

jakekdodd:
:+1:

jnioche:
Looks like we need  to have Xerces as a dependency because of `org/apache/html/dom/HTMLDocumentImpl` which is also used in the JSoupParser if we remove Tika. See [https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/java/com/digitalpebble/storm/crawler/parse/JSoupDOMBuilder.java#L73]. I am wondering whether there is a way of achieving this without using Xerces at all.

PR #116, opened 2015-04-13
Moved Tika parser to external #114

jnioche:
See #114 
@DigitalPebble/committers-crawler please review and test


jnioche:
Annoyingly the sitemap partser in CommonCrawl has a dependency to Tika parsers so we might end up having to declare it in pom. No point moving it to external then... Will investigate

GuiForget:
Looks good to me. Doesn't break our own crawlers but we don't use external. For that reason I didn't dare merging

PR #119, opened 2015-04-15
Rewrite BasicURLNormalizer so URLs don't get double encoded

### LINKED ISSUE #118: URLs with escaped characters get double encoded by the BasicURLNormalizer
http://www.dillards.com/product/ASICS-Womens-GT2000-3-LiteShow%E2%84%A2-Running-Shoes_301_-1_301_504736989

is normalized as 

http://www.dillards.com/product/ASICS-Womens-GT2000-3-LiteShow%25E2%2584%25A2-Running-Shoes_301_-1_301_504736989

GuiForget:
URL encoding is so tricky! And there are basically no good library
that handles it properly, mostly because of the lack of proper support
via the URL and URI classes.

In any event, I rewrote filterQueryElements so that it no longer
double encode the URLs. This is done by simply only modifying the
query string and leaving the other elements unchanged.

I added some 2 unit tests to validate my code.

Fixes #118

### DIFF
--- a/core/src/main/java/com/digitalpebble/storm/crawler/filtering/basic/BasicURLNormalizer.java
+++ b/core/src/main/java/com/digitalpebble/storm/crawler/filtering/basic/BasicURLNormalizer.java
@@ -18,14 +18,15 @@
 package com.digitalpebble.storm.crawler.filtering.basic;
 
 import java.net.MalformedURLException;
-import java.net.URI;
-import java.net.URISyntaxException;
 import java.net.URL;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Scanner;
 import java.util.Set;
 import java.util.TreeSet;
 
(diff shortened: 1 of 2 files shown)


GuiForget:
@jnioche any feedback?

jnioche:
not yet :-( will try to have a look at it tomorrow

GuiForget:
No worries. Just wanted to make sure you had seen it :)

core/src/main/java/com/digitalpebble/storm/crawler/filtering/basic/BasicURLNormalizer.java
```diff
@@ -105,35 +107,45 @@ public void configure(Map stormConf, JsonNode paramNode) {
...
-        List<NameValuePair> paramsToKeep = new ArrayList<>();
-        for (NameValuePair param : params) {
-            if (!queryElementsToRemove.contains(param.getName())) {
-                paramsToKeep.add(param);
+            if (StringUtils.isEmpty(url.getQuery())) {
+                return urlToFilter;
+            }
+
+            List<NameValuePair> pairs = new ArrayList<NameValuePair>();
+            URLEncodedUtils.parse(pairs, new Scanner(url.getQuery()), "UTF-8");
```
jnioche: can we be 100% sure that the encoding used will always be UTF-8?

GuiForget: Several answers for you:
- In absolute no, in practice yes
- the same % that was assumed before my code change since UTF-8 is assumed when you construct a URI from a URL since URI are in UTF-8 always but URL don't have to be.

PR #122, opened 2015-04-17
Replace HTTP protocol implementation

### LINKED ISSUE #23: Replace HTTP protocol implementation
Before I did anything with this, wanted to check and see if there was a reason why HTTP 1.0 was hardcoded. Right now, line 167 of HttpResponse sets the protocol version to 1.0, regardless of whether ``http.useHttp11: true`` is set in the configuration:

```java
StringBuffer reqStr = new StringBuffer("GET ");
            if (http.useProxy()) {
                reqStr.append(url.getProtocol() + "://" + host + portString
                        + path);
            } else {
                reqStr.append(path);
            }

            reqStr.append(" HTTP/1.0\r\n");
```

Is this intentional?

jnioche:
Removes the existing code for the http protocol handling which came from Nutch and uses the httpclient library instead.  

The amount of code is greatly reduced as a result and we should have a more robust processing by leveraging an external library instead of rolling our own low level code. I have seen better results on https compared to the existing version.

The only missing features at this stage are \:
* store.ip.address
* maxContent
* acceptLanguage
* accept
* http.store.responsetime

These are relatively minor and could be added afterwards

Fixes #23


core/pom.xml
```diff
@@ -24,6 +24,7 @@
 		<jsoup.version>1.8.1</jsoup.version>
 		<icu4j.version>54.1.1</icu4j.version>
 		<xerces.version>2.11.0</xerces.version>
+		<httpclient.version>4.4.1</httpclient.version>
```
GuiForget: Bleeding edge :)

core/src/main/java/com/digitalpebble/storm/crawler/protocol/httpclient/HttpProtocol.java
```diff
@@ -0,0 +1,210 @@
...
+
+    private static final org.slf4j.Logger LOG = LoggerFactory
+            .getLogger(HttpProtocol.class);
+
+    private final static PoolingHttpClientConnectionManager CONNECTION_MANAGER = new PoolingHttpClientConnectionManager();
+    static {
+        // Increase max total connection to 200
+        CONNECTION_MANAGER.setMaxTotal(200);
+        // Increase default max connection per route to 20
+        CONNECTION_MANAGER.setDefaultMaxPerRoute(20);
```
GuiForget: These should be configurable

core/src/main/java/com/digitalpebble/storm/crawler/protocol/httpclient/HttpProtocol.java
```diff
@@ -0,0 +1,210 @@
...
+
+        HttpClientBuilder builder = HttpClients.custom()
+                .setUserAgent(userAgent)
+                .setConnectionManager(CONNECTION_MANAGER);
+
+        // use a proxy?
+        if (useProxy) {
+            HttpHost proxy = new HttpHost(proxyHost, proxyPort);
+            DefaultProxyRoutePlanner routePlanner = new DefaultProxyRoutePlanner(
+                    proxy);
```
GuiForget: Can't that be shared and therefore constructed once in configure ?

core/src/main/java/com/digitalpebble/storm/crawler/protocol/httpclient/HttpProtocol.java
```diff
@@ -0,0 +1,210 @@
...
+            builder.setRoutePlanner(routePlanner);
+        }
+
+        LOG.debug("HTTP connection manager stats "
+                + CONNECTION_MANAGER.getTotalStats().toString());
```
GuiForget: Use place holder instead and don't call toString()

core/src/main/java/com/digitalpebble/storm/crawler/protocol/httpclient/HttpProtocol.java
```diff
@@ -0,0 +1,210 @@
...
+
+        CloseableHttpClient httpclient = builder.build();
```
GuiForget: Use a `try-with-resource` to ensure it gets closed

core/src/main/java/com/digitalpebble/storm/crawler/protocol/httpclient/HttpProtocol.java
```diff
@@ -0,0 +1,210 @@
...
+
+        RequestConfig requestConfig = RequestConfig.custom()
+                .setSocketTimeout(timeout).setConnectTimeout(timeout).build();
```
GuiForget: Same here, could that be shared across invocations

core/src/main/java/com/digitalpebble/storm/crawler/protocol/httpclient/HttpProtocol.java
```diff
@@ -0,0 +1,210 @@
...
+                httpget.addHeader("If-Modified-Since", ifModifiedSince);
+            }
+
+            String ifNoneMatch = md.getFirstValue("cachedEtag");
+            if (StringUtils.isNotBlank(ifNoneMatch)) {
+                httpget.addHeader("If-None-Match", ifNoneMatch);
+            }
+        }
+
+        httpget.setConfig(requestConfig);
```
GuiForget: For readability this could be moved to line 119

core/src/main/java/com/digitalpebble/storm/crawler/protocol/httpclient/HttpProtocol.java
```diff
@@ -0,0 +1,210 @@
...
+                ConfUtils.getString(conf, "http.agent.url"),
+                ConfUtils.getString(conf, "http.agent.email"));
+
+        this.responseTime = ConfUtils.getBoolean(conf,
+                "http.store.responsetime", true);
+
+        this.skipRobots = ConfUtils.getBoolean(conf, "http.skip.robots", false);
+
+        robots = new HttpRobotRulesParser(conf);
+    }
```
GuiForget: IMO `configure` should be defined as the first method. More logical that way

core/src/main/java/com/digitalpebble/storm/crawler/protocol/httpclient/HttpProtocol.java
jnioche: sure, bit of a pain as we're static here and the conf would come from configure() which is not. Let's not aim for perfection straight away and do that in the next version ;-) the default values should be fine for most cases

core/src/main/java/com/digitalpebble/storm/crawler/protocol/httpclient/HttpProtocol.java
jnioche: yes could do

core/src/main/java/com/digitalpebble/storm/crawler/protocol/httpclient/HttpProtocol.java
jnioche: will change that

core/src/main/java/com/digitalpebble/storm/crawler/protocol/httpclient/HttpProtocol.java
jnioche: no need to. it is taken care of, see comment line 138

core/src/main/java/com/digitalpebble/storm/crawler/protocol/httpclient/HttpProtocol.java
jnioche: yep, could handle this in configure()

core/src/main/java/com/digitalpebble/storm/crawler/protocol/httpclient/HttpProtocol.java
jnioche: you mean the comment?

core/src/main/java/com/digitalpebble/storm/crawler/protocol/httpclient/HttpProtocol.java
jnioche: will do. thanks for your comments Gui.

jnioche:
@GuiForget done. Please have a look and merge if you're happy. Thanks!

core/src/main/java/com/digitalpebble/storm/crawler/protocol/AbstractHttpProtocol.java
```diff
@@ -0,0 +1,57 @@
...
+        StringBuffer buf = new StringBuffer();
+
+        buf.append(agentName);
+        if (agentVersion != null) {
+            buf.append("/");
+            buf.append(agentVersion);
+        }
+        if (((agentDesc != null) && (agentDesc.length() != 0))
+                || ((agentEmail != null) && (agentEmail.length() != 0))
+                || ((agentURL != null) && (agentURL.length() != 0))) {
```
GuiForget: Could have used StringUtils.isNotBlank or isNotEmpty

core/src/main/java/com/digitalpebble/storm/crawler/protocol/httpclient/HttpProtocol.java
```diff
@@ -0,0 +1,198 @@
...
+    @Override
+    public ProtocolResponse getProtocolOutput(String url, Metadata md)
+            throws Exception {
+
+        LOG.debug("HTTP connection manager stats {}",
+                CONNECTION_MANAGER.getTotalStats());
+
+        // no need to release the connection explicitly as this is handled
+        // automatically
+        CloseableHttpClient httpclient = builder.build();
```
GuiForget: I don't understand. Who closes it?

jnioche: CloseableHttpClient itself. See [https://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/client/CloseableHttpClient.html#execute(org.apache.http.client.methods.HttpUriRequest,%20org.apache.http.client.ResponseHandler)]

core/src/main/java/com/digitalpebble/storm/crawler/protocol/AbstractHttpProtocol.java
jnioche: You are right. Typical case of CTRL-C CTRL-V laziness. Will fix that.

core/src/main/java/com/digitalpebble/storm/crawler/protocol/httpclient/HttpProtocol.java
GuiForget: The connection is released. Yes. But the client isn't closed, possibly hanging on resources that could be freed faster

jnioche: Isn't that taken care of by the PoolingHttpClientConnectionManager?

jnioche: The manager should determine whether to keep the connection open and reuse it in a subsequent call e.g. if both the server and the client config allow it. Which is another advantage compared to the existing http implementation we have.

GuiForget: If I'm not mistaken, you're confusing the connections and the client. The connection manager manages the connection. But you need to close the client at the end of its use.

jnioche:
you are right about the client. silly me!

squashed commits

thanks for reviewing it

PR #125, opened 2015-04-23
Adding MaxDepthFilter class

dang-sugarinc:
This filter will return null if the depth of the URL is > the maxDepth 
parameter. Not setting the maxDepth or setting it to 0 means no limit.
If the depth is not tracked, this filter will do nothing.

### DIFF
--- a/core/src/main/java/com/digitalpebble/storm/crawler/filtering/depth/MaxDepthFilter.java
+++ b/core/src/main/java/com/digitalpebble/storm/crawler/filtering/depth/MaxDepthFilter.java
@@ -1,0 +1,52 @@
+package com.digitalpebble.storm.crawler.filtering.depth;
+
+import java.net.URL;
+import java.util.Map;
+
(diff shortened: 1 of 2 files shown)


dang-sugarinc:
@GuiForget Please comment.

core/src/main/java/com/digitalpebble/storm/crawler/filtering/depth/MaxDepthFilter.java
```diff
@@ -0,0 +1,52 @@
+package com.digitalpebble.storm.crawler.filtering.depth;
```
GuiForget: @jnioche the package name are getting really long :)

jnioche: indeed. Eclipse is pretty good at suggesting / autocompleting so it's not like we have to type it very often

PR #127, opened 2015-04-25
Added bytes fetched and fetch time metrics

jakekdodd:
@jnioche these are useful for monitoring, any thoughts? Would be happy to add some additional refactoring for the commit...several low-hanging fruit where the metrics could be improved

### DIFF
--- a/core/src/main/java/com/digitalpebble/storm/crawler/bolt/FetcherBolt.java
+++ b/core/src/main/java/com/digitalpebble/storm/crawler/bolt/FetcherBolt.java
@@ -671,7 +671,7 @@ public class FetcherBolt extends BaseRichBolt {
         this.metricGauge = context.registerMetric("fetcher",
                 new MultiCountMetric(), 10);
 
-        this.averagedMetrics = context.registerMetric("fetcher_reduced",
+        this.averagedMetrics = context.registerMetric("fetcher_average",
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/digitalpebble/storm/crawler/bolt/FetcherBolt.java
```diff
@@ -664,6 +671,9 @@ public void prepare(Map stormConf, TopologyContext context,
         this.metricGauge = context.registerMetric("fetcher",
                 new MultiCountMetric(), 10);
 
+        this.averagedMetrics = context.registerMetric("fetcher_reduced",
```
jnioche: what about 'fetcher_average' instead of 'fetcher_reduced'?

jakekdodd: Yeah that's fine with me. I doubt we'll be using any reducer other than `MeanReducer`, so that should be just as accurate

jakekdodd:
Changed name to "fetcher_average" and added the metrics to `SimpleFetcherBolt`. I'm ready to merge--any more comments @DigitalPebble/committers-crawler?

jnioche:
Great thanks Jake

PR #129, opened 2015-04-29
Catch exceptions in ParseFilters.filter() method

jakekdodd:
I recently ran into an issue where a single failing ParseFilter would break the entire filtering chain. This is undesirable because it causes a URL to be sent down the Status stream with `Status.ERROR`, due to nothing but a potentially malformed ParseFilter.

@DigitalPebble/committers-crawler can you see any reason why it might be a bad idea to catch and report, but otherwise ignore, exceptions in the `ParseFilters.filter()` method?

### DIFF
--- a/core/src/main/java/com/digitalpebble/storm/crawler/parse/ParseFilters.java
+++ b/core/src/main/java/com/digitalpebble/storm/crawler/parse/ParseFilters.java
@@ -50,7 +50,7 @@ public class ParseFilters implements ParseFilter {
 
     /**
      * loads the filters from a JSON configuration file
-     * 
+     *
      * @throws IOException
      */
     public ParseFilters(Map stormConf, String configFile) throws IOException {
@@ -156,7 +156,11 @@ public class ParseFilters implements ParseFilter {
             Metadata metadata, List<Outlink> outlinks) {
         for (ParseFilter filter : filters) {
             long start = System.currentTimeMillis();
-            filter.filter(URL, content, doc, metadata, outlinks);
+            try {
+                filter.filter(URL, content, doc, metadata, outlinks);
+            } catch (Exception e) {
+                LOG.error("ParseFilter {} failed due to {}", filter.getClass(), e);
+            }
             long end = System.currentTimeMillis();
             LOG.debug("ParseFilter {} took {} msec", filter.getClass()
                     .getName(), (end - start));


jakekdodd:
I actually don't like this idea =) Masking errors like this isn't good, and kind of defeats the purpose of `Status.ERROR`

jnioche:
I think it is preferable to know that a parser needs fixing. It would probably go unnoticed otherwise

jakekdodd:
Agreed, although it still isn't obvious what the problem is unless you inspect the error message that's being passed in the metadata down the status stream.

I don't think the solution in this PR is the right one, but I still think we can improve this. It feels weird to me, for example, that `SiteMapParserBolt` can successfully parse a sitemap --> fail to apply the `ParseFilters` because one filter (in a chain of many that may complete successfully) can't handle XML --> emit `Status.ERROR`, ack, and return, breaking execution of the topology without ever emitting the outlinks.

I guess we could leave it to individual `ParseFilters` to decide whether to gracefully fail, or blow up execution of the rest of the topology.

jnioche:
> it still isn't obvious what the problem is unless you inspect the error message that's being passed in the metadata down the status stream.

you are right. we should at least log the exception as you did in [https://github.com/DigitalPebble/storm-crawler/pull/129/files#diff-69987c422bb8076c1d792c7575d69536R161] even if it is re-thrown to blow things up... I thought we did that already.

> I guess we could leave it to individual ParseFilters to decide whether to gracefully fail, or blow up execution of the rest of the topology.

and/or have a global config and use it in ParseFilters.java?

jakekdodd:
> and/or have a global config and use it in ParseFilters.java?

Yeah I was thinking that's an option. If we do that, maybe we should think about returning some type of report from the filter method? That way a bolt using the parse filters could fine-tune the filters for which it'll tolerate failure. Or instead of a global config, we could add a new method signature for `ParseFilters.filter` with a boolean indicating whether or not to tolerate failures

jnioche:
> If we do that, maybe we should think about returning some type of report from the filter method? That way a bolt using the parse filters could fine-tune the filters for which it'll tolerate failure. 

let's keep it simple. My gut feeling on this would be to leave it as is so that a dodgy parse filter crashes the process but if we want to be lenient then using the logs would be sufficient. Let's not over-engineer this thing.

> Or instead of a global config, we could add a new method signature for ParseFilters.filter with a boolean indicating whether or not to tolerate failures

hmm. I'd rather leave the signatures as is but be able to override the global config on a per filter level by having a parameter that in the JSON config.

Speaking of which, I am beginning to wonder whether it would make more sense to have ParseFilter as an abstract class and not an interface. We could implement some default behaviour for some of the methods + handle the param mentioned above there. Also, we do define names in the config of the [ParseFilters](https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/java/com/digitalpebble/storm/crawler/parse/ParseFilters.java#L95) and I was thinking that it would be nice for the ParseFilter implementations to have access to the names associated with them. Doing that via the abstract super class would be an obvious way.

WDYT?

kkrugler:
In general I favor abstract classes over interfaces - it's easier to do the things you mention, and it's easier to revise the API without breaking every implementation.

Issue #117, opened 2015-04-13
Parse could generate output for more than one document 


jnioche:
At the moment we get a 1:1 relationship between the input and output of a parsing step. We could have a 1:N instead and more than one document in the output.

This is done in Nutch 1.x where the [ParseResult class](https://github.com/apache/nutch/blob/trunk/src/java/org/apache/nutch/parse/ParseResult.java) is a `Map<String,Parse>` with the [Parse](https://github.com/apache/nutch/blob/trunk/src/java/org/apache/nutch/parse/Parse.java) object containing the text and metadata for a URL.

This is useful for instance when a document can be split into subdocuments e.g. posts in a forum thread and we can't or don't want to fetch the subdocuments individually.

GuiForget:
:+1:  from me as we have product pages with multiple products on it

PR #135, opened 2015-06-04
Allowing the parse output to output more than one document

jorgelbg:
This address the https://github.com/DigitalPebble/storm-crawler/issues/117 issue at some extent this also encapsulate in a `ParseResult` the parsing data/metadata of a URL. Each URL has only one `ParseResult` which at the same time has at least 1 `ParseData` instance to hold the data of the "parent" URL in case subdocuments are extracted in any `ParseFilter` the URLs and ParseData of each subdocument is added to the parent ParseResult. So the `ParserBolt` emits each `ParseData` container in the `ParseResult` as a tuple. So at the very least 1 `ParseData`, or any number if subdocuments are extracted, also this implies that there is no difference between the extracted information about the parent URL or any subdocument each `ParseData` gets emitted as a tuple.

I made some simple graph to try to explain this :) I hope it makes things a little more clearer than my explanation.
storm-parsing

### DIFF
--- a/core/src/main/java/com/digitalpebble/storm/crawler/parse/ParseData.java
+++ b/core/src/main/java/com/digitalpebble/storm/crawler/parse/ParseData.java
@@ -1,0 +1,76 @@
+/**
+ * Licensed to DigitalPebble Ltd under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * DigitalPebble licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.digitalpebble.storm.crawler.parse;
+
+import com.digitalpebble.storm.crawler.Metadata;
+
+public class ParseData {
+    private byte[] content;
+    private String text;
+    private Metadata metadata;
+
+    public ParseData() {
+        this.metadata = new Metadata();
+    }
+
+    public ParseData(String text, Metadata metadata) {
+        this.text = text;
+        this.metadata = metadata;
+        this.content = new byte[0];
+    }
+
+    public ParseData(Metadata md) {
+        this.metadata = md;
+    }
+
+    public String getText() {
+        return text;
+    }
+
+    public Metadata getMetadata() {
+        return metadata;
+    }
+
+    public void setText(String text) {
+        this.text = text;
+    }
+
+    public void setMetadata(Metadata metadata) {
+        this.metadata = metadata;
+    }
+
+    public byte[] getContent() {
+        return content;
+    }
+
+    public void setContent(byte[] content) {
+        this.content = content;
+    }
+
+    public void put(String key, String value) {
+        metadata.addValue(key, value);
+    }
+
+    public String get(String key) {
+        return metadata.getFirstValue(key);
+    }
+
+    public String[] getValues(String key) {
+        return metadata.getValues(key);
+    }
+}
--- a/core/src/main/java/com/digitalpebble/storm/crawler/parse/ParseResult.java
+++ b/core/src/main/java/com/digitalpebble/storm/crawler/parse/ParseResult.java
@@ -1,0 +1,123 @@
+/**
+ * Licensed to DigitalPebble Ltd under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * DigitalPebble licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.digitalpebble.storm.crawler.parse;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+import com.digitalpebble.storm.crawler.Metadata;
+
+public class ParseResult implements Iterable<Map.Entry<String, ParseData>> {
+
+    private List<Outlink> outlinks;
+    private Map<String, ParseData> parseMap;
+
+    public ParseResult() {
+        parseMap = new HashMap<>();
+        outlinks = new ArrayList<>();
+    }
+
+    public ParseResult(Map<String, ParseData> map) {
+        if (map == null) {
+            throw new NullPointerException();
+        }
+
+        parseMap = map;
+        outlinks = new ArrayList<>();
+    }
+
+    public boolean isEmpty() {
+        return parseMap.isEmpty();
+    }
+
+    public int size() {
+        return parseMap.size();
+    }
+
+    public List<Outlink> getOutlinks() {
+        return outlinks;
+    }
+
+    public void setOutlinks(List<Outlink> outlinks) {
+        this.outlinks = outlinks;
+    }
+
+    /**
+     * @return An existent instance of Parse for the given URL or an empty one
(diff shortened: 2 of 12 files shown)


core/src/main/java/com/digitalpebble/storm/crawler/spout/RandomURLSpout.java
```diff
@@ -42,9 +42,7 @@
     private boolean active = true;
     private boolean removeAfterSending = true;
 
-    private String[] urls = new String[] { "http://www.lequipe.fr/",
-            "http://www.lemonde.fr/", "http://www.bbc.co.uk/",
-            "http://www.facebook.com/", "http://www.rmc.fr" };
+    private String[] urls = new String[] { "http://firefoxmania.uci.cu/"};
```
jnioche: this doesn't really need to be changed

core/src/main/java/com/digitalpebble/storm/crawler/parse/ParseResult.java
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        return parseInfo.getValues(key);
+    }
+
+    public void put(String URL, String key, String value) {
+        if (parseMap.containsKey(URL)) {
+            parseMap.get(URL).getMetadata().addValue(key, value);
+        } else {
+            ParseData parse = ParseData.empty;
```
jnioche: isn't 'ParseData.empty' immutable?

jnioche:
Thanks Jorge. Could you please apply our formatting to the code?  It can be found at [https://github.com/DigitalPebble/storm-crawler/blob/master/eclipse-formatting-profile.xml].

See [https://github.com/crawler-commons/crawler-commons/pull/76] for a way of doing it with Maven

core/src/main/java/com/digitalpebble/storm/crawler/spout/RandomURLSpout.java
jorgelbg: Yep this slipped through my fingers, I modified to avoid the use a proxy to do the tests, I'll fix this.

core/src/main/java/com/digitalpebble/storm/crawler/parse/filter/SubDocumentsParseFilter.java
```diff
@@ -0,0 +1,41 @@
+package com.digitalpebble.storm.crawler.parse.filter;
```
jnioche: This file is missing the license header. It is also not very useful in practice but is a good illustration of use.
Instead, what about adding it to the [parse test package](https://github.com/DigitalPebble/storm-crawler/tree/master/core/src/test/java/com/digitalpebble/storm/crawler/parse/filter) and use that for checking that we are getting what we expect?

jnioche:
@GuiForget @jakekdodd do you want to give this a try? It would be useful for processing pages containing multiple products

core/src/main/java/com/digitalpebble/storm/crawler/parse/ParseData.java
```diff
@@ -0,0 +1,78 @@
...
+package com.digitalpebble.storm.crawler.parse;
+
+import com.digitalpebble.storm.crawler.Metadata;
+
+public class ParseData {
+    private byte[] content;
+    private String text;
+    private Metadata metadata;
+
+    public static final ParseData empty = new ParseData(Metadata.empty);
```
jnioche: this does not seem to be used anywhere. Remove?

core/src/main/java/com/digitalpebble/storm/crawler/parse/ParseResult.java
```diff
@@ -0,0 +1,129 @@
...
+
+package com.digitalpebble.storm.crawler.parse;
+
+import java.util.*;
+
+import com.digitalpebble.storm.crawler.Metadata;
+
+public class ParseResult implements Iterable<Map.Entry<String, ParseData>> {
+
+    private String URL;
```
jnioche: never used?

core/src/main/java/com/digitalpebble/storm/crawler/parse/ParseResult.java
```diff
@@ -0,0 +1,129 @@
...
+    private List<Outlink> outlinks;
+    private Map<String, ParseData> parseMap;
+
+    public static final ParseResult empty = new ParseResult(
```
jnioche: this does not seem to be used anywhere. Remove?

GuiForget:
I'm a bit confused by the structure of ParseResult. ParseResults holds multiple ParseData. So far so good. But what is the key in the map? I see it's a URL but in which case would the URL be different from the one of the containing ParseResult?

jorgelbg:
@GuiForget The idea behind the map is that each subdocument extracted from the parent page will have it's own identifying URL which could be a totally different URL (pointing to a detailed product view, for instance) or a bookmark in the same parent page, etc.

core/src/main/java/com/digitalpebble/storm/crawler/parse/ParseData.java
jorgelbg: Kept this as a standard way to reference an empty ParseData/ParseResult, right now is not used and could be easily removed.

core/src/main/java/com/digitalpebble/storm/crawler/parse/ParseResult.java
jorgelbg: Doesn't have any real use for the moment, but I think of this attribute as a way of keeping a record of the original "parent" URL, the `ParseFilter` receive the original URL as a parameter, I can't think of a particular use case needing this, remove it then?

jnioche:
@GuiForget Imagine for instance that we process an extended RSS feed containing all the info we need for a document. The subdocuments would have their own individual URL and with this mechanism we could bypass fetching + parsing them  altogether.
For subdocs which do not have their own URLs, a solution would be to generate a dummy one e.g. with an anchor tag.

core/src/main/java/com/digitalpebble/storm/crawler/parse/ParseData.java
jnioche: if it is not used, then let's remove it

core/src/main/java/com/digitalpebble/storm/crawler/parse/ParseResult.java
jnioche: Let's remove it

jakekdodd:
Architecturally, I wouldn't emit each subdocument from the parser bolt. Maybe instead emit the whole `ParseResult` as a field in the default tuple? That way, a downstream bolt can decide what to do with the results--which might include emitting the subdocuments for individual processing, or it might not.

jnioche:
Hi @jakekdodd, thanks for taking the time to comment.

> Maybe instead emit the whole ParseResult as a field in the default tuple?

That would mean having another brand new serializable object to do that and more importantly a substantial change to the contract between bolts.

I expect most users would not use this feature - from my Nutch experience very few people did. I'd rather keep things as simple as possible.

> That way, a downstream bolt can decide what to do with the results--which might include emitting the subdocuments for individual processing, or it might not.

You could implement that logic within a subsequent parse filter maybe? Or have a downstream bolt to decide what to do based on the content of the individual tuple?

jorgelbg:
I was thinking of rewriting the `SubDocumentsParseFilter` and the test around the RSS example, parsing each `url` into a separated document, I think this could be show in a more useful way how to do a Job, right now is a very dummy example. What do you think @jnioche ?

jnioche:
@jorgelbg yes, good idea

jnioche:
Another advantage of this is that ParseFilters can modify the text extracted for the main document. One application of this would be to have a PF taking an XPath as parameter in order to restrict the text indexed to a given HTML element. Something that Nutch users ask all the time, would be a lot more elegant the way it's done here.

@jakekdodd @GuiForget are you happy for me to merge this?

jorgelbg:
@jnioche Taking into account that digitalpebble.com doesn't have a RSS feed can we use a similar to the  sitemap? the general idea is to index each `url` element of the sitemaps.xml file as a separated document. This will keep examples concise around the digitalpebble.com info and also easy to test.

jorgelbg:
@jnioche I've committed an updated version of the `SubDocumentsParseFilter` and the test, working with the sitemap.xml file take a look at it.

core/src/test/java/com/digitalpebble/storm/crawler/parse/filter/SubDocumentsParseFilter.java
```diff
@@ -0,0 +1,90 @@
...
+                ParseData parseData = parse.get(child.getTextContent());
+
+                NodeList childs = node.getChildNodes();
+                for (int j = 0; j < childs.getLength(); j++) {
+                    Node n = childs.item(j);
+                    parseData.put(n.getNodeName(), n.getTextContent());
+                }
+            }
+        } catch (Exception e) {
+            e.printStackTrace();
```
jnioche: yuk. slf4j instead

jnioche:
Looks good. Could you please just fix the issue with the ` e.printStackTrace();` above then squash your commits into a single one?
Thanks

jnioche:
Thanks Jorge!

PR #143, opened 2015-06-17
Improvement to SitemapParserBolt

jnioche:
* added option to sniff content when triggering K/V is missing
* let the underlying sitemapparser guess the mimetype when it is not available


GuiForget:
Looks good!

jnioche:
Thanks Gui

Issue #140, opened 2015-06-08
Organise external content as separate sub-modules


jnioche:
This is something that we had mentioned before in #41.  We currently have everything in external with a whole bunch of dependencies. Instead we should have a think about splitting into separate sub-modules e.g.
* elasticsearch
* tika
* metrics
* solr (see #134)

which is what is done in [storm](https://github.com/apache/storm/blob/master/pom.xml#L15).

An advantage of doing this would be that each module could have its own set of [scripts](https://github.com/DigitalPebble/storm-crawler/blob/master/external/ES_IndexInit.sh) or other resources instead of everything being put in one place.

PR #145, opened 2015-06-18
Organise external content as separate sub-modules

jnioche:
Implements #140 

This splits the various external resources into submodules. 

The only niggle is that I've had to duplicate the formatting file for Eclipse so that the sub-modules *and* the core module are able to access it. [http://code.revelc.net/formatter-maven-plugin/examples.html#Multimodule_Configuration] might be a way of solving it but this is no big deal and could be done later.


jnioche:
Ok, haven't heard back from anyone so I'll take that as a OK.

Issue #147, opened 2015-06-18
Enforce robots meta tags 


jnioche:
See [http://www.google.com/support/enterprise/static/gsa/docs/admin/70/gsa_doc_set/admin_crawl/preparing.html#1076168]

`<meta name="robots" content="noarchive, nofollow"/>`

as well as  `rel="nofollow"` attribute  
`<a href="test1.html" rel="nofollow>no follow</a>`

jnioche:
Nutch equivalent

[https://github.com/apache/nutch/blob/trunk/src/plugin/parse-html/src/java/org/apache/nutch/parse/html/HTMLMetaProcessor.java]

and 

[https://github.com/apache/nutch/blob/trunk/src/java/org/apache/nutch/parse/HTMLMetaTags.java]

rkrombho:
I would really prefer if this behavior could be switched of and on with a configuration flip.
`rel=nofollow` should only be used for ranking links. Crawlers typically still follow those links but remove them from any sort of rank calculation.

Here is what the big search engine cralwers do with it:
https://en.wikipedia.org/wiki/Nofollow#Interpretation_by_the_individual_search_engines

Especially for whole web crawls it may still be important for crawlers to follow those links in order to find new pages irrespective of ranking those type of links.

jnioche:
Thanks for your comments @rkrombho 

There seems to be different interpretations of what this is supposed to do indeed. The [doc for GSA](http://www.google.com/support/enterprise/static/gsa/docs/admin/70/gsa_doc_set/admin_crawl/preparing.html#1076168) which I mentioned previously implies that the links are not followed at all ("_The search appliance crawler retrieves and archives the document in the search appliance cache, but does not follow links on the Web page to other documents_").

Your suggestion of a config switch makes sense. What about having something like `robots.noFollow.strict` with a boolean value, where the outlinks would be filtered out completely when the mode is set to strict and left otherwise?

BTW you could always handle the outlinks in anyway you want if you extracted them from the DOM in a bespoke ParseFilter.

PR #148, opened 2015-06-19
Enforce robots meta instructions

jnioche:
Implements #147 

* Added RobotsTags to represent the instructions provided in the HTTP headers or META tags.
* JSoupParserBolt bypasses the extraction of outlinks and AbstractIndexerBolt filters out document according to these values.
* JSoupParserBolt filters out links with `rel="nofollow"`
* Added RobotsTagsTest and JSoupParserBoltTest.

Will add the same behaviour to the Tika parser after merging #145 


jnioche:
Have added a new config `robots.noFollow.strict`  following the conversation in #147 
If set to false the outlinks are not removed from a page but their anchors are not tracked. Ideally we'd want the source URL not to be tracked as well but given the way we currently do it this is proving a bit tricky. We can maybe add that later, in most cases people would use the strict mode.

@rkrombho @DigitalPebble/committers-crawler thoughts anyone?

PR #151, opened 2015-06-26
JSoupParserBolt does a basic check of the content type

jnioche:
+ refactored handling of exceptions

We can improve the checking mechanism later but at least it will prevent non-html documents to be parsed with JSOUP.


jnioche:
note to self \: Tika based parser should use the same error handling mechanism instead of duplicating  code

PR #152, opened 2015-06-29
Adding support for [SOLR] Write Spout, StatusUpdaterBolt and Indexer

jorgelbg:
Solr resources for StormCrawler, this add support for:
* IndexerBolt
* MetricsConsumer
* SolrSpout/StatusUpdaterBolt
* SolrConnection
* SolrCrawlTopology

The `SolrConnection` class have support for both the SolrCloud and a single Solr server by the ConcurrentUpdateSolrClient (CUSC). For using SolrCloud a `solr.TYPE.zkhost` needs to be specified instead of the `solr.TYPE.url`, also an additional parameter `solr.TYPE.collection` needs to be provided for the SolrCloud. I've detected a bug in the solr-solrj library regarding the use of the collection parameter in some methods (https://issues.apache.org/jira/browse/SOLR-7729), once this is fixed should simplify the code for the CUSC case, for now I recommend using the `url` parameter for the CUSC case to point directly to the solr collection that wants to be used(`external-conf.yaml`).

The `solr.metrics.commit.size` config option is basically a batch size to issue a hard commit against the server, this is not required (as is initialized in 250 by default) and could be leveraged by configuring autocommit in the server. 

The `MetricsConsumer` has support for automatic Document Expiration (TTL), a couple of additional parameters are provided in this class: `solr.metrics.ttl` and `solr.metrics.ttl.field` for configuring the Date language expression in the first case and the target field in the later. Of course this also needs some configuration in the Solr side. The date expression used in `solr.metrics.ttl` is not validated in the code. 

For `SolrCrawlTopology` to work at least an URL in the status collection is required to be present.

### DIFF
--- a/external/solr/external-conf.yaml
+++ b/external/solr/external-conf.yaml
@@ -1,0 +1,24 @@
+# Solr indexer bolt
+solr.indexer.url: "http://localhost:8983/solr/collection1"
+solr.indexer.threads: 1
+solr.indexer.queue.size: 100
+solr.indexer.commit.size: 1
+
+# Solr spout and persistence bolt
+solr.status.url: "http://localhost:8983/solr/status"
+solr.status.threads: 1
+solr.status.queue.size: 100
+solr.status.commit.size: 1
+
+# Solr MetricsConsumer
+solr.metrics.url: "http://localhost:8983/solr/metrics"
+solr.metrics.threads: 1
+solr.metrics.queue.size: 100
+solr.metrics.commit.size: 1
+
+# For SolrCloud, this settings instead of solr.indexer.url
+#
+#   solr.indexer.zkhost: "http://localhost:9983/"
+#   solr.indexer.collection: collection1
+#
+# the same applies for the spout/persistence bolt and the metricsconsumer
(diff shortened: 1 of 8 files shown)


jnioche:
Thanks, have merged with a few modifications. Mostly added a separate topo to inject from local seed files into a SOLR index.

BTW please add yourself to the main pom in the list of committers.

Could you also add a README.md to external/solr to describe what it contains and how to use it?

re-SOLRSpout \: would it make sense to use *field collapsing* on a configurable field name so that we could guarantee a good mix of URLs per domain?

jorgelbg:
:+1:  on the separated topology its more useful this way. I was thinking of adding a script that add some URLs, perhaps the same URLs as the RandomURLSpout for testing, but this is indeed more practical. 

Done, modifying the pom file :)

I'm going to add the README.md file with the explanation. 

I think that using  _field collapsing_ can be a nice add, in this way we kind of guarantee _diversity_ in the seed links, in this case we should add a parameter for configuring the field and the amount of URLs per bucket too.

jorgelbg:
:+1:  on the separated topology its more useful this way. I was thinking of adding a script that add some URLs, perhaps the same URLs as the RandomURLSpout for testing, but this is indeed more practical. 

Done, modifying the pom file :)

I'm going to add the README.md file with the explanation. 

I think that using  _field collapsing_ can be a nice add, in this way we kind of guarantee _diversity_ in the seed links, in this case we should add a parameter for configuring the field and the amount of URLs per bucket too?

jnioche:
field collapsing \: yes we'd need a parameter for the field and number of URLs. Let's track that as a separate PR.

PR #153, opened 2015-06-29
Expand XPathFilter to accept a list of expressions as an argument.

dang-sugarinc:


jnioche:
Thanks @dang-sugarinc. Could you modify [https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/resources/parsefilters.json] to illustrate this change?

Did you apply the formatting with `mvn java-formatter:format`?

dang-sugarinc:
@jnioche Changes made. Let me know if there's anything else I missed.

jnioche:
Thanks @dang-sugarinc

Issue #155, opened 2015-07-02
[SOLR] add README file


jnioche:
as mentioned in #152 we need a README file in external/solr to describe what we have there and how to use it e.g create and configure basic indices, run the topologies etc...

jnioche:
Merged in #157

PR #157, opened 2015-07-08
Adding a README file to the solr module

jorgelbg:
A basic README file containing the provided resources, how to use them, and the different configuration parameters available. Issue https://github.com/DigitalPebble/storm-crawler/issues/155

@jnioche Can you take a look?

### DIFF
--- a/external/solr/README.md
+++ b/external/solr/README.md
@@ -1,0 +1,71 @@
+storm-crawler-solr
+==================
+
+Set of Solr resources for StormCrawler that will allow you to create topologies that consume from a Solr collection and store metrics, status or parsed content into Solr.
+
+## How to use
+
+In your project you can use this by adding the following dependency:
+
+```xml
+<dependency>
+    <groupId>com.digitalpebble</groupId>
+    <artifactId>storm-crawler-solr</artifactId>
+    <version>0.6</version>
+</dependency>
+```
+
+## Available resources
+
+* `IndexerBolt`: Implementation of `AbstractIndexerBolt` that allows to index the parsed data and metadata into a specified Solr collection.
+
+* `MetricsConsumer`: Class that allows to store Storm metrics in Solr.
+
+* `SolrSpout`: Spout that allows to get the initial URLs from a specified Solr collection.
+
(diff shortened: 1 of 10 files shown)


jnioche:
Looks great! Don't we need to predefine the schema for the various indices in SOLR? or can it guess like ES does?

jorgelbg:
This depends on the configuration I think we can provide an example schema.xml file, the one mandatory field should be the url field, the rest is configurable through the topology config file. For testing I used an example configuration that except the mandatory field (url for the IndexerBolt and the StatusUpdaterBolt) it can store/index everything else that you throw at it, using dynamic fields. The downside is that everything is stored as a string, not so useful but handy for testing purposes.

This could work at least as an example, basically I use something like https://github.com/arafalov/simplest-solr-config with the minimum configuration possible, very generic.

I see some options here: 
* Add this default configuration to the module so that can be used directly and explain the basic requirements in the doc.
* Use a more specific schema.xml file (at least with the default fields for each collection in the indexer, metrics, status) that could be attached or included in the README, that it includes the fields extracted by default by the `JsoupParserBolt`. 
* Use the schemaless mode and reference in the README to something like https://cwiki.apache.org/confluence/display/solr/Schemaless+Mode.

Any preference on one particular option?

jnioche:
> Add this default configuration to the module so that can be used directly and explain the basic requirements in the doc.

+1 & mention the schemaless mode in the README as well.

Thanks

jorgelbg:
Ok, I'm going to commit the 3 collections config into this same PR.

jnioche:
Hi @jorgelbg. Could you update the README to mention the schemaless mode and explain the SOLR configs you just added?

jorgelbg:
@jnioche I've added a section "Solr configuration" in the last commit https://github.com/jorgelbg/storm-crawler/commit/8f6f8c26207a73e8bec721c195c9bccb3351bd1e#diff-172fd82a13532797e7558b92543c817aR73 should I add a more comprehensive explanation?

jnioche:
Sorry, had missed that bit when looking at the commits this morning. Thanks!

Issue #149, opened 2015-06-24
Indexer bolts use canonical URL 


jnioche:
We already have a parse filter that extracts canonical URLs. We should add to AbstractIndexerBolt the ability to use them instead of the URL when available

jorgelbg:
Adding this in the `AbstractIndexerBolt` means that it should be then optional to use in the indexers? perhaps creating and additional method that returns the right URL based on the Tuple received by the bolt, something like:

```java
    protected String valueForURL(Tuple tuple) {
        String url = tuple.getStringByField("url");
        Metadata metadata = (Metadata) tuple.getValueByField("metadata");

        if (metadata.getFirstValue("canonical") != null) 
            url = metadata.getFirstValue("canonical");

        return url;
    }
```

should work but this must be used instead of the default `url` var extracted from the Tuple in the execute method. This means that we should change this in the existing indexer implementations (ES/Solr/StdOut). Perhaps something like:

```java
String url = valueForURL(tuple);
...
```

Any thoughts on this?

jorgelbg:
Adding this in the `AbstractIndexerBolt` means that it should be then optional to use in the indexers? perhaps creating and additional method that returns the right URL based on the Tuple received by the bolt, something like:

```java
    protected String valueForURL(Tuple tuple) {
        String url = tuple.getStringByField("url");
        Metadata metadata = (Metadata) tuple.getValueByField("metadata");

        if (metadata.getFirstValue("canonical") != null) 
            url = metadata.getFirstValue("canonical");

        return url;
    }
```

should work but this must be used instead of the default `url` var extracted from the Tuple in the execute method. This means that we should change this in the existing indexer implementations (ES/Solr/StdOut). Perhaps something like:

```java
String url = valueForURL(tuple);
...
```

Any thoughts on this?

jnioche:
That's pretty much what I had in mind. Later on we could do the same as in Nutch and for redirections decide which of the source or target URLs is the nicest to display. 

We'd probably need to check that the value of canonical is OK e.g. it belongs to the same domain and make its value absolute if necessary.

jorgelbg:
I'm going to add the validation in the canonical URL and submit a PR with the change.

jnioche:
Implemented in #161

PR #161, opened 2015-07-16
Adding support for using the canonical URL in the IndexerBolts

jorgelbg:
This address the https://github.com/DigitalPebble/storm-crawler/issues/149 issue which provides a method in the `AbstractIndexerBolt` that returns the right URL to be indexed. In this case if a canonical URL is found its used instead of the original URL. Some sanity checks are done:
* The URL is resolved to avoid relative URLs
* The host of the canonical URL is compared with the host of the original URL and only used if is the same.

I've also updated the `StdOutIndexer` and the `IndexerBolt` of both Solr and Elasticsearch.

I wanted to test the logic in the new method so I've added a basic testing framework very similar to `com.digitalpebble.storm.crawler.parse` to easy the task. I've also created a `DummyIndexer` that is used to test the output, since the IndexerBolts are the end of the chain. I wrote tests for the canonical URL and also some scenarios for document and metadata filtering. This tests are independent of the parsing tests.

I don't think this is suitable for testing Solr or Elasticsearch modules, mainly because each of the `IndexerBolts` of this "framework" interacts with a Solr/ES backend (outside of Storm Crawler) and for this cases using something like solr-test-framework (for Solr) makes more sense. 

Would love some feedback about this.

### DIFF
--- a/core/src/main/java/com/digitalpebble/storm/crawler/indexing/AbstractIndexerBolt.java
+++ b/core/src/main/java/com/digitalpebble/storm/crawler/indexing/AbstractIndexerBolt.java
@@ -17,6 +17,8 @@
 
 package com.digitalpebble.storm.crawler.indexing;
 
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
@@ -26,20 +28,26 @@ import java.util.regex.Pattern;
 
 import org.apache.commons.lang.ArrayUtils;
 import org.apache.commons.lang.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import backtype.storm.task.OutputCollector;
 import backtype.storm.task.TopologyContext;
 import backtype.storm.topology.OutputFieldsDeclarer;
 import backtype.storm.topology.base.BaseRichBolt;
+import backtype.storm.tuple.Tuple;
 import clojure.lang.PersistentVector;
 
 import com.digitalpebble.storm.crawler.Metadata;
 import com.digitalpebble.storm.crawler.util.ConfUtils;
 import com.digitalpebble.storm.crawler.util.RobotsTags;
+import com.digitalpebble.storm.crawler.util.URLUtil;
 
 /** Abstract class to simplify writing IndexerBolts **/
 @SuppressWarnings("serial")
 public abstract class AbstractIndexerBolt extends BaseRichBolt {
+
+    private final Logger LOG = LoggerFactory.getLogger(getClass());
 
     /**
      * Mapping between metadata keys and field names for indexing Can be a list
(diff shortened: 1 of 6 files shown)


jorgelbg:
One thing that I forgot should we make this canonical behavior configurable i.e on/off switch in the config file?

kkrugler:
I would have thought that the more general URL normalization would have handled the case of relative URLs upstream of this bolt.

> From: Jorge Luis Betancourt Gonzalez
> Sent: July 15, 2015 10:41:25pm PDT
> To: DigitalPebble/storm-crawler
> Subject: [storm-crawler] Adding support for using the canonical URL in the IndexerBolts (#161)
> 
> This address the #149 issue which provides a method in the AbstractIndexerBolt that returns the right URL to be indexed. In this case if a canonical URL is found its used instead of the original URL. Some sanity checks are done:
> 
> The URL is resolved to avoid relative URLs
> The host of the canonical URL is compared with the host of the original URL and only used if is the same.
> I've also updated the StdOutIndexer and the IndexerBolt of both Solr and Elasticsearch.
> 
> I wanted to test the logic in the new method so I've added a basic testing framework very similar to com.digitalpebble.storm.crawler.parse to easy the task. I've also created a DummyIndexer that is used to test the output, since the IndexerBolts are the end of the chain. I wrote tests for the canonical URL and also some scenarios for document and metadata filtering. This tests are independent of the parsing tests.
> 
> I don't think this is suitable for testing Solr or Elasticsearch modules, mainly because each of the IndexerBolts of this "framework" interacts with a Solr/ES backend (outside of Storm Crawler) and for this cases using something like solr-test-framework (for Solr) makes more sense.
> 
> Would love some feedback about this.
> 
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/DigitalPebble/storm-crawler/pull/161
> 
> Commit Summary
> 
> Adding support for using the canonical URL if found instead of the original URL for indexing purpouses.
> Updating the Elasticsearch and Solr IndexerBolts to use the new method available.
> Removing some unnecesary comment
> File Changes
> 
> M core/src/main/java/com/digitalpebble/storm/crawler/indexing/AbstractIndexerBolt.java (32)
> M core/src/main/java/com/digitalpebble/storm/crawler/indexing/StdOutIndexer.java (2)
> A core/src/test/java/com/digitalpebble/storm/crawler/indexer/BasicIndexingTest.java (207)
> A core/src/test/java/com/digitalpebble/storm/crawler/indexer/DummyIndexer.java (90)
> A core/src/test/java/com/digitalpebble/storm/crawler/indexer/IndexerTester.java (73)
> M external/elasticsearch/src/main/java/com/digitalpebble/storm/crawler/elasticsearch/bolt/IndexerBolt.java (2)
> M external/solr/src/main/java/com/digitalpebble/storm/crawler/solr/bolt/IndexerBolt.java (2)
> Patch Links:
> 
> https://github.com/DigitalPebble/storm-crawler/pull/161.patch
> https://github.com/DigitalPebble/storm-crawler/pull/161.diff
> —
> Reply to this email directly or view it on GitHub.
> 

--------------------------
Ken Krugler
+1 530-210-6378
http://www.scaleunlimited.com
custom big data solutions & training
Hadoop, Cascading, Cassandra & Solr

jnioche:
Hi @kkrugler. I think what @jorgelbg means is that whatever value found in the canonical tag could be a relative URL. The main URL will of course be an absolute one but the canonical tag does not go through filtering and normalisation : it's not a URL we fetch, just one we use for indexing.

jorgelbg:
@jnioche @kkrugler exactly in the canonical tag I've found about anything, valid an not valid. If a relative URL is found in the value of the canonical tag, then this URL is resolved into its absolute form.

jnioche:
@jorgelbg could you please squash your commits after making the changes above? Thanks!

Issue #156, opened 2015-07-03
SOLR : investigate use of field collapsing in Spout


jnioche:
See #152 

This would be a way of ensuring a good diversity of URLs in the results.

jnioche:
Implemented in #162. Thanks @jorgelbg

PR #162, opened 2015-07-20
Adding support for field collapsing in the Solr Spout

jorgelbg:
A couple of parameters were added to handle diversity Issue #156: 

* `solr.status.diversity.field.name`: Indicates the name of the field to be collapsed on.
* `solr.status.diversity.field.bucket`: Indicates the maximum bucket size of each expanded group. 

One good example to add to the wiki/documentation around this is to use the `status` field, which can have a small set of values, so it will be easy to explain and is a field that is populated by default, and provide a good balance between each type of URLs. 

I'm going to make other commit adding some explanation to the README file in the Solr module, if thats ok.

A better param name is welcome :smile: 

Also this PR fix a small typo in a parameter configuration of the `MetricsConsumer` class.

### DIFF
--- a/external/solr/src/main/java/com/digitalpebble/storm/crawler/solr/persistence/StatusUpdaterBolt.java
+++ b/external/solr/src/main/java/com/digitalpebble/storm/crawler/solr/persistence/StatusUpdaterBolt.java
@@ -24,15 +24,16 @@ import org.apache.solr.common.SolrInputDocument;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import backtype.storm.task.OutputCollector;
-import backtype.storm.task.TopologyContext;
-import backtype.storm.topology.OutputFieldsDeclarer;
-
 import com.digitalpebble.storm.crawler.Metadata;
 import com.digitalpebble.storm.crawler.persistence.AbstractStatusUpdaterBolt;
 import com.digitalpebble.storm.crawler.persistence.Status;
 import com.digitalpebble.storm.crawler.solr.SolrConnection;
 import com.digitalpebble.storm.crawler.util.ConfUtils;
+import com.digitalpebble.storm.crawler.util.URLUtil;
+
+import backtype.storm.task.OutputCollector;
+import backtype.storm.task.TopologyContext;
+import backtype.storm.topology.OutputFieldsDeclarer;
 
 public class StatusUpdaterBolt extends AbstractStatusUpdaterBolt {
 
@@ -79,6 +80,9 @@ public class StatusUpdaterBolt extends AbstractStatusUpdaterBolt {
         SolrInputDocument doc = new SolrInputDocument();
 
         doc.setField("url", url);
+
+        doc.setField("host", URLUtil.getHost(url));
+
         doc.setField("status", status);
 
         doc.setField("metadata", metadata.toString());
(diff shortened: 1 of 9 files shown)


jnioche:
I'd rename `solr.status.diversity.field.name` into `solr.status.bucket.field` and `solr.status.diversity.field.bucket` into `solr.status.bucket.maxsize`.

Note that `status` is not a great field to use. Instead it would be good to have the statusupdater create a custom field (configurable) based on the values provided by URLPartitioner. This could be used for sharding like in [https://github.com/DigitalPebble/storm-crawler/blob/master/external/elasticsearch/src/main/java/com/digitalpebble/storm/crawler/elasticsearch/persistence/StatusUpdaterBolt.java#L120] as well as the spouts for the field collapsing. Let's do that later.

Did you try this PR in a test crawl?

jorgelbg:
I've renamed the parameters, a lot better now :smile: 

@jnioche `status` is not a great field to use in a real case, but for testing purposes could be a nice example. I realize that for a real scenario you'll a use a better field, to balance the crawl, and as you say something that could be used along with the URLPartitioner, but I was thinking to use it as an example taken from the fields that are indexed by default in the status collection (url, status, metadata, nextFetchDate). Actually I was using it this way for testing purposes. But if you have a better example that I can include in the README, it would be great!

Yep, I've done a couple of test crawls, did you find any issues?

jnioche:
> But if you have a better example that I can include in the README, it would be great!

we can add the bucket field as part of the SOLR status updater bolt and refer to it in the README then

> Yep, I've done a couple of test crawls, did you find any issues?

not at all, haven't managed to use it yet which is why I wanted to ask before merging

jorgelbg:
> we can add the bucket field as part of the SOLR status updater bolt and refer to it in the README then

Since the default partition mode is "byHost" should we add a host field in the `StatusUpdaterBolt` then? 

Since I haven't added some tests to the solr module yet, running a couple of crawls is my main form of testing.

jorgelbg:
I've rebased the `solr-readme` branch into this one to avoid a merge that was causing conflicts and then squashed the commits related to the field-collapsing feature, this includes the additional section in the README and doesn't generate any conflict. I think is ok to merge?

external/solr/src/main/java/com/digitalpebble/storm/crawler/solr/persistence/StatusUpdaterBolt.java
```diff
@@ -79,6 +80,9 @@ public void store(String url, Status status, Metadata metadata,
         SolrInputDocument doc = new SolrInputDocument();
 
         doc.setField("url", url);
+
+        doc.setField("host", URLUtil.getHost(url));
```
jnioche: I'd rather we did that properly later + have you modified the schema files accordingly?

jnioche:
Could you add `solr.status.bucket.field` with `host` as a value to [https://github.com/DigitalPebble/storm-crawler/blob/master/external/solr/solr-conf.yaml]

jorgelbg:
Done

external/solr/src/main/java/com/digitalpebble/storm/crawler/solr/persistence/StatusUpdaterBolt.java
jorgelbg: Yep, the schema auto-updates except from the key field.

PR #169, opened 2015-08-14
Correction of a NPE in metadata transfert function

ludovic-boutros:
A correction of the issue 168.

Thx,

Ludovic.


jnioche:
Thanks. I've just merged this but on second thoughts I'm wondering whether it wouldn't be better to throw an exception instead - the input metadata should never be null and returning an empty md would hide the issue. Can you think of a reason why metadata should be null?

ludovic-boutros:
I can check.
This appends while using the Tika parser and the Solr external module.

jnioche:
@ludovic-boutros have reverted it. Would be fab if you could find where Metadata is null. Again this should never happen

PR #170, opened 2015-08-14
If each url is unique per host, the Solr spout is locked

ludovic-boutros:
The expand result of Solr is empty if each group contains only one document.
That causes a lock in the Solr spout.

### DIFF
--- a/external/solr/src/main/java/com/digitalpebble/storm/crawler/solr/persistence/SolrSpout.java
+++ b/external/solr/src/main/java/com/digitalpebble/storm/crawler/solr/persistence/SolrSpout.java
@@ -172,24 +172,27 @@ public class SolrSpout extends BaseRichSpout {
             QueryResponse response = connection.getClient().query(query);
             SolrDocumentList docs = new SolrDocumentList();
 
+            int numhits = response.getResults().size();
+      
             if (StringUtils.isNotBlank(diversityField)) {
-                Map<String, SolrDocumentList> expandedResults = response
-                        .getExpandedResults();
+              //get expand result
+              Map<String, SolrDocumentList> expandedResults = response
+                      .getExpandedResults();
 
-                for (String key : expandedResults.keySet()) {
-                    docs.addAll(expandedResults.get(key));
-                }
+              for (SolrDocument doc : response.getResults()) {
+                String df = (String) doc.get(diversityField);
 
-                // if all the urls are unique
-                // the expand result is empty
-                if (expandedResults.isEmpty()) {
-                  docs = response.getResults();
+                //is there multiple urls for this host
+                SolrDocumentList dfList = expandedResults.get(df);
+                if (dfList != null) {
+                  docs.addAll(dfList);
+                } else {
+                  docs.add(doc);
                 }
+              }
             } else {
-                docs = response.getResults();
+              docs = response.getResults();
             }
-
-            int numhits = docs.size();
 
             // no more results?
             if (numhits == 0)
--- a/external/solr/src/main/java/com/digitalpebble/storm/crawler/solr/persistence/SolrSpout.java
+++ b/external/solr/src/main/java/com/digitalpebble/storm/crawler/solr/persistence/SolrSpout.java
@@ -180,11 +180,6 @@ public class SolrSpout extends BaseRichSpout {
                     docs.addAll(expandedResults.get(key));
                 }
 
-                // if all the urls are unique
-                // the expand result is empty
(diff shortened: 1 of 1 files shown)


jnioche:
@jorgelbg could you please comment / review?

ludovic-boutros:
In fact this fix is not enough.
There is no lock anymore but in some configurations, some websites are never crawled.
 
The iteration on the Solr result is not correct.
I will post a new pull request soon.

jnioche:
What about the following strategy for handling the diversity of URLs in the SOLR spout \: we'd have a configurable number of buckets and map the host|domain|IP to a given bucket number (between 0 and max bucket). The spout instances would then determine which range of the buckets they are in charge of based on the instance number and total number of spout instances. Each spout instance would then run a query per bucket and populate an internal queue. This is not as accurate as doing an exact limit per host or domain but it means that having several instances of the spouts would share the loads quite easily + might be faster if the bucket field is used as a cache. The status updater bolt would compute the host to bucket mapping.
What do you think?

ludovic-boutros:
Julien,

this is what I have implemented currently ;)

I will update the pull request if I can.

ludovic-boutros:
Ok, I have added a commit which should implement your strategy.

100 buckets returned each time (should be configurable ?).
The max size of a bucket can be configured with the property "solr.status.bucket.maxsize"

external/solr/src/main/java/com/digitalpebble/storm/crawler/solr/persistence/SolrSpout.java
```diff
@@ -172,20 +172,28 @@ private void populateBuffer() {
...
-                        .getExpandedResults();
-
-                for (String key : expandedResults.keySet()) {
-                    docs.addAll(expandedResults.get(key));
+              //get expand result
+              Map<String, SolrDocumentList> expandedResults = response
+                      .getExpandedResults();
+
+              for (SolrDocument doc : response.getResults()) {
+                String df = (String) doc.get(diversityField);
```
jorgelbg: The main caveat I see here is that we're iterating twice on the main result set, since the CollapsingQParser already "collapsed" the main result set into one document per value in the bucket field (i.e. the host field) all we need to do is "merge" the results in the main result set with the ones already extracted from the expanded section. 

Basically something like:
```java
docs.addAll(response.getResults());
```
This guarantees that the `docs` var contains all the expanded results and also the "main" document in the results list, without the caveat of iterating over the main result set twice.

ludovic-boutros:
@jorgelbg The expanded result is a map, and should be used with this in mind.  
Here is the documentation:

_Inside the expanded section there is a map with each group head pointing to the expanded documents that are within the group. As applications iterate the main collapsed result set, they can access the expanded map to retrieve the expanded groups._

Now, I understand your point of view and I can propose this optimization.
If the sizes of the main result and of the expanded result are equal we can use the expanded result directly.

What do you think about this ?

jorgelbg:
The CollapsingQParser  collapses the result set to a single document per group, meaning that in the main result section we only get 1 document per bucket/group. In the other hand the expanded component gives us a map in the expanded section, containing the value of the bucket field mapped with the documents (URLs) with the same field bucket value, but since we're collapsing the fields using the CollapsingQParser, 1 document per each bucket/group is returned in the main Solr results section(`response.getResults()`), and the rest of the documents of each bucket/group (if any) gets returned in the expanded section (`response.getExpandedResults()`), this is why when an URL is unique in a bucket, no corresponding bucket/group is found in the expanded section. If all URLs are unique per bucket then no expanded sections gets returned at all. 

In the code as-is we iterate on the expanded results and add all the documents found in each group into a list of URLs for emitting https://github.com/DigitalPebble/storm-crawler/blob/master/external/solr/src/main/java/com/digitalpebble/storm/crawler/solr/persistence/SolrSpout.java#L176-L181.

Bottom line we need to process all the URLs found in the expanded section (if any) which what we do now, and also process the document-per-bucket that comes back *collapsed* in the main result section.

ludovic-boutros:
Hi @jorgelbg, 
ok, I understand.

The main collapsed result is not included in the collapsed result therefore we can add both results without using the map.
 
But, your change is not enough, the variable _numHits_ must be equal to the main result size in order to iterate correctly thru the URL buckets.

jnioche:
@ludovic-boutros can this one be closed now that I've merged #176?

ludovic-boutros:
@jnioche yes this one can be closed. Thx.

PR #171, opened 2015-08-14
Solr general improvements

ludovic-boutros:
Move on Solr 5.2.1
Add http-client dependency in order to be able to run solr test topologies
Sorry for the pom formatting, I'm using Netbeans and I've not found how to use different formatting for code and poms.
Would it be possible to use the same formatting (2 spaces for tabs ?)

Thx

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@ target
 # Ignore IntelliJ IDEA artifacts
 *.iml
 .idea
+
+# Ignore Netbeans config files
+nbactions.xml
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -23,7 +23,6 @@
 		<jsoup.version>1.8.1</jsoup.version>
 		<icu4j.version>54.1.1</icu4j.version>
 		<xerces.version>2.11.0</xerces.version>
-		<httpclient.version>4.4.1</httpclient.version>
 	</properties>
 
 	<build>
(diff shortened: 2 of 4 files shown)


jnioche:
Would be great if we could handle XML files with  `mvn java-formatter:format` - I'll have a look.

+1 to netbeans config in .gitignore
+1 to upgrade of SOLR version

re-httpclient \: don't we inherit it automatically from core?

ludovic-boutros:
Solr uses an older version of http-client which overrides the storm crawler one.

jnioche:
@ludovic-boutros What about using [http://plugins.netbeans.org/plugin/50877/eclipse-code-formatter-for-java] to apply our Eclipse formatting? Would that work?

ludovic-boutros:
Tested last week, it does not work, at least on my machine... :(

jnioche:
SOLR 5.2.1 uses the same version of httpclient as sc-core 4.4.1 so there should not be any conflict. Building with `mvn clean install` gives the correct dependency.
Is there a way to reproduce the issue?

jorgelbg:
I wasn't able to reproduce this, at least using my local university maven repo.

ludovic-boutros:
@jnioche SOLR 5.1.0 uses httpclient 4.3.1. That's why I had to force the version of this module.
Now, you're right SOLR 5.2.1 uses the same version (SOLR 4.4.1).
I can revert this change if you prefer or exclude the dependency (this is another option).

jnioche:
salut @ludovic-boutros. We could leave it as-is and just upgrade the version of SOLR for now.

ludovic-boutros:
okay ! :)

jnioche:
solr 5.3.0 has been released in the meantime. I'll upgrade the version and also add the .gitignore change in a separate commit

jnioche:
committed in [35900d5] and [9767611611650d1fb39533]
thanks @ludovic-boutros

PR #173, opened 2015-08-20
loadConf() already calls putAll()

SyBen:
At the end of ConfUtils.loadConf(), you either have a (Config + yaml options) Map or a Config Map. Either way, there is no need for a second putAll().

### DIFF
--- a/core/src/main/java/com/digitalpebble/storm/crawler/protocol/httpclient/HttpProtocol.java
+++ b/core/src/main/java/com/digitalpebble/storm/crawler/protocol/httpclient/HttpProtocol.java
@@ -177,10 +177,10 @@ public class HttpProtocol extends AbstractHttpProtocol implements
 
     public static void main(String args[]) throws Exception {
         HttpProtocol protocol = new HttpProtocol();
+        Config conf = new Config();
 
         String url = args[0];
-        Config conf = ConfUtils.loadConf(args[1]);
-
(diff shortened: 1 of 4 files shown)


jnioche:
Not quite. It is possible to have multiple -conf XXX on the command line, for instance to separate a generic configuration file from a more specific one. The change you suggested would only load the last file specified.

What we could do instead would be to change `ConfUtils.loadConf(resource)` into `ConfUtils.loadConf(Config conf, String resource)` so that it adds to an existing Config object. What do you think?

SyBen:
My bad, I missed that use case. Sounds good to me, As it would match the other methods behavior of the same class.

core/src/main/java/com/digitalpebble/storm/crawler/protocol/httpclient/HttpProtocol.java
```diff
@@ -177,10 +177,10 @@ public BaseRobotRules getRobotRules(String url) {
 
     public static void main(String args[]) throws Exception {
         HttpProtocol protocol = new HttpProtocol();
+	Config conf = new Config();
```
jnioche: formatting? you can apply the formatting rules with `mvn java-formatter:format'

jnioche:
Merci Sylvain. Could you please squash your commits after applying the formatting rules? Thanks!

jnioche:
Merged, thanks!

PR #176, opened 2015-08-26
Fixing issue with unique URL per host causing the Solr spout to lockup 

jorgelbg:
Merge the main result set with the expanded result set when a diversity field is used.
Use the size of the main results section for pagination


jorgelbg:
If its ok, I can merge this tomorrow!

ludovic-boutros:
+1 for me.

jnioche:
thanks guys

Issue #168, opened 2015-08-13
NPE in function MetadataTransfert.filter()


ludovic-boutros:
Hi,

the parameter 'metadata' can be null in the function MetadataTransfert.filter().
It throws a NPE later.

Just adding the test and returning 'md' fix this issue.

jnioche:
@ludovic-boutros did you get the NPE after I reverted your PR? Would be good to find where we get a null metadata.

ludovic-boutros:
Not yet. I will try soon.

ludovic-boutros:
Ok I reproduce the issue while running the Solr seed injector example. 
Will try to figure out what's wrong here.

ludovic-boutros:
It seems that the FileSpout emits null metadata...

jnioche:
So this would mean that [https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/java/com/digitalpebble/storm/crawler/bolt/URLPartitionerBolt.java#L69] should also check for the value of metadata. In the case you found we were emitting a field metadata but with a null value. I'll fix this.

jnioche:
@ludovic-boutros please close this issue if you are confident that the problem is now solved. Thanks

PR #179, opened 2015-08-27
Initialize the metadata in the StringTabScheme, should never be null

ludovic-boutros:
This is a fix for the issue #168


jnioche:
good catch, thanks!

ludovic-boutros:
de rien ;)

Issue #177, opened 2015-08-26
[External][Solr] The Solr storage should use 'autoSoftCommit' and 'autoCommit'


ludovic-boutros:
Currently the Solr storage modules send hard commits to Solr. 

I think this is not the good way to proceed.
These commits cause too much pressure on Solr. See the following screenshot.

![capture d ecran 2015-08-12 a 10 10 18](https://cloud.githubusercontent.com/assets/13764888/9488752/5a1f5b62-4bdc-11e5-82c4-ca2e59235618.png)

Instead, these modules should let Solr commit the documents with 'autoCommit' and 'autoSoftCommit' sections in the 'solrconfig' file.

jorgelbg:
I think the more generic approach could be make this an option in the configuration file? In some cases I've seen than relying only on autoCommit or autoSoftCommit can be "problematic" that's why the module issue a hard commit every now and then. What do you think?

ludovic-boutros:
Hi @jorgelbg,

I think that the usage of Solr in SC can be considered as a NRT use case.
In this type of use cases, I think explicit hard commits are not a good solution.

I really don't like having explicit commits in my code ;)

I've never had any bad experiences with autocommits, could you elaborate a bit please ? 

Another solution could be to remove explicit hard commits and add a _commitWithin_ parameter to each _add_ command ?

This would be an additional parameter. That would work too I think.

jnioche:
@ludovic-boutros did you get an improvement to the perfs of the status updater by using soft commits? 

I expected the commit strategy (auto|soft|hard) to be done entirely on the SOLR config side and not at all in the code. Is that required because the client buffers the docs to index? Isn't there another SOLR client implementation which has a max doc / max time parameter? 

The current implementation would be problematic when a few docs are processed e.g. end of a crawl as we would not get the minimal size before we commit. See [https://github.com/DigitalPebble/storm-crawler/blob/master/external/aws/src/main/java/com/digitalpebble/stormcrawler/aws/bolt/CloudSearchConstants.java#L27] for how I handled that for CloudSearch.

@jorgelbg the handling of errors is incorrect. At the moment you just log the error which means that the input tuple gets acked. It would need to be failed either explicitely or by rethrowing the exception so that AbstractStatusUpdaterBolt can deal with it.

ludovic-boutros:
@jnioche I've tested with _commitWithin_ and it works quite well. 
But I would prefer a strategy in the solrconfig file too.

[...]

PR #180, opened 2015-08-28
[External][Solr] The Solr storage should use 'autoSoftCommit' and 'autoCommit'

ludovic-boutros:
This is a fix proposal for the issue #177.

- No explicits commit in the code
- Add autoCommit and autoSoftCommit in solrconfig files
- Update Lucene version in solrconfig files

### DIFF
--- a/external/solr/solr-conf.yaml
+++ b/external/solr/solr-conf.yaml
@@ -1,21 +1,12 @@
 # Solr indexer bolt
 solr.indexer.url: "http://localhost:8983/solr/collection1"
-solr.indexer.threads: 1
-solr.indexer.queue.size: 100
-solr.indexer.commit.size: 1
 
 # Solr spout and persistence bolt
 solr.status.url: "http://localhost:8983/solr/status"
-solr.status.threads: 1
-solr.status.queue.size: 100
-solr.status.commit.size: 1
 solr.status.bucket.field: host
 
 # Solr MetricsConsumer
 solr.metrics.url: "http://localhost:8983/solr/metrics"
-solr.metrics.threads: 1
-solr.metrics.queue.size: 100
-solr.metrics.commit.size: 1
 
 # For SolrCloud, this settings instead of solr.indexer.url
 #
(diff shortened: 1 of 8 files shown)


jnioche:
looks good. @jorgelbg any comments?

jorgelbg:
Looks good to me, I'm going to take it for a spin later today.

jorgelbg:
@ludovic-boutros @jnioche :+1: looks good tom me I've done a couple of basic crawl tests and haven't found any problems.

jnioche:
Thanks. Just to be sure I understand correctly @ludovic-boutros, are the documents buffered somehow or do we do 1 request to SOLR per document to index? The latter would seem rather costly

ludovic-boutros:
@jnioche HttpSolrClient and CloudSolrClient do not buffer any updates. 
We could implement a buffering/batch system with callbacks in order to acknowledge updates correctly.
It could be done in a second phase if needed but:
- I don't think it is needed currently
- It could be a bit tricky with SolrCloud if we want to keep routing awareness on the client side

This implementation uses Solr like a NoSql DB and lets the update/commit management be done on the server side.

jnioche:
> HttpSolrClient and CloudSolrClient do not buffer any updates. 
thanks for the clarification @jorgelbg

jorgelbg:
@ludovic-boutros why buffering/queueing of documents with SolrCloud could be tricky? At this point we are not supporting any kind of routing in the client side, except that the CloudSolrClient is topology-aware thanks to Zookeeper and as such it is able of routing the document to the right shard/collection leader. 

For routing purposes we can use the `KEY!ID` syntax to influence routing of documents into specific shards, but still the routing of the document is done on Solr itself. Also an specific field could be used (depending on how the collection is created) to route the documents.

ludovic-boutros:
@jnioche @jorgelbg We forgot to remove the batch size parameters in the wiki.
@jorgelbg I will try to give you an answer soon, quite busy this week ;)

ludovic-boutros:
@jorgelbg ok you are right, I just checked the distribution of a collection of documents in the code. It seems to split correctly the documents across the correct shards.

Which solution would you implement ? 
- create a SolrClient which acknowledge tuples
- other

What should/could be done if only one doc throws an error ?
I remember that Solr is not really wordy on this in its messages.
Perhaps, this should be discussed in another issue.

jnioche:
Hi. Could you please open a new issue for it + create one for fixing the README file 
Thanks!

PR #184, opened 2015-09-11
Include missing JAR dependencies in pom.xml

allfro:
Updated the dependencies section to include snakeyaml and commons-lang as dependencies. Otherwise, the storm program and topology throw CNFEs.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -24,6 +24,8 @@
 		<icu4j.version>54.1.1</icu4j.version>
 		<xerces.version>2.11.0</xerces.version>
 		<httpclient.version>4.4.1</httpclient.version>
+		<snakeyaml.version>1.1.6</snakeyaml.version>
+		<commons.version>2.6</commons.version>
 	</properties>
 
 	<build>
@@ -188,5 +190,16 @@
 			<version>${httpclient.version}</version>
 		</dependency>
 
+		<dependency>
+			<groupId>org.yaml</groupId>
+			<artifactId>snakeyaml</artifactId>
+			<version>${snakeyaml.version}</version>
+		</dependency>
+		
+		<dependency>
+			<groupId>commons-lang</groupId>
+			<artifactId>commons-lang</artifactId>
+			<version>${commons.version}</version>
+		</dependency>
 	</dependencies>
 </project>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -25,7 +25,7 @@
 		<xerces.version>2.11.0</xerces.version>
 		<httpclient.version>4.4.1</httpclient.version>
 		<snakeyaml.version>1.1.6</snakeyaml.version>
-		<commons.version>2.6</commons.version>
+		<commons.lang.version>2.6</commons.lang.version>
 	</properties>
 
 	<build>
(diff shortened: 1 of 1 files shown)


jnioche:
Not clear why you are getting these exceptions. Both dependencies are inherited from `org.apache.storm:storm-core:jar:0.9.5:provided`, so if you are running the topology with the 'storm' command they should be automatically on the classpath. 

Can you give some steps to reproduce the issue? Thanks

allfro:
I am using storm 0.10 which apparently does not provide these jars as I am
getting ClassNotFoundException's when I view the error log from storm-ui. I
built the jar file using the exact instructions you have in the README.md
and submitted the job to my pseudo distributed cluster the way you
suggested as well. Everything was a copy and paste build and deploy which
resulted in the errors I encountered.

jnioche:
> I am getting ClassNotFoundException's when I view the error log from storm-ui

Sounds like it is more of a Storm issue than a SC one per-se, as there is nothing specific to SC in the log viewer.  Maybe open an issue in Storm instead?

allfro:
I should be more specific. The CNFEs show up under the error columns of
certain bolts. I can send you the error log but this patch appeared to fix
the issue.

allfro:
I just double checked to make sure and I can confirm that the snakeyaml and commons jars are not present in version 0.10 of storm. I don't think this is a storm issue since they probably don't rely on those jars anymore. According to their changelog they replaced snakeyaml with jvyaml.

jnioche:
>  The CNFEs show up under the error columns of certain bolts
ok different story indeed

we still rely on storm 0.5 so not yet an issue but wouldn't do any harm to add them explicitly to the pom to futureproof it.

I'd change `commons.version` to `commons.lang.version`, WDYT?

allfro:
Done

jnioche:
thanks @allfro 
Would be great to hear how you use SC? Feel free to drop a message on the mailing list if you have the time and inclination ;-)

allfro:
Sure,

We're currently investigating replacing nutch with it or rewriting nutch on
top of spark. Will let you know if this is a viable alternative as we are
looking to use storm in our infrastructure and the programming model is
much simpler than nutch's. Also, I'm currently investigating the use of
Redis as a spout where the spout would listen for key expirations and use
the key as the URL to crawl next; might simplify the architecture of the
fetchers but increase complexity of the spout. Have you investigated
whether this is something people have tried to do? Will let you know if we
have any success.

Cheers,

Nadeem

jnioche:
>  Also, I'm currently investigating the use of
Redis as a spout where the spout would listen for key expirations and use
the key as the URL to crawl next; might simplify the architecture of the
fetchers but increase complexity of the spout. Have you investigated
whether this is something people have tried to do? 

Have you seen the [SimpleFetcherBolt](https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/java/com/digitalpebble/storm/crawler/bolt/SimpleFetcherBolt.java). It's a trimmed down version if the multithreaded one, but it means that the queueing per host/domain has to be done upstream e.g. at the spout level.

re-Redis @jakekdodd used it - see [http://www.slideshare.net/ontopic/storm-crawler-apacheconna2015] slide 23, not sure he used key expirations though.

BTW great work investigating NUTCH-1084

allfro:
Thanks. Would you be up to up voting the issue on JIRA? Really hope this
can get committed.

PR #186, opened 2015-09-16
Fixed snakeyaml version

mattburns:
Was 1.1.6 which looks like a typo. This is a build failure since it can't be found in maven repo.


jnioche:
Thanks @mattburns

PR #188, opened 2015-09-19
Added missing storm.topology property to pom.xml

allfro:


jorgelbg:
Considering that `com.digitalpebble.storm.crawler.CrawlTopology` is provided as an example topology of what can be done, the general idea is that you build your own topology and provide the right class as a commando line, as stated in https://github.com/DigitalPebble/storm-crawler/blob/master/README.md.

allfro:
I understand,

I figure a default would be nice for noobs like myself.

jnioche:
I'd be tempted to get rid of the variable 'storm.topology' altogether and rely on the standard '-Dexec.mainClass' instead. Having it adds nothing IMHO.

I'd rather avoid having user-dependent variables in the pom. In practice most people would run the topologies either from Eclipse or by calling the Storm command.

PR #209, opened 2015-10-30
PerSecondReducer for Storm metrics + 

jnioche:
Used in Fetcher bolts for byte and docs per sec.


core/src/main/java/com/digitalpebble/storm/crawler/bolt/SimpleFetcherBolt.java
```diff
@@ -150,6 +152,9 @@ public void prepare(Map stormConf, TopologyContext context,
         this.averagedMetrics = context.registerMetric("fetcher_average",
```
jnioche: needs renaming to fetcher_average_perdoc for consistency across Fetcher bolts

core/src/main/java/com/digitalpebble/storm/crawler/util/PerSecondReducer.java
```diff
@@ -0,0 +1,42 @@
...
+                            + "`. Supported types are Double, Long, Integer.");
+        }
+        return accumulator;
+    }
+
+    @Override
+    public Object extractResult(TimeReducerState accumulator) {
+        // time spent
+        long sec = (System.currentTimeMillis() - accumulator.started) / 1000;
+        return new Double(accumulator.sum / (double) sec);
```
jnioche: TODO add check that don't have division by 0 - even if very unlikely

Issue #207, opened 2015-10-29
Store the metadata field in the status collection to be searchable


jorgelbg:
Similar to #197 #201 but for Solr, in this case since Solr have a different approach to nested objects/documents, perhaps a better approach could be to flatten the metadata field using a prefix, combined with a dynamic setting? 

Thoughts on this are welcome, I think that using nested documents could be an overkill, plus since Solr doesn't "hide" the child documents the status collection will be packed with additional metadata/child documents.

ludovic-boutros:
@jorgelbg That would be great. I also agree with the flattening of the metadata.

jnioche:
+1 thanks!

jnioche:
@jorgelbg shall we mark it for release 0.7?

jorgelbg:
@jnioche yes sure! I'll submit a PR later today

PR #210, opened 2015-10-30
Updating solr module to store the metadata field as separated fields

jorgelbg:
This address issue #207. Also contains an update for the documentation of the Solr module

### DIFF
--- a/external/solr/src/main/java/com/digitalpebble/storm/crawler/solr/persistence/SolrSpout.java
+++ b/external/solr/src/main/java/com/digitalpebble/storm/crawler/solr/persistence/SolrSpout.java
@@ -225,11 +225,12 @@ public class SolrSpout extends BaseRichSpout {
                     if (key.startsWith(prefix)) {
                         Collection<Object> values = doc.getFieldValues(key);
 
+                        key = StringUtils.replace(key, prefix, "", 1);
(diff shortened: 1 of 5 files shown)


external/solr/src/main/java/com/digitalpebble/storm/crawler/solr/persistence/SolrSpout.java
```diff
@@ -208,20 +218,19 @@ private void populateBuffer() {
...
+                    String key = keyIterators.next();
+
+                    if (key.startsWith(prefix)) {
+                        Collection<Object> values = doc.getFieldValues(key);
+
+                        Iterator<Object> valueIterator = values.iterator();
+                        while (valueIterator.hasNext()) {
+                            String value = (String) valueIterator.next();
+
+                            metadata.addValue(StringUtils.replace(key, prefix, "", 1), value);
```
jnioche: the replacement could be done just once outside the while loop as it will be the same for all the values

jorgelbg:
Done, I've also updated the `schema.xml` file for the status collection, since some metadata values are multivalued (path) I've defined a dynamic field to catch all fields starting with the `metadata` prefix as multivalued.

jnioche:
Thanks @jorgelbg

PR #223, opened 2015-11-27
Modified example to do recursive crawling in local mode

w0mbat:


jnioche:
I'd create a separate StatusUpdater for this instead of modifying StdOutStatusUpdater

core/src/main/java/com/digitalpebble/storm/crawler/spout/RandomURLSpout.java
```diff
@@ -42,15 +43,14 @@
     private boolean active = true;
     private boolean removeAfterSending = true;
 
-    private String[] urls = new String[] { "http://www.lequipe.fr/",
-            "http://www.lemonde.fr/", "http://www.bbc.co.uk/",
-            "http://www.facebook.com/", "http://www.rmc.fr" };
-
-    public RandomURLSpout(String... urls) {
```
jnioche: This constructor is very useful. It should be kept and populate the HashMap just like the default constructor does.

core/src/main/resources/urlfilters.json
```diff
@@ -4,7 +4,7 @@
       "class": "com.digitalpebble.storm.crawler.filtering.depth.MaxDepthFilter",
       "name": "MaxDepthFilter",
       "params": {
-        "maxDepth": 0
+        "maxDepth": 10
```
jnioche: That's unrelated to the functionality of this PR and can/should be overridden in the user config. It should not be changed here.

w0mbat:
@jnioche I updated my pull request. I added a separate topology, seperate configs and also documented the commands. This way it can run out-of-the-box.

Hope its better now :)

jnioche:
Duplicating the conf + topology + filters is unnecessary. See #224

Issue #233, opened 2016-01-20
Field collapsing in ES Spout


jnioche:
The ES Spout currently sorts per _nextFetchDate_ which offers very few guarantees regarding the diversity of hosts retrieved by the queries. This is improved by randomizing the results, however the latter has an impact on the memory used by ES for the field caching. Limiting the field caching leads to poor performance  and the eviction mechanism slows the queries quite a lot.

A more natural solution to the problem would be to use [Field Collapsing](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-top-hits-aggregation.html) e.g. on the _metadata.hostname_ field. This would guarantee a good diversity of hosts and optimal performance for the crawler.

Below is an example of what the ElasticSearch query would look like

```json
{
  "query": {
    "range": {
      "nextFetchDate": {
        "lt": "now"
      }
    }
  },
  "size": 0,
  "aggs": {
    "hostname": {
      "terms": {
        "field": "metadata.hostname",
        "size": 50
      },
      "aggs": {
        "top_tags_hits": {
          "top_hits": {
            "size": 5
          }
        }
      }
    }
  }
}
```

This is of course compatible with the sharding mechanism in #232

PR #234, opened 2016-01-21
Elasticsearch Spout using FieldCollapsing

jnioche:
See #233 

Currently implemented as a separate implementation from the existing one.

@DigitalPebble/committers-crawler comments welcome as usual


jnioche:
Replaced by #237 which has a better name for the spout + based on fresh version of branch.

PR #238, opened 2016-01-27
Use default of "nextFetchDate" for bucket sorting

mattburns:
The `bucketSortField` variable was initialised with the default of `nextFetchDate` but then later overwritten.
Unlikely to be a problem since the default is also in the conf file.


jnioche:
This is on purpose. Otherwise if the users do not want to use the filter and do not set it in their config file at all the default value (nextFetchDate) would still be used. If they set an empty value then that's fine. 

We could instantiate the field with null or an empty string instead.

jnioche:
cea63a2 instanciates with an empty string, which we can also use as default as per your patch. Thanks

Issue #216, opened 2015-11-17
StatusUpdaterBolts must be able to ack/fail explicitly


jnioche:
The class [AbstractStatusUpdaterBolt](https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/java/com/digitalpebble/storm/crawler/persistence/AbstractStatusUpdaterBolt.java) currently acks each tuple after calling the store method. Some implementations e.g. ElasticSearch one have a buffering client which means that we don't know instantly whether an update has succeeded or failed. We could change the abstract class so that extensions can fail or ack themselves.

w0mbat:
I also experienced the problem in the last couple of days where we had some connection issues with ES and the tuples were gone

jnioche:
you mean they'd been acked regardless of whether the ES indexing has succeeded? Maybe try the branch in #241

w0mbat:
@jnioche yes, they were acked although indexing failed. Thanks - Will try later

PR #241, opened 2016-01-28
AbstractStatusUpdaterBolt allow proper acking of tuples

jnioche:
Decoupled acking of tuples in AbstractStatusUpdaterBolt + ES StatusUpdaterBolt now properly acks tuples **after** they've been sent to ES.

Implements #216 

This prevents URLs from being excessively replayed by the ES spout. They were acked before being actually sent to ES, removed from the cache in the spout then refetched immediately as their status had not been changed.

~~@mattburns do you want to give this a try?~~ Still WIP and not quite ready to be used.


jnioche:
Replaced by #263

PR #244, opened 2016-01-29
Ignore OSX system files

mattburns:
.DS_Store is just a system file for storing custom folder attributes


jnioche:
thanks @mattburns

Issue #246, opened 2016-02-03
Reduce metrics chatter


mattburns:
When writing metrics to Elasticsearch, the index can get big enough to affect performance of Elasticsearch. Here are my thoughts on things that may help reduce the burden on Elasticsearch:

1. Log fewer stats
  1. Configurable (crawler-conf) whitelist/blacklist of metric names?
  1. Maybe just skip writing of internal ones (starting with "__") used by storm-ui.
1. Log less frequently
  1. Configurable granularity of update frequency (time bucket size currently hardcoded to 10 seconds)
1. Regularly prune the DB
  1. Have two indexes (metrics-fine and metrics-coarse?) one with high resolution (update frequency) but short TTL*, and the other the opposite.
  1. Have a silly utility class that can prune the metrics index using a . eg, "delete metrics older than X" or "delete metrics with name Y"
  1. Use the new [delete by query plugin](https://www.elastic.co/guide/en/elasticsearch/plugins/2.2/plugins-delete-by-query.html)?

*Note, [elastic have deprecated _ttl](https://www.elastic.co/guide/en/elasticsearch/reference/2.2/mapping-ttl-field.html), we should probably move to index-per-timeframe?

What are your thoughts to those suggestions? I'd prefer logging less over pruning because it also reduces the network chatter etc. I'm tempted by the whitelist/blacklist idea because it seems pretty straight forward. I was thinking you could specify something like this in the crawler-conf:

```
metrics.whitelist:
- fetcher_counter.fetched
- fetcher_counter.exception
```

Thoughts?

jnioche:
Black/whitelist would be a good initial approach, followed by configurable per component update frequency.
I am less keen on the regular pruning of the DB too. Having 2 indices fine vs coarse would be tricky as for pruning this is pretty much what _ttl does. Since we are on ES 1.x for now we can just use it. Later on we could indeed look at index-per-timeframe.

I expect that the more bolt instances (e.g. SimpleFetcher) in a topology, the more acute the problem becomes.

The metrics index could be handled by a separate ES machine to avoid degrading the perf of the one used for the status and/or index.

[...]

PR #250, opened 2016-02-08
Allow the ES metrics writer to be configurable about which metrics it…

mattburns:
… stores in ES. Refs #246


jnioche:
thanks @mattburns

PR #251, opened 2016-02-09
Fixed whitelist blacklist to allow filtering by parent. Fixes #246

### LINKED ISSUE #246: Reduce metrics chatter

mattburns:


mattburns:
This change adds the handy feature of being able to use partial keys for the white/black lists. Eg, you can just specify "fetcher_counter" to get all the metrics of that type.
It also fixes a bug (sorry!). I wasn't correctly handling the case for metrics were parents of a map (eg fetcher_average holds a map with keys : bytes_fetched, fetch_time, wait_time)

PR #258, opened 2016-02-19
ES-based Spouts use a static Client instance

jnioche:
The ES clients are thread-safe, there is no reason why they should not be shared among the Spout instances within the same worker. 

@mattburns what do you think?


mattburns:
Sounds like a good idea to me.

For future reference, here's a good pattern when you need to synchronize access to the construction of a a shared object: https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom

Totally irrelevant in this case because it happens within the `open` method and so doesn't need to be optimised for millions of repeated invocations. I just thought it was interesting ;)

PR #261, opened 2016-02-26
Automatic metadata to ES mapping

w0mbat:
This PR implements the automatic mapping of fields specified in indexer.md.mapping to either ES fields or objects, depending on if the metadata has subfields or not.
This makes it easy to specify e.g. `- parse=parse` and ES will have an object called parse containing all the fields and objects which were inside the parse metadata object.

### DIFF
--- a/core/src/main/java/com/digitalpebble/storm/crawler/bolt/FetcherBolt.java
+++ b/core/src/main/java/com/digitalpebble/storm/crawler/bolt/FetcherBolt.java
@@ -524,9 +524,6 @@ public class FetcherBolt extends BaseRichBolt {
                     response.getMetadata().setValue("fetch.statusCode",
                             Integer.toString(response.getStatusCode()));
 
-                    response.getMetadata().setValue("fetch.loadingTime",
-                            String.valueOf(timeFetching));
-
                     // passes the input metadata if any to the response one
                     response.getMetadata().putAll(metadata);
 
--- a/core/src/main/java/com/digitalpebble/storm/crawler/bolt/FetcherBolt.java
+++ b/core/src/main/java/com/digitalpebble/storm/crawler/bolt/FetcherBolt.java
@@ -524,6 +524,9 @@ public class FetcherBolt extends BaseRichBolt {
                     response.getMetadata().setValue("fetch.statusCode",
                             Integer.toString(response.getStatusCode()));
 
+                    response.getMetadata().setValue("fetch.loadingTime",
+                            String.valueOf(timeFetching));
+
                     // passes the input metadata if any to the response one
                     response.getMetadata().putAll(metadata);
 
--- a/core/src/main/java/com/digitalpebble/storm/crawler/bolt/SimpleFetcherBolt.java
+++ b/core/src/main/java/com/digitalpebble/storm/crawler/bolt/SimpleFetcherBolt.java
@@ -344,6 +344,9 @@ public class SimpleFetcherBolt extends BaseRichBolt {
             response.getMetadata().setValue("fetch.statusCode",
                     Integer.toString(response.getStatusCode()));
 
+            response.getMetadata().setValue("fetch.loadingTime",
+                    Long.toString(timeFetching));
+
             response.getMetadata().putAll(metadata);
 
             // determine the status based on the status code
--- a/core/src/main/java/com/digitalpebble/storm/crawler/bolt/SimpleFetcherBolt.java
+++ b/core/src/main/java/com/digitalpebble/storm/crawler/bolt/SimpleFetcherBolt.java
@@ -344,9 +344,6 @@ public class SimpleFetcherBolt extends BaseRichBolt {
             response.getMetadata().setValue("fetch.statusCode",
                     Integer.toString(response.getStatusCode()));
 
-            response.getMetadata().setValue("fetch.loadingTime",
-                    Long.toString(timeFetching));
-
             response.getMetadata().putAll(metadata);
 
             // determine the status based on the status code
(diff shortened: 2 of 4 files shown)


w0mbat:
@jnioche I corrected the commits and removed the change to the metadata class

core/src/main/java/com/digitalpebble/storm/crawler/bolt/FetcherBolt.java
```diff
@@ -524,6 +524,9 @@ public void run() {
                     response.getMetadata().setValue("fetch.statusCode",
                             Integer.toString(response.getStatusCode()));
 
+                    response.getMetadata().setValue("fetch.loadingTime",
```
jnioche: why not but should be in a separate PR

core/src/main/java/com/digitalpebble/storm/crawler/bolt/SimpleFetcherBolt.java
```diff
@@ -344,6 +344,9 @@ public void execute(Tuple input) {
             response.getMetadata().setValue("fetch.statusCode",
                     Integer.toString(response.getStatusCode()));
 
+            response.getMetadata().setValue("fetch.loadingTime",
+                    Long.toString(timeFetching));
```
jnioche: Same as above

w0mbat:
@jnioche I removed the two parts and will create a separate PR

jnioche:
Am probably missing something but I can't see a way of switching the functionality on/off based on configuration. We'd want the existing behavior to remain the same but have a way of structuring the objects on a per key basis. What about having a new config entry specifying for which key prefixes structured objects would be generated?

w0mbat:
@jnioche My goal was not to make my changes switchable on/off, because I didn´t change the behaviour, if the mapping definitions in `indexer.md.mapping` stay the same:
If you map a field, it will create a field in ES - if you map an object it will create an object in ES and this even completely recursively.

We use it for example to recursively map the `microdata.product=product` to an object in ES and this one has subobject and fields.

jnioche:
>  I didn´t change the behaviour, if the mapping definitions in indexer.md.mapping stay the same
> If you map a field, it will create a field in ES - if you map an object it will create an object in ES and this even completely recursively.

Not clear what you mean by 'field' and 'object' - all we deal with is a Map<String, String[]>.

> We use it for example to recursively map the microdata.product=product to an object in ES and this one has subobject and fields.

Could you clarify what would the value associated with 'microdata.product' look like? 

If the value is in Json, what we could have would be a config to specify key names for which the String values would be parsed from Json and indexed as a structured object in ES.

What I understood was that you wanted to have e.g.

microdata.product.name = "A"
microdata.product.serial= "xxx"
microdata.product.availability= "0"

and convert that into a structured object.

w0mbat:
ok I try to clear it up:
The pain I tried to solve was that we have a lot of fields that are extracted by the microdata parser and we wanted everything under `microdata.product` to be a json object inside ES - with the existing way we were not able to do it. In contrast we had to do something like this:

```
- microdata.product.name=product-name
- microdata.product.image=product-image
- microdata.product.offers.availability=product-offers-availability
- microdata.webpage.breadcrumb=webpage-breadcrumb
```
To us, this was very inconvenient, as we had to a) parse the data, b) put it into the metadata map and c) add a mapping in the crawler-conf.yml to store the new field in ES.
The way we wanted it to look like was to be able to tell the ES mapper to take everything under `microdata.product` and map it to an ES object like this one:

```json
product: {
image: "https://images.otto.de/asset/mmo/formatz/aeg-waschmaschine-lavamat-l14as7-aplusplusplus-7-kg-1400-u-min-weiss-15403117.jpg",
additionalProperty: {
name: "variationId",
value: "504747671-M48"
},
description: "AEG Waschmaschine LAVAMAT L14AS7, A+++, 7 kg, 1400 U/Min",
title: "Frontlader",
url: "https://www.otto.de/p/aeg-waschmaschine-lavamat-l14as7-aplusplusplus-7-kg-1400-u-min-508571361/",
contentURL: "https://edge.mycliplister.com/cls/content/clip/86605/940305/aeg-waschmaschine-lavamat-l14as7-aplusplusplus-7-kg-1400-u-min.flv",
reviewCount: "(819)",
ratingValue: "* * * * *",
name: "AEG Waschmaschine LAVAMAT L14AS7, A+++, 7 kg, 1400 U/Min",
sku: "28763089",
aggregateRating: "#customer-reviews",
brand: "AEG",
thumbnailUrl: "https://edge.mycliplister.com/cls/content/thumb/86605/940305/aeg-waschmaschine-lavamat-l14as7-aplusplusplus-7-kg-1400-u-min.jpg"
}
```

As I was building it, I wanted to still support the existing syntax of mapping. For example it would still be possible to use `microdata.product.name=product-name` as a mapping directive and the changes I did would result in the previous behaviour of storing a plain string field in ES.

jnioche:
Thanks for the explanation. The functionality makes sense but I find the code in [AbstractIndexer](https://github.com/w0mbat/storm-crawler/blob/a99730fe488158922860321e67aff8ebbe146607/core/src/main/java/com/digitalpebble/storm/crawler/indexing/AbstractIndexerBolt.java#L173) difficult to understand and possibly error prone.

If I understand correctly for each mapping defined in the conf, you iterate on all the metadata keys beginning with the key specified in the mapping then add them to the Map and replace the prefix. 

Later on in the ES indexing bolt you iterate on the filtered Map, split it based on '.'  (which is not a constraint you had in AbstractIndexer) then build the structured object. 

Now if you have the following mappings \:

```
test=fname
test1=field1
test12=field12
test13=field13
```

the output would depend on the order in which the elements from metadata2field.entrySet()  are returned - unless I am mistaken.

I can foresee situations where users will not get the results they expect even if all they want is a straightforward mapping. I'd be more comfortable with a more consistent mechanism whereby we'd explicitly have a config for metadata paths for which we'd want to build a structured object. This could be done within the ES IndexingBolt only and made more generic if necessary later on. 

For instance something like 'indexer.md.mapping.object', which in your case would have for value 
'microdata.product'. You could then take the input metadata and build a Map accordingly.

I think this would make it easier to understand and debug + we'd be sure not to break the existing behaviour.

What do you think?

PR #257, opened 2016-02-17
Updated elasticsearch to 2.2.0

w0mbat:

PR #275, opened 2016-04-20
Update to Elasticsearch 2.x

jnioche:
Based on #257, uses ES 2.3.1

The only limitation is that the node client does not work for now, but the transport one does.

@w0mbat thanks for your work on #257 - I've built a new branch but yours was a great starting point. Any chance you could give this one a try?

### DIFF
--- a/external/elasticsearch/ES_IndexInit.sh
+++ b/external/elasticsearch/ES_IndexInit.sh
@@ -29,9 +29,6 @@ curl -XPOST localhost:9200/status -d '
       "_all": {
         "enabled": false
       },
-      "_id": {
-        "path": "url"
-      },
       "properties": {
         "nextFetchDate": {
           "type": "date",
--- a/external/elasticsearch/es-conf.yaml
+++ b/external/elasticsearch/es-conf.yaml
@@ -3,7 +3,7 @@
 # ES indexer bolt
 # Transport client will be used if at least one address is specified
 # Node client will be used otherwise
-# es.indexer.addresses: "localhost:9300"
+es.indexer.addresses: "localhost:9300"
 es.indexer.index.name: "index"
 es.indexer.doc.type: "doc"
 es.indexer.cluster.name: "elasticsearch"
@@ -12,7 +12,7 @@ es.indexer.create: false
 # ES metricsConsumer
 # Transport client will be used if at least one address is specified
 # Node client will be used otherwise
-# es.metrics.addresses: "localhost:9300"
+es.metrics.addresses: "localhost:9300"
 es.metrics.index.name: "metrics"
 es.metrics.doc.type: "datapoint"
 es.metrics.cluster.name: "elasticsearch"
@@ -31,9 +31,7 @@ es.metrics.ttl: -1
 # ES spout and persistence bolt
 # Transport client will be used if at least one address is specified
 # Node client will be used otherwise
-# es.status.addresses: 
-# - localhost:9300
-# - localhost:9301
+es.status.addresses: "localhost:9300"
 es.status.index.name: "status"
 es.status.doc.type: "status"
 es.status.cluster.name: "elasticsearch"
--- a/external/elasticsearch/pom.xml
+++ b/external/elasticsearch/pom.xml
@@ -18,7 +18,7 @@
 
 	<properties>
 		<!-- dependency versions -->
-		<elasticsearch.version>2.2.1</elasticsearch.version>
+		<elasticsearch.version>2.3.1</elasticsearch.version>
 	</properties>
 
 	<build>
--- a/external/elasticsearch/pom.xml
+++ b/external/elasticsearch/pom.xml
@@ -18,7 +18,7 @@
 
 	<properties>
 		<!-- dependency versions -->
-		<elasticsearch.version>1.7.2</elasticsearch.version>
+		<elasticsearch.version>2.2.1</elasticsearch.version>
 	</properties>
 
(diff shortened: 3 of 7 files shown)


w0mbat:
@jnioche great! So, period replacement was missing, right? I will definitely give it a try tomorrow.
Should I just close the other PR?

jnioche:
@w0mbat I also merged in the recent changes from the master branch as well. I think  #257 can be closed indeed.
Not sure why the period is not an issue with the metrics index though, we'll see.

w0mbat:
@jnioche looks good to me. Crawling and indexing works good.
One thing that looks pretty bad imo are the fields where `.` are replaced with `%2E`:
`metadata.url%2Epath` or `metadata.fetch%2Eerror%2Ecount` for example.
Any way we can fix it, because `metadata.depth` in comparison looks good again.

jnioche:
@w0mbat The reason why `metadata.depth` is correct is because metadata is an embedded object in the JSON representation whereas 'fetch.error' is a flat field name. There is no way to have a dot within the field name and generating structured objects e.g 'metadata -> fetch -> error' would be an overkill. Does not really matter how they are represented in ES as long as we get them in and out of it correctly.

mattburns:
Hi @jnioche ,
As you know, Elasticsearch 2.x no longer supports dots in fieldnames. Your workaround seems to be just html-encoding the dot character to get them in and out of ES. Isn't that just cheating? ;)
Why not replace them with a character that is allowed (like "_")?
-Matt

ndtreviv:
@jnioche Can you confirm that metadata is a nested object? Only I notice that in the `AggregationSpout` you're creating a `TermBuilder` on the field: `"metadata." + partitionField` without adding it to a nested Aggregator, eg:

```
NestedBuilder aggregations = AggregationBuilders.nested(partitionField).path("metadata");

            TermsBuilder termAggregation = AggregationBuilders.terms("partition")
                    .field("metadata." + partitionField).size(maxBucketNum);

aggregations.subAggregation(termAggregation);
```

If I create my mapping on my index treating `hostname` as a nested field in `metadata` it won't work without that nested Aggregator...

Am I getting confused?

jnioche:
> Why not replace them with a character that is allowed (like "_")?

what's wrong with encoding/decoding? the logic of how it is represented in ES is not that relevant and saves us from having to rename all the keys in SC that use a dot

ndtreviv:
Just to confirm, this is how we're creating the mapping in one of our integration tests:

```
client.admin().indices().preparePutMapping("status").setType("status")
    .setSource(jsonBuilder().prettyPrint().startObject()
        .startObject("status")
            .startObject("properties")
                .startObject("metadata")
                    .field("type", "nested")
                    .startObject("properties")
                        .startObject("hostname")
                            .field("type", "string")
                            .field("index", "not_analyzed")
                        .endObject()
                    .endObject()
                .endObject()
            .endObject()
       .endObject()
    .endObject())
    .execute().actionGet();
```

ndtreviv:
> what's wrong with encoding/decoding? the logic of how it is represented in ES is not that relevant and saves us from having to rename all the keys in SC that use a dot

It does mean that if you're querying the data you have to change the query as well. But given that you'd have to change the way you accessed the data from the resulting SearchHit if it were nested, you're having to many a change anyway...

ndtreviv:
Also, FYI for other people finding this, if you've got data already in your status index, you can't simply upgrade your es and restart, because you get this:
```
java.lang.IllegalStateException: unable to upgrade the mappings for the index [status], reason: [Field name [error.cause] cannot contain '.']
```
You'll need to migrate the data.

ndtreviv:
I've been looking into migrating our data from `metadata.url.path` to `metadata.url%2Epath` so that we can install elasticsearch 2.x and restore a 1.x snapshot into it, or simply start it up pointing at the old data.

However, the query that I've been using to find documents where `metadata.url.path` exists also brings back documents that have already been "upgraded" (ie: had their `url.path` replaced with `url%2Epath`).

This is a problem for us, because it means we can't upgrade our data, which means we can't make it elasticsearch 2.x friendly without re-fetching everything, which in turn is a great argument for replacing the `%2E` separator with something a bit more appropriate.

I can't just starting up an es 2.x pointing at 1.x data - it won't start, complaining about the mappings having fields with '.' in.

I can't delete mappings and re-create, because es lacks that functionality.

I can't restore a 1.x snapshot into 2.x because it complains - again - about fields having the '.' in:
```
cannot restore index [status] because it cannot be upgraded]; nested: IllegalStateException[unable to upgrade the mappings for the index [status], reason: [Field name [error.cause] cannot contain '.']]
```

The only way is to mutate the data. Using knapsack to export data to JSON and modifying that is  a no-go because of the size of the data.  

And now, writing a program to iterate over the data looking for old-format documents and replacing with new doesn't work, because of the afore-mentioned reason (in querying I can't distinguish between them).

Shall I raise this as an issue?

ndtreviv:
Actually, I think you can ignore that inability to distinguish between the two on querying. I seem to be having another problem (https://discuss.elastic.co/t/write-consistency-on-a-single-node-cluster/56542/3).

jnioche:
FYI @ndtreviv dots in field names are back in the latest version of ES[https://www.elastic.co/blog/elasticsearch-2-4-0-released].

ndtreviv:
We've migrated all of our data now. Is stormcrawler going to be reverted to use dots again? (Hint: please say no)

PR #287, opened 2016-05-25
Improve FetcherBolt

XciD:
- Improve the rotation of Queues
- Check if an url is already on a queue

### DIFF
--- a/core/src/main/java/com/digitalpebble/storm/crawler/bolt/FetcherBolt.java
+++ b/core/src/main/java/com/digitalpebble/storm/crawler/bolt/FetcherBolt.java
@@ -17,43 +17,6 @@
 
 package com.digitalpebble.storm.crawler.bolt;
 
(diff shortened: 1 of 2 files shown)


jnioche:
Thanks @XciD. Could we split this into 2 separate PRs (and squash the commits if necessary)

Usually we control the unicity of the URLs at the spout level, but I suppose that with e.g. queue implementations you don't have any control on that. Is that your case?

I wonder what the overhead of checking for duplicates is, any idea?

XciD:
In my case I've multiple spout so I can control on a spout but not all (maybe with a redis). 
No idea about the overhead, My queues size are low so I think there is no impact...

jnioche:
Do you have multiple instances of the same type of spout or multiple spouts of different types?

XciD:
Differents types, in fact I've a SeedSpout and an OutlinkSpout

PR #288, opened 2016-05-25
FetcherBolt : real loop in FetchItemQueues in order to rotate completly the map 

XciD:
Provide a real loop in FetchItemQueues in order to rotate completly the map and not restart a the begin on each new entry

### DIFF
--- a/core/src/main/java/com/digitalpebble/storm/crawler/bolt/FetcherBolt.java
+++ b/core/src/main/java/com/digitalpebble/storm/crawler/bolt/FetcherBolt.java
@@ -17,43 +17,6 @@
 
 package com.digitalpebble.storm.crawler.bolt;
 
-import java.io.File;
-import java.net.InetAddress;
-import java.net.MalformedURLException;
-import java.net.URL;
-import java.net.UnknownHostException;
-import java.text.SimpleDateFormat;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Deque;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Map.Entry;
-import java.util.concurrent.LinkedBlockingDeque;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.atomic.AtomicLong;
-
-import org.apache.commons.lang.StringUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import com.digitalpebble.storm.crawler.Metadata;
-import com.digitalpebble.storm.crawler.filtering.URLFilters;
-import com.digitalpebble.storm.crawler.persistence.Status;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/digitalpebble/storm/crawler/bolt/FetcherBolt.java
```diff
@@ -336,42 +324,61 @@ public synchronized FetchItemQueue getFetchItemQueue(String id) {
...
                 return null;
+            }
 
             FetchItemQueue start = null;
 
             do {
-                FetchItemQueue fiq = queues.get(it.next());
+                Map.Entry<String, FetchItemQueue> nextEntry;
+                FetchItemQueue fiq;
+                Iterator i = queues.entrySet().iterator();
```
jnioche: How costly is this?

You need to synchronize on the map when iterating over the entry set - [https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#synchronizedMap(java.util.Map)]

XciD: ```getFetchItem``` is already synchronized.

jnioche: > getFetchItem method is already synchronized.

of course

Would it make sense to create the iterator only once outside the do while loop? It would be more efficient and we would not need to keep track of the starting element at all?

XciD: I've just test both cases : 

Case 1 : 
```java
Iterator i = queues.entrySet().iterator();
Map<String, FetchQueue> tmpQueues = new LinkedHashMap<>();

try{
    do {
        // Stuff                
    
        // We remove the entry to reput at the end of the map
        i.remove();
        // Stuff
        tmpQueues.put(nextEntry.getKey(), nextEntry.getValue());
    
        // Stuff
    } while (i.hasNext());
    return null;
}finally {
    this.queues.putAll(tmpQueues);
}
```
Case 2 : 
```java
do {
    Iterator i = queues.entrySet().iterator();
    // Stuff                
    
    // We remove the entry to reput at the end of the map
    i.remove();
    // Stuff
    queues.put(nextEntry.getKey(), nextEntry.getValue());
    
    // Stuff
} while (!queues.isEmpty());
return null;
```

With 1000 Iterations, Case 2 is faster than one... about 10%

jnioche: Thanks for measuring this. Pretty convincing, will merge.

jnioche:
BTW have you been running this code for a while or is it relatively fresh?

jnioche:
@XciD thanks again for this contribution, much appreciated.

XciD:
Using since three or four day for now :)

jnioche:
Bug fixed in [55d6e46b3aa4df43d0494c1bf1937022206fb94e] where the start element could be removed if empty. Was getting an endless loop.

PR #289, opened 2016-05-25
FetcherBolt : check if we already have a FetchItem on a queue 

XciD:
Check if we already have a FetchItem on a queue.

### DIFF
--- a/core/src/main/java/com/digitalpebble/storm/crawler/bolt/FetcherBolt.java
+++ b/core/src/main/java/com/digitalpebble/storm/crawler/bolt/FetcherBolt.java
@@ -185,6 +185,24 @@ public class FetcherBolt extends BaseRichBolt {
             return new FetchItem(url, u, t, queueID);
         }
 
+        /**
+         * Used on FetchItemQueue when adding a new FetchItem on Set.
+         *  If the they have the same url then the FetchItem is already on the Set
+         * @param object
+         * @return
+         */
+        @Override
+        public boolean equals(Object object) {
+            if (object == null)
+                return false;
+            if (object == this)
+                return true;
+            if (!(object instanceof FetchItem))
+                return false;
+
+            return ((FetchItem) object).url.equals(this.url);
+        }
+
     }
 
     /**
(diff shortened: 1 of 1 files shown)


jnioche:
Continuing the discussion from #287 

>>>> Usually we control the unicity of the URLs at the spout level, but I suppose that with e.g. queue implementations you don't have any control on that. Is that your case?

>>> In my case I've multiple spout so I can control on a spout but not all (maybe with a redis). 

>> Do you have multiple instances of the same type of spout or multiple spouts of different types?

> Differents types, in fact I've a SeedSpout and an OutlinkSpout

what backends do they connect to? e.g. RabbitMQ queues?

I assume you populate whichever backend is used by the OutlinkSpout with the tuples with a status value of DISCOVERED from the status stream.  The seed spout is probably populated from an external source and you want to revisit them periodically. Am I right so far?

If you extend AbstractStatusUpdaterBolt to populate the backend with the outlinks you will get deduplication thanks to the internal cache, however as it is limited in size this does not guarantee that you won't get duplicates at all **but** neither do you get that with this PR \: a URL could have been just pulled from the internal queues before being added again.  

Out of curiosity, what sort of things do you crawl for?

If your crawl is recursive, why don't you use a different backend for your seeds and discovered links like for instance Elasticsearch?

XciD:
Both are connected to ES. 

Seed index are save seed to crawl periodically
Outlink index is used as buffer. I'm requesting and deleting from ES using your bucket configuration in order to have one url by host.

I also handle a QueuesHandler on the spout to offer a rotation of host in nextTuple.

For now I'm just looking about opensource crawler for my company

jnioche:
So if I understand correctly the problem you are trying to solve here is when the same URL comes from both your seed spout and the outlink spout. How likely is this to happen? In most cases people would not encounter this situation so I am a bit reluctant to make the code more complicated than necessary.

> I also handle a QueuesHandler on the spout to offer a rotation of host in nextTuple.

Which implementation of ES Spout are you using? [AggregationSpout](https://github.com/DigitalPebble/storm-crawler/blob/master/external/elasticsearch/src/main/java/com/digitalpebble/storm/crawler/elasticsearch/persistence/AggregationSpout.java)?

> For now I'm just looking about opensource crawler for my company

Great, feel free to share your thoughts when you're done exploring the various options. Always interesting to get feedback.

jnioche:
Closing for now as no progress. Feel free to reopen

PR #318, opened 2016-07-20
Update Jackson Databind Version to newer version

king-of-the-ring:
Some other Maven Modules depend on the new Jackson Databind.
All tests have passed.


jnioche:
Thanks @king-of-the-ring

king-of-the-ring:
@jnioche When is it expected to update Maven repo. ?

https://mvnrepository.com/artifact/com.digitalpebble.stormcrawler/storm-crawler-core

jnioche:
Not before septembre i think. We just released 1.0.

PR #322, opened 2016-08-11
Add information how to create index patterns in Kibana

sebastian-nagel:


jnioche:
thanks @sebastian-nagel

Issue #327, opened 2016-08-31
adding a dateLastProcessed field?


owenrh:
Hi,

We're looking to add some additional diagnostic capability to the _status_ index by adding a _dateLastProcessed_ field.  

We could push it in as a metadata field in our own crawler code, but we were wondering if it would make sense as a top-level field that most storm-crawler users would benefit from?

jnioche:
Thanks @owenrh (and great to see that you joined @mattburns and his gang)

Would that value be different from the fetch date? If so it could be added by the AbstractStatusUpdaterBolt, if not then it would be created by the Fetcher bolt implementations and be called e.g. lastFetchDate to mirror nextFetchDate. Having this could be useful for implementing an adaptive scheduling and it would be good for diagnostics indeed.

owenrh:
Thanks for getting back to me so quickly Julien.

Re: fetch date - it would be the timestamp of the last attempted fetch, so either _lastFetchDate_ or _lastProcessedDate_ would make sense. I think I suggested the latter just because I felt it didn't imply a successful fetch, which I thought _lastFetchDate_ might. (But maybe that's just a matter of documentation).

If you think it would be useful in the core then I'm happy to create a PR for it, and mod the AbstractStatusUpdaterBolt.

PR #328, opened 2016-09-05
Added metadata.lastProcessedDate to the status stream.

owenrh:


jnioche:
thanks @owenrh 
I'd rather not change the method signatures just for this and it should not be a mandatory feature. Didn't we mention a different name as well? It is not really when it was fetched but when the process was finished which is not quite the same.
What I had in mind was simply to add it to the metadata in the AbstractStatusUpdater class - users can decide whether to persist it in the storage by adding the key name to the config. It would end up being in the metadata instead of being a brand new field.
What do you think?

owenrh:
@jnioche sure, I think I misunderstood the convo we have on the [issue](https://github.com/DigitalPebble/storm-crawler/issues/327). The _metadata_ approach is definitely less intrusive. Also I thought you weren't too keen on the _lastProcessedDate_ field name, but I obviously got the wrong end of the stick. 

So you'd be cool with the metadata solution with a field name of _lastProcessedDate_? (If so , I'll just trash this PR and create a separate one to avoid having to reverse commits).

jnioche:
+1 Thanks

jnioche:
thanks @owenrh

PR #330, opened 2016-09-08
Spout and StatusUpdaterBolt for Redis

jnioche:
See README

This is pretty minimalistic but it is a good illustration on how to use a key/value store for handling the status of URLs and the crawl frontier. Redis does not need any configuration so it should be only a matter of installing it and run the topology. 

It could also be an acceptable solution for medium sized crawls where the URLS are not revisited but where we can have recursivity.


jnioche:
Haven't had much feedback on this one, am leaning towards putting it in a separate repo

PR #332, opened 2016-09-16
Fetch exception improvements

owenrh:
Addresses some issues where the fetch.exception gets written as the hostname rather than the actual error that occurred.

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/bolt/SimpleFetcherBolt.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/bolt/SimpleFetcherBolt.java
@@ -402,16 +402,19 @@ public class SimpleFetcherBolt extends BaseRichBolt {
                 message = "";
 
             // common exceptions for which we log only a short message
-            if (exece.getCause() instanceof java.util.concurrent.TimeoutException)
+            if (exece.getCause() instanceof java.util.concurrent.TimeoutException) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/digitalpebble/stormcrawler/bolt/SimpleFetcherBolt.java
```diff
@@ -402,16 +402,20 @@ public void execute(Tuple input) {
                 message = "";
 
             // common exceptions for which we log only a short message
-            if (exece.getCause() instanceof java.util.concurrent.TimeoutException)
+            if (exece.getCause() instanceof java.util.concurrent.TimeoutException) {
                 LOG.error("Socket timeout fetching {}", urlString);
-            else if (message.contains(" timed out"))
+                message = "Socket timeout fetching";
+            } else if (message.contains(" timed out")) {
```
jnioche: could have a OR on the condition above to simplify the code, just like we do for unknown host

jnioche:
thanks @owenrh

core/src/main/java/com/digitalpebble/stormcrawler/bolt/SimpleFetcherBolt.java
owenrh: Yeah, definitely. To be honest I was so busy putting in the message delta I didn't look at the original flow.

PR #333, opened 2016-09-16
Per day metrics indices

owenrh:
Hi Julien, I think I've addressed the headers, etc, that we discussed. Let me know if there's anything else.

### DIFF
--- a/external/elasticsearch/ES_IndexInit.sh
+++ b/external/elasticsearch/ES_IndexInit.sh
@@ -107,6 +107,51 @@ curl -XPOST localhost:9200/metrics -d '
 }
 '
 
+# uncomment below if using per day/month indices pattern
+#curl -XPOST localhost:9200/_template/storm-metrics-template -d '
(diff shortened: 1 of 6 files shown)


external/elasticsearch/ES_IndexInit.sh
```diff
@@ -107,6 +107,51 @@ curl -XPOST localhost:9200/metrics -d '
 }
 '
 
+# uncomment below if using per day/month indices pattern
+#curl -XPOST localhost:9200/_template/storm-metrics-template -d '
+#{
+#  "template": "metrics-*",
```
jnioche: can't we modify the template pattern so that it can be used on metrics*? In that case, we could uncomment the whole lot and remove the previous section

owenrh: Yeah, good call.

jnioche:
thanks @owenrh

PR #336, opened 2016-09-21
Fix last processed date for DISCOVERED

owenrh:
Fixes a bug in the original _metadata.lastProcessedDate_ code, where it is recording the last processed date for all URLs, even newly discovered outlinks.


owenrh:
Cool. I hadn't realised you'd made the subsequent changes. I'll put my code inline with yours.

jnioche:
thanks @owenrh, your code is not  up to date with the main repo and can't be merged. Could you open a fresh PR against the current version? (also this way we won't have the full history of the 4 commits for such a small change)

owenrh:
Sure. Will do.

PR #338, opened 2016-09-22
Fix for metadata.lastProcessedDate, should only be added when status …

owenrh:
…is not DISCOVERED. Take 2.


jnioche:
thanks @owenrh

PR #342, opened 2016-09-28
Removing politeness from robots denial

owenrh:
See title.


jnioche:
Makes sense. Thanks @owenrh

PR #349, opened 2016-10-06
ES Cache URLs post ack/fail

### LINKED ISSUE #340: Delay in updates to status index
Sebastian observed that his WARC files contained between 10-15% duplicates when crawling with ES. The most likely explanation is that there is a delay between the moment where an update is committed to the ES index and reflected in the search and when the corresponding tuple is acked (which happens a lot sooner). This issue is more likely on small crawls where the diversity of URLs is low.

One way around it is to set `es.status.min.delay.queries` to a larger value than the default 2 secs i.e. the minimal amount of time allowed between two queries to ES. By setting it to 60 secs, Seb saw a drop to  0.5-1% duplicates. This means that it can take more than a minute between the moment a tuple is acked and when the update is reflected in the search results.

Idea : we could remove the tuples from the beingProcessed hash after some additional time, this way we wouldn't refetch the same URL too soon. The various ES spouts have an overlap in code, we could create an abstract class and share the functionality there.

jnioche:
Fixes #340 

The abstract class contains an extension of Map for the URLs in process, this extension also contains a Guava cache which gets populated when entries are removed from the map.

The time during which the entries remain in the cache is configurable. We also report 2 separate metrics for the # of entries in the map and in the cache.

Not yet tested but reviews welcome as usual

FYI @sebastian-nagel

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/util/MetadataTransfer.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/util/MetadataTransfer.java
@@ -22,6 +22,7 @@ import java.util.Set;
 
 import org.apache.commons.lang.StringUtils;
 
+import com.digitalpebble.stormcrawler.Constants;
 import com.digitalpebble.stormcrawler.Metadata;
 
 import clojure.lang.PersistentVector;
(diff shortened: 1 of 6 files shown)


jnioche:
2nd commit made to the wrong branch by mistake

PR #355, opened 2016-10-12
Handle redirections via meta tag

### LINKED ISSUE #350: Handle redirections via meta tag
We currently don't handle redirs in html, a page such as the one below gets treated as fetched and gets indexed with an empty content, outlinks do not get discovered

```
 <html>    
  <head>      
    <meta http-equiv="refresh" content="0;URL=http://www.apollocolors.com/site" />    
  </head>    
</html>     
```

at the very least we should treat the link as discovered but ideally the main URL should get REDIRECTIOn

jnioche:
With JSoup, fixes #350


jnioche:
Replaced with #362 (where commits have been squashed)

PR #360, opened 2016-10-13
Add metadata.discoveryDate field to enable tracking discovery rate/gr…

owenrh:
…owth against hosts/domain.


jnioche:
Great idea! Thanks

PR #373, opened 2016-11-02
Improved coverage of HTML redirections

prokopidis:


jnioche:
Thanks @prokopidis 
I had put the tests in JSoupParserBoltTest but it makes sense to have them separately as this could be used also from Tika. Will add the missing license header in a separate commit

PR #407, opened 2017-01-12
Add adaptive fetch scheduler

sebastian-nagel:
Add implementation of Scheduler which 
- checks fetch status (304 not modified) or compares previous and current signature
  to decide whether a page has changed since the previous fetch
- if changed, decreases the fetch interval
- otherwise increases it up to a configurable limit

See also commoncrawl/news-crawl#15.

Possible TODOs:
- add unit tests
- simplify configuration
  - interaction with parse filters to provide signatures (old and new)
  - reduce amount of persisted metadata (need metadata to be kept in the topology but not necessarily persisted in storage)

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/parse/filter/MD5SignatureParseFilter.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/parse/filter/MD5SignatureParseFilter.java
@@ -44,7 +44,7 @@ import com.fasterxml.jackson.databind.JsonNode;
  * <dt>keyNameCopy</dt>
  * <dd>name of the metadata field to hold a temporary copy of the signature used
  * to decide by signature comparison whether the document has changed. If not
- * defined, the signature is not copied.</dd>
+ * defined or empty, the signature is not copied.</dd>
  * </dl>
  *
  */
@@ -63,8 +63,9 @@ public class MD5SignatureParseFilter extends ParseFilter {
         Metadata metadata = parseData.getMetadata();
         if (copyKeyName != null) {
             String signature = metadata.getFirstValue(key_name);
-            if (signature != null)
+            if (signature != null) {
                 metadata.setValue(copyKeyName, signature);
+            }
         }
         byte[] data = null;
         if (useText) {
@@ -94,9 +95,9 @@ public class MD5SignatureParseFilter extends ParseFilter {
             key_name = node.asText("signature");
         }
         node = filterParams.get("keyNameCopy");
-        if (node != null && node.isTextual()) {
+        if (node != null && node.isTextual()
+                && StringUtils.isNotBlank(node.asText(""))) {
             copyKeyName = node.asText("signatureOld");
-            System.out.println("HERE: signatureOld" + copyKeyName);
         }
     }
 
--- a/core/src/main/java/com/digitalpebble/stormcrawler/parse/filter/MD5SignatureParseFilter.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/parse/filter/MD5SignatureParseFilter.java
@@ -33,6 +33,20 @@ import com.fasterxml.jackson.databind.JsonNode;
 /**
  * Computes a signature for a page, based on the binary content or text. If the
  * content is empty, the URL is used.
+ *
+ * Configuration properties:
+ * <dl>
+ * <dt>useText</dt>
+ * <dd>compute signature on plain text, instead of binary content</dd>
+ * <dt>keyName</dt>
+ * <dd>name of the metadata field to hold the signature (default:
+ * &quot;signature&quot;)</dd>
+ * <dt>keyNameCopy</dt>
+ * <dd>name of the metadata field to hold a temporary copy of the signature used
+ * to decide by signature comparison whether the document has changed. If not
+ * defined, the signature is not copied.</dd>
+ * </dl>
+ *
  */
 public class MD5SignatureParseFilter extends ParseFilter {
 
@@ -40,11 +54,18 @@ public class MD5SignatureParseFilter extends ParseFilter {
 
     private boolean useText = false;
 
+    private String copyKeyName = null;
+
     @Override
     public void filter(String URL, byte[] content, DocumentFragment doc,
             ParseResult parse) {
         ParseData parseData = parse.get(URL);
         Metadata metadata = parseData.getMetadata();
+        if (copyKeyName != null) {
+            String signature = metadata.getFirstValue(key_name);
+            if (signature != null)
+                metadata.setValue(copyKeyName, signature);
+        }
         byte[] data = null;
         if (useText) {
             String text = parseData.getText();
@@ -71,6 +92,11 @@ public class MD5SignatureParseFilter extends ParseFilter {
         node = filterParams.get("keyName");
         if (node != null && node.isTextual()) {
             key_name = node.asText("signature");
+        }
+        node = filterParams.get("keyNameCopy");
+        if (node != null && node.isTextual()) {
+            copyKeyName = node.asText("signatureOld");
+            System.out.println("HERE: signatureOld" + copyKeyName);
         }
     }
 
--- a/core/src/main/java/com/digitalpebble/stormcrawler/persistence/AdaptiveScheduler.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/persistence/AdaptiveScheduler.java
@@ -1,0 +1,322 @@
+/**
+ * Licensed to DigitalPebble Ltd under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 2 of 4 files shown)


core/src/main/java/com/digitalpebble/stormcrawler/protocol/AbstractHttpProtocol.java
```diff
@@ -96,4 +99,14 @@ protected static String getAgentString(String agentName,
         return buf.toString();
     }
 
+    /**
+     * Return new instance of {@link SimpleDateFormat} to format dates in HTTP
+     * headers, cf. <a href=
+     * "https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.3">sec3.3 in
+     * RFC 2616</a>.
+     */
+    public static SimpleDateFormat getHttpDateFormat() {
```
jnioche: unless this is used by any of the code in the protocol package, this should be put somewhere else e.g. within the AdaptiveScheduler class itself?

core/src/main/java/com/digitalpebble/stormcrawler/parse/filter/SignatureCopyParseFilter.java
```diff
@@ -0,0 +1,61 @@
...
+import com.digitalpebble.stormcrawler.Metadata;
+import com.digitalpebble.stormcrawler.parse.ParseData;
+import com.digitalpebble.stormcrawler.parse.ParseFilter;
+import com.digitalpebble.stormcrawler.parse.ParseResult;
+import com.fasterxml.jackson.databind.JsonNode;
+
+/**
+ * Adds a copy of the current signature to metadata.
+ */
+public class SignatureCopyParseFilter extends ParseFilter {
```
jnioche: No need for a separate ParseFilter, let's add a config element to the existing Signature one so that it makes a copy of any preexisting value before generating a new one.

jnioche:
>  reduce amount of persisted metadata (need metadata to be kept in the topology but not necessarily persisted in storage)

The filtering could be done between scheduling and storing

[https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/java/com/digitalpebble/stormcrawler/persistence/AbstractStatusUpdaterBolt.java#L143]

e.g you would not need to persist 'signatureold' but would still need 'signature'

core/src/main/java/com/digitalpebble/stormcrawler/protocol/AbstractHttpProtocol.java
sebastian-nagel: The date format is used to fill the metadata field "last-modified" which is used to send If-Modified-Since requests. That's the reason why it's in the protocol package.

core/src/main/java/com/digitalpebble/stormcrawler/parse/filter/SignatureCopyParseFilter.java
sebastian-nagel: +1 good idea

core/src/main/java/com/digitalpebble/stormcrawler/persistence/AdaptiveScheduler.java
```diff
@@ -0,0 +1,321 @@
...
+            metadata.remove(FETCH_INTERVAL_KEY);
+            metadata.remove(SIGNATURE_KEY);
+            metadata.remove(SIGNATURE_OLD_KEY);
+
+            // fall-back to DefaultScheduler
+            return super.schedule(status, metadata);
+        }
+
+        Calendar now = Calendar.getInstance(Locale.ROOT);
+
```
jnioche: what about 
```
final String now_string = httpDateFormat.format(now.getTime());
```

and use it below instead of 

`httpDateFormat.format(now.getTime())`

core/src/main/java/com/digitalpebble/stormcrawler/protocol/AbstractHttpProtocol.java
jnioche: Sure, but it is not used at all within the package. We don't check that the values sent in the protocol match this format, we just take the Strings as granted. I understand the logic but it is more practical and understandable to put it in the scheduler class I think.

core/src/main/java/com/digitalpebble/stormcrawler/persistence/AdaptiveScheduler.java
sebastian-nagel: Yes, of course. That's more straightforward.

core/src/main/java/com/digitalpebble/stormcrawler/protocol/AbstractHttpProtocol.java
sebastian-nagel: Ok, I'll move it to the scheduler and update the pr. Thanks for the review.

core/src/main/java/com/digitalpebble/stormcrawler/parse/filter/MD5SignatureParseFilter.java
```diff
@@ -72,6 +93,11 @@ public void configure(Map stormConf, JsonNode filterParams) {
         if (node != null && node.isTextual()) {
             key_name = node.asText("signature");
         }
+        node = filterParams.get("keyNameCopy");
+        if (node != null && node.isTextual()) {
+            copyKeyName = node.asText("signatureOld");
+            System.out.println("HERE: signatureOld" + copyKeyName);
```
jnioche: yuk! a System.out! Either remove or use a slf4j logger here

core/src/main/java/com/digitalpebble/stormcrawler/parse/filter/MD5SignatureParseFilter.java
```diff
@@ -33,18 +33,39 @@
...
     private boolean useText = false;
 
+    private String copyKeyName = null;
+
     @Override
     public void filter(String URL, byte[] content, DocumentFragment doc,
             ParseResult parse) {
         ParseData parseData = parse.get(URL);
         Metadata metadata = parseData.getMetadata();
+        if (copyKeyName != null) {
```
jnioche: StringUtils.isNotBlank would be better\: the value could be an empty String or spaces, we wouldn't want that as key and it probably isn't what the user intended to set

core/src/main/java/com/digitalpebble/stormcrawler/parse/filter/MD5SignatureParseFilter.java
```diff
@@ -33,18 +33,39 @@
...
+            String signature = metadata.getFirstValue(key_name);
+            if (signature != null)
```
jnioche: while we're on this one, I tend to put curly braces around single lines but this is a matter of personal preference

jnioche:
Why not move down https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/java/com/digitalpebble/stormcrawler/persistence/AbstractStatusUpdaterBolt.java#L143 at the same time? This would simplify things a bit (see comment above)

core/src/main/java/com/digitalpebble/stormcrawler/parse/filter/MD5SignatureParseFilter.java
sebastian-nagel: Sorry, that was for debugging to see whether the test for a non-existent configuration property works. I wanted to avoid two properties (a boolean whether to copy the signature and the key name).

core/src/main/java/com/digitalpebble/stormcrawler/parse/filter/MD5SignatureParseFilter.java
sebastian-nagel: Agreed, but a check for an empty strings could be made already in the configure method, not for any call of the filter method.

core/src/main/java/com/digitalpebble/stormcrawler/parse/filter/MD5SignatureParseFilter.java
sebastian-nagel: ok

sebastian-nagel:
Yes, of course.

sebastian-nagel:
PR updated. In doubt, I would test the final version of the adaptive scheduler in live. But that would a day or two to see whether fetch times / intervals are set as expected.

jnioche:
@sebastian-nagel 

>  In doubt, I would test the final version of the adaptive scheduler in live

that would be great. It's OK to fix things later if you find anything wrong as this is a brand new functionality and we won't release a new version just now

Thanks again for this

PS : any chance you could document how the adaptive scheduler works in the WIKI?

sebastian-nagel:
Yes, I'll do it. The Javadoc is a good starting point, but add a concrete example will surely make it more comprehensive.

PR #423, opened 2017-02-06
custom scheduling is now deterministic

### LINKED ISSUE #420: DefaultSchedule: configuration of custom schedule with status and metadata not deterministic
Custom schedule rules based on status and metadata are prioritized by their order in the configuration file (see [1](https://github.com/DigitalPebble/storm-crawler/commit/cf3e07f58e8656fa62a60112a2950e8c919408f1#diff-a2a0d55a1e9f8b09634a3874b7aa3633R64)). However, the order from the configuration file is not preserved, configuration properties in the configuration map passed to DefaultScheduler.init(Map stormConf) seem to be in random order.

The test units in [DefaultSchedulerTest]() also use a simple HashMap instead of a order-preserving structure. Renaming the metadata key (e.g., to "aTestKey") can make the tests fail.

jnioche:
Fixes #420 

We don't expect the entries to be in any particular order and the behaviour should now be deterministic.

@sebastian-nagel could you please have a look and test?

### DIFF
--- a/archetype/src/main/resources/archetype-resources/crawler-conf.yaml
+++ b/archetype/src/main/resources/archetype-resources/crawler-conf.yaml
@@ -61,8 +61,6 @@ config:
 
   # custom fetch interval to be used when a document has the key/value in its metadata
   # and has been fetched succesfully (value in minutes)
-  # the patterns are tested in sequential order, the most specific ones where a status 
-  # is set should be placed first.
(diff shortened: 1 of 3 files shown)


sebastian-nagel:
Looks good. Tested the fix, rules with metadata and status specified and fall-back rules without status are now applied as expected.

PR #431, opened 2017-03-07
Basic Authentication over Http Proxy

ndtreviv:
Modified HttpProtocol to optionally accept a proxy username, password and auth scheme. It currently only supports basic authentication because that's the only one I can test against.

Additional configuration parameters to leverage this functionality are:

```
http.proxy.user: [proxy username, if not present authentication is not used]
http.proxy.pass: [proxy password, if not present authentication is not used]
http.proxy.authscheme: [proxy auth scheme, defaults to Basic, only supports Basic]
```

Also removed a throws Exception because it never throws that exception, and cleaned up the builder usage code.


core/src/main/java/com/digitalpebble/stormcrawler/protocol/httpclient/HttpProtocol.java
```diff
@@ -99,24 +107,42 @@ public void configure(final Config conf) {
         String proxyHost = ConfUtils.getString(conf, "http.proxy.host", null);
         int proxyPort = ConfUtils.getInt(conf, "http.proxy.port", 8080);
 
+        String proxyUser = ConfUtils.getString(conf, "http.proxy.user", null);
+        String proxyPass = ConfUtils.getString(conf, "http.proxy.pass", null);
+        String proxyAuthScheme = ConfUtils.getString(conf, "http.proxy.authscheme", AuthSchemes.BASIC);
+        List<String> authSchemes = Lists.newArrayList();
```
jnioche: `= new ArrayList<>();`

takes even less characters and does not require the import of the shaded lib

core/src/main/java/com/digitalpebble/stormcrawler/protocol/httpclient/HttpProtocol.java
```diff
@@ -29,12 +31,16 @@
...
 import org.apache.http.HttpResponse;
 import org.apache.http.StatusLine;
+import org.apache.http.auth.AuthScope;
+import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.client.ClientProtocolException;
 import org.apache.http.client.ResponseHandler;
+import org.apache.http.client.config.AuthSchemes;
 import org.apache.http.client.config.CookieSpecs;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.config.RequestConfig.Builder;
```
jnioche: Can be removed now, same for ClientProtocolException

core/src/main/java/com/digitalpebble/stormcrawler/protocol/httpclient/HttpProtocol.java
```diff
@@ -99,24 +107,42 @@ public void configure(final Config conf) {
...
-        requestConfigBuilder.setConnectTimeout(timeout);
-        requestConfigBuilder.setConnectionRequestTimeout(timeout);
-        requestConfigBuilder.setCookieSpec(CookieSpecs.STANDARD);
-        requestConfig = requestConfigBuilder.build();
+        requestConfig = RequestConfig.custom()
+                .setSocketTimeout(timeout)
+                .setConnectTimeout(timeout)
+                .setConnectionRequestTimeout(timeout)
+                .setCookieSpec(CookieSpecs.STANDARD)
+                .setProxyPreferredAuthSchemes(authSchemes)
```
jnioche: what happens if authSchemes is empty? line 122 above

as suggested - let's have just the basic scheme. If so, what value would be used if we did not call this method?

core/src/main/java/com/digitalpebble/stormcrawler/protocol/httpclient/HttpProtocol.java
jnioche: Let's keep the code simple for now\ : since we support only the basic scheme let's just have it. We'll change the code later if/when other schemes are supported

core/src/main/java/com/digitalpebble/stormcrawler/protocol/httpclient/HttpProtocol.java
```diff
@@ -99,24 +107,42 @@ public void configure(final Config conf) {
...
+        List<String> authSchemes = Lists.newArrayList();
+
         boolean useProxy = proxyHost != null && proxyHost.length() > 0;
 
         // use a proxy?
         if (useProxy) {
+
+            if (StringUtils.isNotBlank(proxyUser) && StringUtils.isNotBlank(proxyPass)) {
+                if (!SUPPORTED_AUTH_SCHEMES.contains(proxyAuthScheme)) {
+                    LOG.error("Configured proxy auth scheme: {} is not supported. Proxy authentication will not be used", proxyAuthScheme);
```
jnioche: see comment above about just using basic scheme

core/src/main/java/com/digitalpebble/stormcrawler/protocol/httpclient/HttpProtocol.java
```diff
@@ -99,24 +107,42 @@ public void configure(final Config conf) {
...
+                    basicAuthCreds.setCredentials(new AuthScope(proxyHost, proxyPort),
+                            new UsernamePasswordCredentials(proxyUser, proxyPass));
+                    builder.setDefaultCredentialsProvider(basicAuthCreds);
+                }
+            }
+
             HttpHost proxy = new HttpHost(proxyHost, proxyPort);
-            DefaultProxyRoutePlanner routePlanner = new DefaultProxyRoutePlanner(
-                    proxy);
+            DefaultProxyRoutePlanner routePlanner = new DefaultProxyRoutePlanner(proxy);
```
jnioche: Formatting ;-) did you run `mvn java-formatter:format` before committing?

ndtreviv: D'oh!

ndtreviv: It's SO thin!

ndtreviv:
@jnioche Do you mean add them to crawler-conf.yaml? Or crawler-defaults.yaml? I can't find the other http.proxy params in there, so not sure if I'm looking in the right place.

As for the wiki, I haven't had to do that before, so am still working out how! It looks like it's a separate repo or something? I'll figure it out and have a go tomorrow.

Otherwise I've made all suggested changes.

core/src/main/java/com/digitalpebble/stormcrawler/protocol/httpclient/HttpProtocol.java
jnioche: I'm happy to change for longer lines (or shorter ones) What matters is that we don't get stuff in diffs that are due to formatting discrepancies

jnioche:
Could you remove MD5SignatureParseFilter.java from this PR? The formatting's probably off (although I ran the formatter yesterday and it did not get picked up) but it's outside the topic of the PR

Could you squash all commits into a single one? 

[crawler-default.yaml](https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/resources/crawler-default.yaml) could you please add the other http.proxy params commented out? 

WIKI -> https://github.com/DigitalPebble/storm-crawler.wiki.git
never used it as I tend to do the changes online but should work

Thanks a lot!

ndtreviv:
@jnioche Gah! It turns out I cannot git today. I'm going to cancel this and submit another one. 
Also, it looks like when we forked it didn't fork the wiki, and I can't find out how to do that as a subsequent task in order to add that to the PR. Sorry.

jnioche:
> Also, it looks like when we forked it didn't fork the wiki, and I can't find out how to do that as a subsequent task in order to add that to the PR. Sorry.

Probably can't be done as part of the PR. no probs!

PR #432, opened 2017-03-08
HTTP Proxy over Basic Authentication

ndtreviv:
Modified HttpProtocol to optionally accept a proxy username, password and auth scheme. It currently only supports basic authentication because that's the only one I can test against.

Additional configuration parameters to leverage this functionality are:
```
http.proxy.user: [proxy username, if not present authentication is not used]
http.proxy.pass: [proxy password, if not present authentication is not used]
```
Also removed a throws Exception because it never throws that exception, and cleaned up the builder usage code.


jnioche:
Thanks @ndtreviv

Issue #32, opened 2014-12-04
Add handling of cookies to HTTP protocol


jnioche:
The cookies info could be stored in the metadata (and inherited from the parent URL) and used by the Protocol implementation.

foromer4:
I would like to implement this. @jnioche, could you please add some info on what classes should be changed to save and read these stored cookies? thanks

jnioche:
Here is how I see it. You should only need to modify things at the protocol level. Basically, the cookies are stored as text in the metadata of the parent URL, transferred to the outlinks which get persisted to the status in the usual ways, see https://github.com/DigitalPebble/storm-crawler/wiki/MetadataTransfer. 

When the tuple for an outlink is sent down the topology, we'll need to convert the info from the metadata into cookies in the [getProtocolOutput() method](https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/java/com/digitalpebble/stormcrawler/protocol/httpclient/HttpProtocol.java#L124) so that the low level protocol stuff sends it.

> from what I can tell. JSoupParserBolt might be the place to detect cookies.
Nope. Cookies are nothing more than http headers with an attitude.

> However, I'm not sure how to do it only once for each website?
We don't. We could have different cookies information for different URLs belonging to the same host or domain.

> Does Storm crawler keep info to connect each url to a specific parent url? 
Part of the optional path info in the metadatam see Wiki page about MetadataTransfer

> ANd does it store information for the parent url?
Like any other URL, nothing more.

Hope this helps

foromer4:
@jnioche , So you want the crawler to visit a ceratin page and collect from the response the cookies, and then use the same cookies for discovered links? and if it find other cookies in the discovered links add them also to their "child links" and so on?

Might it be beneficial to set the desired cookies once per site , or per regex in the config, instead of passing them from parent to child?

jnioche:
> if it find other cookies in the discovered links add them also to their "child links" and so on?

they would replace the existing ones instead of adding to them I think

> Might it be beneficial to set the desired cookies once per site , 

there is no global repository for sites info in SC

> or per regex in the config, instead of passing them from parent to child?

you could set them as seed metadata instead of keeping them in the config and have an option in the protocol implementation to NOT replace an existing value. This would be more flexible I think as you could modify it on the fly instead of relaunching the topology

which problem are you trying to solve or to put it differently, any reason you think my suggestion is a bad idea? 

Thanks

foromer4:
[...]

PR #446, opened 2017-03-29
Issue #32 Add handling of cookies to HTTP protocol  - implemented

foromer4:
Implemented http cookies in response based on request,
using CookieConverter from nutch as a refrence, but with some changes
to support domain and path better.

Still need to add the flag : `http.use.cookies` to the documentation.
Also a few issued I encountered and some suggestions:
* When running the maven formatting it changed many files so I didn't commit those format changes.
* It seems worthwhile to add to the [Getting started page](http://stormcrawler.net/getting-started/) an explanation on how to debug the toplogy in the IDE itself.
* Squashing all comiits caused some issues, might be better to do this on the [merge ](https://help.github.com/articles/about-pull-request-merges/)iteslt.
* Finally, a recursive crawler is such a common scenario, it might be worthwhile to add a sample topology to the archtype. I had to do some tweaking for this change in order to be able to test it on a recursive crawler.

Thanks!
Omer


jnioche:
Hi @foromer4 

Thanks for this, please find comments below

> using CookieConverter from nutch as a refrence

it does not come from Apache Nutch but had been written as a bespoke development based on Nutch. Let's remove references to Nutch as no credit is due for this particular functionality.

> Still need to add the flag : http.use.cookies to the documentation.

yes, we can modify the WIKI post merging

> When running the maven formatting it changed many files so I didn't commit those format changes.

Strange. What differences did you find? could you apply the formatting only to the files you changed?

> It seems worthwhile to add to the Getting started page an explanation on how to debug the toplogy in the IDE itself.

Why not, if you found it useful, chances are that others might too.

> Squashing all comiits caused some issues, might be better to do this on the merge iteslt.

Your PR can't be merged as-is. I'd prefer it if you opened a new clean one against the main branch if that's OK.

> Finally, a recursive crawler is such a common scenario, it might be worthwhile to add a sample topology to the archtype. I had to do some tweaking for this change in order to be able to test it on a recursive crawler.

I don't think it would be worth adding another topo just for this but we could modify the example one so that it uses the MemoryUpdaterBolt. Feel free to open a new issue for this.

Could you please open add licenses on top of new files?

I like the work you did on the test classes.

None of our existing tests requires an internet connection. If you want to test a URL manually, you can run the protocol class instead. Can we remove the ones where you connect to an external server?

Thanks

jnioche:
@foromer4 see https://github.com/DigitalPebble/storm-crawler/wiki/Debug-with-Eclipse

foromer4:
Thank you Julien

PR #449, opened 2017-03-30
Resolve issue #32 - implement cookie handeling

foromer4:
OK , trying again, with all requested adjustments.

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/protocol/AbstractHttpProtocol.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/protocol/AbstractHttpProtocol.java
@@ -35,11 +35,14 @@ public abstract class AbstractHttpProtocol implements Protocol {
 
     protected boolean storeHTTPHeaders = false;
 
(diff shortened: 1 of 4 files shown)


foromer4:
I still see the formatting issues, as I suspected - they are caused by the formatter plugin, when I run it it changes formatting in all files. 
I just commit the 4 changed files, but they too are formatted in a way that the dif tools don't understand. 

Might it be related to the fact I'm running on windows?
some issue like this [one ](http://stackoverflow.com/questions/1889559/git-diff-to-ignore-m) maybe?

jnioche:
thanks @foromer4, will apply the formatting in a separate commit.

I'll also add a conf file to the repo to specify that we want line endings to be linux style. I hadn't realised that the Eclipse formatting rules did not cover that at all.  See https://help.github.com/articles/dealing-with-line-endings/#platform-all

Thanks again for your contribution and your patience

foromer4:
thank you @jnioche  for your help and guidance - hopefully next time would go smoother

PR #482, opened 2017-07-03
Ip rotate

XciD:
IP Rotation for storm crawler


jnioche:
To do this with OKHttp -> https://stackoverflow.com/questions/30118534/global-state-and-local-address-in-okhttp

jnioche:
Thanks @XciD. What about having a new protocol implementation which simply creates N instances of the existing one, each with their own config and round robin on them when it gets a request?
This would keep the code simpler by not modifying the FetcherBolt + would work with the SimpleFetcherBolt.

PR #487, opened 2017-07-18
Fixing max fetch error count comparison so it can cope with configura…

owenrh:
…tion changes to max fetch errors, and not miss counts that have gone above the threshold.

It is currently possible with configuration changes to get scenarios where the fetch error count against a URL exceeds the _maxFetchErrors_ configuration parameter. This is due to the current comparator using _exactly equal to_.

This fix just uses _greater than or equal_ instead, so you can never skip over the threshold and get _bad_ documents cycling through the crawl.


jnioche:
Makes sense. Thanks @owenrh

Issue #491, opened 2017-08-17
Add timezone to date-format used in AbstractStatusUpdaterBolt


tendres1980:
The date-format used in the AbstractStatusUpdaterBolt does not specify a timezone. The format is used for the metadata 'discoveryDate' and 'lastProcessedDate'. When trying to store these dates for example in an elastic index, this leads to problems since the default timezone for elastic is UTC. When these dates are stored in elastic (using a date-format) for a different timezone, the dates are no longer correct. 

For example I am running the storm-crawler on a machine with CEST timezone, getting a date like '2017-08-17T14:16:00'. When looking at the date in elastic through kibana it shows as 'August 17th 2017, 16:16:00.000' which is the original date interpreted as UTC.

jnioche:
Hi @tendres1980. Thanks for opening this issue. any reason why you closed it? I see you created #492 so I presume this is still valid. If so please reopen it - it should be closed only when the problem is solved or invalid. Thanks!

jnioche:
Actually not related to #492 and probably closed as fixed in master branch in 80098788f4733f9f416135f81e8aabff1a648602
duplicates #477

PR #492, opened 2017-08-17
set a standard timezone for dateformat in adaptivescheduler

tendres1980:
Set a standard timezone in AdaptiveScheduler since not all timezones (e.g. "CEST") are supported by joda-format e.g. used by elasticsearch


jnioche:
Fixes #491

jnioche:
We could also convert the Calendar instance to an Instant instead of a Date and use its toString() representation which follows ISO-8601 and is in UTC.
What do you think @tendres1980 ?

jnioche:
Implemented in #f07c282
Thanks @tendres1980 for reporting the issue and suggesting a solution.

PR #520, opened 2017-12-14
Upgrade WARC module to latest version of storm-hdfs, fixes #590

### LINKED ISSUE #590: ES change value of es.status.reset.fetchdate.after
By default it is set to -1, which means that the same nextFetchDate will be reused as long as it returns results. If a single site has loads of URLs scheduled at the same time, e.g. coming from a sitemap, only URLs from that site will be fetched. 

Will change to 120 secs by default, can always be changed by the users.

sebastian-nagel:
Tested so far in small scale:
  - WARC files validate
  - file rotation by size works
A large scale test will follow during the next days.

A first trial to implement a class GzipHdfsWriter extending AbstractHDFSWriter (AbstractHDFSWriter needs to be returned by `AbstractHdfsBolt.makeNewWriter(path, tuple)`) wasn't successful:
- the field `offset` in AbstractHDFSWriter is only visible in the package `org.apache.storm.hdfs.common`, it cannot be used in a Storm-crawler package
-  the method `write(tuple)` which accesses the package-internal `offset` is final:
```
abstract public class AbstractHDFSWriter {

    long offset;

    final public long write(Tuple tuple) throws IOException{
        doWrite(tuple);
        this.needsRotation = rotationPolicy.mark(tuple, offset);

        return this.offset;
    }
}
```

I didn't find a way to operate directly on the output stream and offset while still extending the classes in `org.apache.storm.hdfs.common`. That's why compression is moved to a GzipRecordFormat. As a consequence, the option to do the rotation on uncompressed offsets (content length) is dropped.

### DIFF
--- a/external/warc/pom.xml
+++ b/external/warc/pom.xml
@@ -32,7 +32,7 @@
 		<dependency>
 			<groupId>org.apache.storm</groupId>
 			<artifactId>storm-hdfs</artifactId>
-			<version>1.0.3</version>
+			<version>1.1.0</version>
 		</dependency>
 
 	</dependencies>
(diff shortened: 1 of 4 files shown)


jnioche:
looks good, thanks @sebastian-nagel

PR #522, opened 2017-12-21
BasicURLNormalizer to optionally convert IDN host names to ASCII/Punycode

sebastian-nagel:
- conversion of IDN host names to their ASCII/Punycode equivalent is triggered if param "hostIDNtoASCII" is true
- avoids duplicate page content
- get clean ASCII keys if sharding/routing is done by host or domain


jnioche:
Great stuff, thanks @sebastian-nagel

PR #524, opened 2017-12-29
Store content-length as field value and allow to use inner classes of FetchBolt

KarstenRauch:
Extra line in FetchBolt like discussed in
https://stackoverflow.com/questions/48013771/store-content-length-as-field-value-aka-in-metadata-of-indexed-docs

Switch from private to public inner classes in FetchBold because we need nearly the same function to crawle extra xml-sites with metadate for each pdf. So why not allow to use the same code?


jnioche:
No progress, closing for now

PR #543, opened 2018-03-07
Default URL filter: exclude localhost and private address spaces

sebastian-nagel:
To avoid that information is leaked and exposed in a public search interface, a web crawler should never follow links to localhost, loop-back addresses and private address spaces, see commoncrawl/news-crawl#21. These are excluded by regular expressions for http/https URLs.

Of course, could comment out the rules to enable test crawls of the local Apache web server.


jnioche:
thanks @sebastian-nagel

PR #575, opened 2018-05-28
status update: log error if elastic search reports an unexpected problem

cruftex:
Hi Julien,

this PR prints errors, if we get some in the callback from elastic search.
It happened to me that the ES index switched to read-only and I wondered why nothing gets updated, since there is no failure indication.

Cheers,

Jens


jnioche:
Thanks @cruftex

jnioche:
Modified and committed. Thanks @cruftex

cruftex:
Great! Sorry for not following up on this, too much on my plate at the moment.

PR #589, opened 2018-06-29
Add simple basic auth system

pierrehts:
Hello,

This is a simple basic auth system that allow user to add these parameters into the yaml configuration :

-   http.basicauth.enabled: true
-   http.basicauth.user: "root"
-   http.basicauth.password: "root"

This way it allows stormcrawler to be used in most of the intranets.

If this PR is accepted I'll produce the documentation associated.

Thanks

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/protocol/httpclient/HttpProtocol.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/protocol/httpclient/HttpProtocol.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.util.ArrayList;
+import java.util.Base64;
 import java.util.Collection;
 import java.util.LinkedList;
 import java.util.List;
@@ -54,7 +55,6 @@ import org.apache.http.message.BasicHeader;
 import org.apache.http.util.Args;
 import org.apache.http.util.ByteArrayBuffer;
 import org.apache.storm.Config;
-import java.util.Base64.Encoder;
 import org.slf4j.LoggerFactory;
 
 import com.digitalpebble.stormcrawler.Metadata;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/digitalpebble/stormcrawler/protocol/httpclient/HttpProtocol.java
```diff
@@ -54,6 +54,7 @@
 import org.apache.http.util.Args;
 import org.apache.http.util.ByteArrayBuffer;
 import org.apache.storm.Config;
+import org.apache.storm.shade.org.yaml.snakeyaml.external.biz.base64Coder.Base64Coder;
```
jnioche: I'd rather we relied on non-shaded dependencies - snakeyaml is in our pom already - or even better on the standard Java classes e.g. https://docs.oracle.com/javase/8/docs/api/java/util/Base64.Encoder.html

core/src/main/java/com/digitalpebble/stormcrawler/protocol/httpclient/HttpProtocol.java
```diff
@@ -112,6 +113,30 @@ public void configure(final Config conf) {
             defaultHeaders.add(new BasicHeader("Accept", accept));
         }
 
+        boolean useBasicAuth = ConfUtils.getBoolean(conf,
```
jnioche: Instead of having a separate configuration for activating or deactivating the authentication, what about checking if http.basicauth.user is not null? If it is set, we use it. That would keep things simpler.

pierrehts:
Thanks for your review @jnioche , I adjusted the code accordingly.

Also, I removed the WARN about missing password to allow the use of an empty password if necessary.

pierrehts:
Thanks for the review, I adjusted the code accordingly,
Also, I modified the logic to allow user to set an empty password if necessary.

jnioche:
Thanks @pierrehts. I have modified https://github.com/DigitalPebble/storm-crawler/wiki/Protocols to indicate that httpclient has a mechanism for basic authentication. I can give you write access to the WIKI if you want to update https://github.com/DigitalPebble/storm-crawler/wiki/Configuration.
Would be great to add the same mechanism to the okhttp implementation, would you be interested in having a look at that?
Thanks again

jnioche:
Hi @pierrehts, have sent you an invite to become a collaborator. You should be able to write to the WIKI after that.
Thanks for documenting the new properties. 
BTW if you are using StormCrawler for an organisation, why not add it to https://github.com/DigitalPebble/storm-crawler/wiki/Powered-By?

PR #592, opened 2018-07-12
Added ability to implement custom indexes names based on metadata information, fixes #591

### LINKED ISSUE #591: wish: Expose some fields on StatusUpdaterBolt to allow custom naming strategy implementation
Hi @jnioche , we are working with ES integration of the crawler and for some reasons we want to extend the functionality of **StatusUpdaterBolt** , basically we want to apply our custom naming strategy for handling many status indexes and content indexes names in parallel. Names will depend on same information stored previously in metadata. eg. `status_${some_property_name}`.
The existing code fits perfect to our needs, so we want to extend existing class and only override behavior that you have on `store` method.

Our main issue basically consist in the limited scope of all variables in **StatusUpdaterBolt** that we need to interact in `store` method of our child class that will be extending **StatusUpdaterBolt**.

Having those vars as protected or having getters for those will allow to extend this class and allow custom implementations in a future.

Obviously, we can implement our whole custom implementation extending AbstractStatusUpdaterBolt, but just want to check if you want to avoid overriding of this class for any special reason.

Thanks for your time in advance.

jcruzmartini:
Added `getIndexName(Metadata m)` method in bolts responsible for interacting with ES indexes. This allow implementing custom indexes names based on some metadata information.

### DIFF
--- a/external/elasticsearch/src/main/java/com/digitalpebble/stormcrawler/elasticsearch/bolt/DeletionBolt.java
+++ b/external/elasticsearch/src/main/java/com/digitalpebble/stormcrawler/elasticsearch/bolt/DeletionBolt.java
@@ -13,6 +13,7 @@ import org.elasticsearch.action.delete.DeleteRequest;
 import org.elasticsearch.client.RestHighLevelClient;
 import org.slf4j.LoggerFactory;
 
+import com.digitalpebble.stormcrawler.Metadata;
 import com.digitalpebble.stormcrawler.elasticsearch.ElasticSearchConnection;
 import com.digitalpebble.stormcrawler.util.ConfUtils;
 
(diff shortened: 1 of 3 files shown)


external/elasticsearch/src/main/java/com/digitalpebble/stormcrawler/elasticsearch/metrics/StatusMetricsBolt.java
```diff
@@ -102,6 +103,7 @@ public Object getValueAndReset() {
     public void execute(Tuple input) {
         _collector.ack(input);
 
+        Metadata metadata = (Metadata) input.getValueByField("metadata");
```
jcruzmartini: this wont work, because when it is a tick we do not have metadata information, and this bolt is triggering the search only when ticks. 
I will try to fix this asap

jnioche:
Thanks @jcruzmartini. Looks good, will you add the logic for Spouts as part of the same PR?

jcruzmartini:
@jnioche we realized that _AggregationSpout_ modifications is not needed at all, because we can use wildcards in index name config setting, replacing

` es.status.index.name: "status"`
by
` es.status.index.name: "status*"`

is enough, now we are searching against all our monthly indexes successfully .

jnioche:
@jcruzmartini let's leave it as it is for now, we can always add that later if needed by others. Thanks again for your contrib.
BTW if you are using SC in prod, why not let everybody know by joining the list on https://github.com/DigitalPebble/storm-crawler/wiki/Powered-By?

PR #594, opened 2018-07-20
Allow to connect all threads to a single domain

orliac:
So far in HttpProtocol.java there was an hard limit of 20 threads allowed per route:
`CONNECTION_MANAGER.setDefaultMaxPerRoute(20);`

This was creating problems while crawling on very limited number of domains and having set large numbers for configuration parameters `fetcher.threads.per.queue` and `fetcher.threads.number` (e.g. several hundreds of threads for both). Problem observed was of type:
```
org.apache.http.conn.ConnectionPoolTimeoutException: Timeout waiting for connection from pool
        at org.apache.http.impl.conn.PoolingHttpClientConnectionManager.leaseConnection(PoolingHttpClientConnectionManager.java:292) 
```
With the proposed solution we potentially allow all threads to connect to a single domain.

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/protocol/httpclient/HttpProtocol.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/protocol/httpclient/HttpProtocol.java
@@ -95,7 +95,8 @@ public class HttpProtocol extends AbstractHttpProtocol implements
                 200);
         CONNECTION_MANAGER.setMaxTotal(maxFetchThreads);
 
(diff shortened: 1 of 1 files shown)


jnioche:
Thanks @orliac, I'd rather it was set to the value of _fetcher.threads.per.queue_ if its value is > 20.
Setting it to the value of _maxFetchThreads_ is an overkill as in most cases crawlers behave politely and limit the number of threads to 1.

orliac:
Hi @jnioche, perfect, thank you!

PR #599, opened 2018-08-07
Added byteLength to Metadata

macheteaustria:
FetcherBolt already read the byteLength from the response, but didn't add it to Metadata.
(This could be helpful if you want to index the document size)


jnioche:
Thanks, could you please add the same to SimpleFetcherBolt as well? Bear in mind that this is the length fetched which may have been trimmed, not necessarily the real size of the document.

jnioche:
thanks @macheteaustria

PR #613, opened 2018-09-26
Batch PreparedStatements in SQL status updater bolt, fixes #610

### LINKED ISSUE #610: Batch PreparedStatements in SQL status updater bolt
Each document is currently updated in an individual call to SQL, which can be a bottleneck. It would be more efficient to batch the preparedstatements.

jnioche:
This PR changes the behaviour of the SQL StatusUpdaterBolt by batching the INSERTs, i.e. the discovered URLs which are far more frequent than updates.
 
@cruftex plz let me know what you think. we can make it SQL agnostic in a separate step.

### DIFF
--- a/external/sql/src/main/java/com/digitalpebble/stormcrawler/sql/Constants.java
+++ b/external/sql/src/main/java/com/digitalpebble/stormcrawler/sql/Constants.java
@@ -26,6 +26,7 @@ public class Constants {
     public static final String MYSQL_BUFFERSIZE_PARAM_NAME = "mysql.buffer.size";
     public static final String MYSQL_MAX_DOCS_BUCKET_PARAM_NAME = "mysql.max.urls.per.bucket";
     public static final String MYSQL_MIN_QUERY_INTERVAL_PARAM_NAME = "mysql.min.query.interval";
+    public static final String SQL_UPDATE_BATCH_SIZE_PARAM_NAME = "sql.update.batch.size";
 
     private Constants() {
     }
--- a/external/sql/src/main/java/com/digitalpebble/stormcrawler/sql/SQLUtil.java
+++ b/external/sql/src/main/java/com/digitalpebble/stormcrawler/sql/SQLUtil.java
@@ -21,6 +21,7 @@ import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.SQLException;
 import java.util.Map;
+import java.util.Properties;
 
 import com.digitalpebble.stormcrawler.util.ConfUtils;
 
@@ -30,7 +31,7 @@ public class SQLUtil {
     }
 
     @SuppressWarnings({ "rawtypes", "unchecked" })
-    static Connection getConnection(Map stormConf) throws SQLException {
+    public static Connection getConnection(Map stormConf) throws SQLException {
         // SQL connection details
         String url = ConfUtils.getString(stormConf,
                 Constants.MYSQL_URL_PARAM_NAME,
@@ -39,7 +40,12 @@ public class SQLUtil {
                 Constants.MYSQL_USER_PARAM_NAME);
         String password = ConfUtils.getString(stormConf,
                 Constants.MYSQL_PASSWORD_PARAM_NAME);
-        return DriverManager.getConnection(url, user, password);
+        Properties props = new Properties();
+        props.setProperty("user", user);
+        props.setProperty("password", password);
+        props.setProperty("rewriteBatchedStatements", "true");
+        props.setProperty("useBatchMultiSend", "true");
+        return DriverManager.getConnection(url, props);
     }
 
 }
--- a/external/sql/src/main/java/com/digitalpebble/stormcrawler/sql/StatusUpdaterBolt.java
+++ b/external/sql/src/main/java/com/digitalpebble/stormcrawler/sql/StatusUpdaterBolt.java
@@ -117,73 +117,84 @@ public class StatusUpdaterBolt extends AbstractStatusUpdaterBolt {
         ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
         long delay = 5000L;
         long period = 1000L;
-        executor.scheduleAtFixedRate(() -> sendInsertBatch(), delay, period, TimeUnit.MILLISECONDS);
+        executor.scheduleAtFixedRate(() -> {
+            try {
+                sendInsertBatch();
+            } catch (SQLException ex) {
+                LOG.error(ex.getMessage(), ex);
+                throw new RuntimeException(ex);
+            }
+        }, delay, period, TimeUnit.MILLISECONDS);
     }
 
     @Override
-    public void store(String url, Status status, Metadata metadata,
-            Date nextFetch) throws Exception {
+    public synchronized void store(String url, Status status,
+            Metadata metadata, Date nextFetch) throws Exception {
 
-        synchronized (this) {
-            StringBuilder mdAsString = new StringBuilder();
-            for (String mdKey : metadata.keySet()) {
-                String[] vals = metadata.getValues(mdKey);
-                for (String v : vals) {
-                    mdAsString.append("\t").append(mdKey).append("=").append(v);
-                }
-            }
+        // check whether the batch needs sending
+        sendInsertBatch();
 
-            int partition = 0;
-            String partitionKey = partitioner.getPartition(url, metadata);
-            if (maxNumBuckets > 1) {
-                // determine which shard to send to based on the host / domain /
-                // IP
-                partition = Math.abs(partitionKey.hashCode() % maxNumBuckets);
-            }
+        boolean isUpdate = !status.equals(Status.DISCOVERED);
 
-            PreparedStatement preparedStmt = this.insertPreparedStmt;
+        // aleady have an entry for this DISCOVERED URL
+        if (!isUpdate && waitingAck.containsKey(url)) {
+            return;
+        }
 
-            // create in table if does not already exist
-            if (!status.equals(Status.DISCOVERED)) {
-                preparedStmt = connection.prepareStatement(updateQuery);
-            }
-
-            preparedStmt.setString(1, url);
-            preparedStmt.setString(2, status.toString());
-            preparedStmt.setObject(3, nextFetch);
-            preparedStmt.setString(4, mdAsString.toString());
-            preparedStmt.setInt(5, partition);
-            preparedStmt.setString(6, partitionKey);
-
-            // updates are not batched
-            if (!status.equals(Status.DISCOVERED)) {
-                preparedStmt.executeUpdate();
-                preparedStmt.close();
-                eventCounter.scope("sql_updates_number").incrBy(1);
-                return;
-            }
-
-            // code below is for inserts i.e. DISCOVERED URLs
-            preparedStmt.addBatch();
-
-            // keep track of URL to ack the tuples later
-            waitingAck.put(url, new LinkedList<Tuple>());
-
-            // check the batch size
-            currentBatchSize++;
-
-            eventCounter.scope("sql_inserts_number").incrBy(1);
-
-            if (currentBatchSize == batchMaxSize) {
-                sendInsertBatch();
+        StringBuilder mdAsString = new StringBuilder();
+        for (String mdKey : metadata.keySet()) {
+            String[] vals = metadata.getValues(mdKey);
+            for (String v : vals) {
+                mdAsString.append("\t").append(mdKey).append("=").append(v);
             }
         }
+
+        int partition = 0;
+        String partitionKey = partitioner.getPartition(url, metadata);
+        if (maxNumBuckets > 1) {
+            // determine which shard to send to based on the host / domain /
+            // IP
+            partition = Math.abs(partitionKey.hashCode() % maxNumBuckets);
+        }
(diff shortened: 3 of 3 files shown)


cruftex:
Practical: Maybe if you assign me as reviewer I can annotate directly.

Findings in the code:

https://github.com/DigitalPebble/storm-crawler/blob/007a1701ab0019057d389b5b9bba1df85b3f82d4/external/sql/src/main/java/com/digitalpebble/stormcrawler/sql/StatusUpdaterBolt.java#L161
Don't close the prepared statement. It's actually a bug, nobody is creating it again for the update case.

https://github.com/DigitalPebble/storm-crawler/blob/007a1701ab0019057d389b5b9bba1df85b3f82d4/external/sql/src/main/java/com/digitalpebble/stormcrawler/sql/StatusUpdaterBolt.java#L210
You don't need to close and prepare again.

https://github.com/DigitalPebble/storm-crawler/blob/007a1701ab0019057d389b5b9bba1df85b3f82d4/external/sql/src/main/java/com/digitalpebble/stormcrawler/sql/StatusUpdaterBolt.java#L205
Don't use finally here, since it is confusing. The finally only makes sense if another exception happens then `SQLException`, however, in this case something is terribly wrong (e.g. OOM), there is no need to do a nice cleanup.

https://github.com/DigitalPebble/storm-crawler/blob/007a1701ab0019057d389b5b9bba1df85b3f82d4/external/sql/src/main/java/com/digitalpebble/stormcrawler/sql/StatusUpdaterBolt.java#L205
"empty catch block": In this case you should not have an empty catch block, since you might not see whether the prepared statement fails. However, you don't need to do that, see above.

https://github.com/DigitalPebble/storm-crawler/blob/007a1701ab0019057d389b5b9bba1df85b3f82d4/external/sql/src/main/java/com/digitalpebble/stormcrawler/sql/StatusUpdaterBolt.java#L120
Schedule is doing something every second. I would start the scheduler when the first entry comes in and stop it after the batch is send. Why? I have a golden rule: "If there is nothing to do, don't do anything!". 

https://github.com/DigitalPebble/storm-crawler/blob/007a1701ab0019057d389b5b9bba1df85b3f82d4/external/sql/src/main/java/com/digitalpebble/stormcrawler/sql/StatusUpdaterBolt.java#L158
Why are updates not batched? Have two different prepared statements.

https://github.com/DigitalPebble/storm-crawler/blob/007a1701ab0019057d389b5b9bba1df85b3f82d4/external/sql/src/main/java/com/digitalpebble/stormcrawler/sql/StatusUpdaterBolt.java#L159
https://github.com/DigitalPebble/storm-crawler/blob/007a1701ab0019057d389b5b9bba1df85b3f82d4/external/sql/src/main/java/com/digitalpebble/stormcrawler/sql/StatusUpdaterBolt.java#L147
There is a tendency in the code for quite longish methods. Dealing with the same condition more then one time in a method would be a good indicator for restructuring. Personally, I would restructure everything and make about 5 small methods out of it. If you stay with the long method, I would recommend to make the code more "speakable" by introducing a proper name for the condition: "boolean needsUpdate = !status.equals(Status.DISCOVERED);"

https://github.com/DigitalPebble/storm-crawler/blob/007a1701ab0019057d389b5b9bba1df85b3f82d4/external/sql/src/main/java/com/digitalpebble/stormcrawler/sql/StatusUpdaterBolt.java#L137
Maybe the partition thing should be removed here? With partitioning in this place you could send the data to different databases. If there is sharding in the database, the partitioning logic would be probably defined in the database. For cleanliness purposes I would remove everything that is not used.

https://github.com/DigitalPebble/storm-crawler/blob/007a1701ab0019057d389b5b9bba1df85b3f82d4/external/sql/src/main/java/com/digitalpebble/stormcrawler/sql/StatusUpdaterBolt.java#L233
Yep, in case of an update you need to ack right away. That's quite confusing here. It's an indicator that some restructuring could be good.

Maybe its worth to take a look UPSERT to solve the insert/update problem. Here is an overview which also contains the references how to do it in MySQL:
https://wiki.postgresql.org/wiki/UPSERT#UPSERT_as_implemented_in_practice

I think there needs to be more testing and error handling, since the code should be robust for things like double inserts.

cruftex:
>> Don't use finally here, since it is confusing. The finally only makes sense if another exception happens then SQLException, however, in this case something is terribly wrong (e.g. OOM), there is no need to do a nice cleanup.
> Unless I am mistaken, the finally is applied whether the SQLException was caught or not, which is a way of making sure that we reset things in either case.

"the finally is applied whether the SQLException was caught or not": That's correct, but also applies if you use don't use finally at all:

````
try {
catch {SQLException ex} {
// executed in case of exception thrown
}
// executed when SQLException is thrown in the try block or no exception is thrown
````

Finally will execute for a third case: any other exception. But in this case it means "hope is lost" and you don't need to take care of that, or, even might call cleanup and drop the connection.

> I read (but can't find where) that it is a good practice to close them to avoid memory leakage.

Resource-wise it would be better to drop the whole connection if nothing happens.

If you do execute only a single statement per prepared statement in case of an update, don't use the prepared statement at all. Using the prepared statement is counter productive and gives you additional database calls.

>> Why are updates not batched? Have two different prepared statements.
> Because that would mean duplicating a lot of fields and make the code more complex for a dubious gain, given that updates happen once per dozens of inserts.

It always depends on the use case. For a narrow target, frequent crawl you would have more updates then inserts.

> Given that there doesn't seem to be a universal way of doing it across various SQL servers, this is also not a good way of being 'SQL neutral'.

Totally agree on that!

>> Maybe the partition thing should be removed here? With partitioning in this place you could send the data to different databases. If there is sharding in the database, the partitioning logic would be probably defined in the database. For cleanliness purposes I would remove everything that is not used.
> As I said, we can refine things further but for now I wanted to keep the existing behaviour + having that soft partitioning is a cheap way of leveraging multiple spouts and can have a positive impact on performance.

That means that it is related to the spout but this code section duplicates in every status updater. That feels strange. Maybe that aspect can be moved elsewhere to avoid the duplication. Could that go in the super class? But okay, one step after another ;)

>> Yep, in case of an update you need to ack right away. That's quite confusing here. It's an indicator that some restructuring could be good.
> That would mean changing the way the super class deals with acking which is always a possibility but completely outside the scope of this particular issue.

Mind the cache. For batched inserts its not updated. Since a crawl discovers the same URLs many times, I expect a lot of duplicate key exceptions when you run that code. The super class is designed with a synchronous update scheme in mind, delayed and async updates really don't fit in well.

cruftex:
> Mind the cache. For batched inserts its not updated.

Oops. That's only partially true. It's updated but delayed, after the batch is run through.
However, it's very likely that you discover one URL more then once. If that happens in a short timeframe, that leads to a duplicate key exceptions, and in this case it would fail the whole batch. I think there may be content that makes this code fail continuously.

cruftex:
The elastic search status updater has a cache that keeps track of the operations in fly. I think the same is needed here. Its worth think about unifying this with the cache in the super class.

rzo1:
Very nice discussion going on here :)

I did not read the actual PR, but fromm
 your discussion I can get that this is very MySQL specific. If we would stick to plain SQL, wenn can easily provide unit tests with HSQLDB or similar.

Maybe it is also worth to look into connection pooling besides batching? :)

cruftex:
Hi @rzo1 

> Maybe it is also worth to look into connection pooling besides batching? :)

I thought about that as well. Connection pooling eliminates the cost of creating a connection in case you need it, plus, it would keep connection depending on the level of parallelism you actually experience.

If I understand it correctly, the way it works is, that you configure the level of parallelism you want in the configuration and storm creates as many status updates and distributes the load. Each status updater has its own connection, so there is no need for a connection pool, in terms of better performance. 

A connection pool might save resources by using a lower number of connections, but I think its okay to open all needed connections at the beginning. It even has benefits since you get errors right away if you hit limits.

rzo1:
I thought about automatic re-connection in case of timeout, stale/idle connections or others things like that - the things a DBA has to deal with it otherwise. 

It really has some benefits but can easily added later if the implementation does the right things :) - we even use something similar combined with the ES StormCrawler variant for some research purpose.

cruftex:
> I thought about automatic re-connection in case of timeout, stale/idle connections or others things like that - the things a DBA has to deal with it otherwise.

You are right. I did not think about that aspect at first. You might loose a connection because of a database restart or a network problem. A connection pool has the resilience already build in.

Another thing is that you want to use always more then one SQL connection to leverage parallelism in the database. Since needed always, doing this with multiple bolts is a bit inconvenient. So its useful that one SQL updater bolt uses more connections if needed.

jnioche:
> The elastic search status updater has a cache that keeps track of the operations in fly. I think the same is needed here.

https://github.com/DigitalPebble/storm-crawler/blob/007a1701ab0019057d389b5b9bba1df85b3f82d4/external/sql/src/main/java/com/digitalpebble/stormcrawler/sql/StatusUpdaterBolt.java#L76

>  Its worth thinking about unifying this with the cache in the super class.

or at least in a new super class, not all updater bolts need batching and not all of them are asynchronous

jnioche:
> Very nice discussion going on here :)

Thanks for joining in!

> 
> I did not read the actual PR, but fromm
> your discussion I can get that this is very MySQL specific. 
> If we would stick to plain SQL, we can easily provide unit tests with HSQLDB or similar.

There shouldn't be anything specific in this part of the code (or is there?)
See discussion in #610

jnioche:
> However, it's very likely that you discover one URL more then once. If that happens in a short timeframe, that leads to a duplicate key exceptions, and in this case it would fail the whole batch. I think there may be content that makes this code fail continuously.

@cruftex I assume this is an issue for the INSERTs only.  Would that be an issue even if the query contains IGNORE?

jnioche:
@cruftex 

> If you do execute only a single statement per prepared statement in case of an update, don't use the prepared statement at all.

OK, but is there a convenient way of escaping the values? IIRC normal statements are prone to exceptions if the content is not normalised

rzo1:
> There shouldn't be anything specific in this part of the code (or is there?)
> See discussion in #610

From the source I can see, that the constants are named with an MySQL prefix.
In addition, the driver properties are (at least for some values) specific to the MySQL driver.

Maybe we should configure the database driver via JSON (quite similar to other configuration in SC) and let the developer decide, which properties are needed for the specific environment, i.e. the jdbc driver.

In addition, we have some MySQL specific SQL in the codebase:

- "INSERT IGNORE" or "REPLACE INTO" does not exist for example in PostgreSQL - they have "UPSERT" since 9.5 - but this would also DB-vendor specific.

However, this could be done in a next step in improving the SQL module (besides the MySQL specific window functions in the SQLSPut).

While investigating the table creation script, I had the following question: Does the current code limit an URL to be only 255 characters long? I cannot remember exactly. But if not, this needs to be adapted.

> @cruftex I assume this is an issue for the INSERTs only. Would that be an issue even if the query contains IGNORE?

Yes. This is only relevant for INSERTs.

What about (1) keep a cache (2) check with a "SELECT EXISTS(SELECT url FROM table WHERE url = :url)" query in case we had no cache hit (3) insert the tuple. That would reduce the possibility of duplicate key exceptions but of course cannot rule them out totally.

> If you do execute only a single statement per prepared statement in case of an update, don't use the prepared statement at all.

At least for some JDBC drivers, the prepared statement should be preferred in terms o af local statement cache or just for escaping of values. I think, that the initial overhead of a prepared statement in this context is neglectible.

jnioche:
Have removed the finally section  + add check of cache to prevent duplicate INSERTs
There was also a bug as we were flushing the batch before the last entry added to it had been put in the cache, these entries could never be acked.
Will give it a try and will commit unless someone has an objection. We can then deal with improving it further in separate steps.

rzo1:
I agree, that generalizing away from MySQL ist a separate improvement, which should be tackled next in order to add unit tests.

cruftex:
Good morning everyone! I had a few days off, which clears the mind.

About the general SQL approach:
IMHO, the only standard way to deal with the "insert or update" problem, is actually to do a query first. If a result comes back, an update is needed, if not an insert is needed. The query and insert could be done within a transaction. But this means probably poor performance and poor parallelism. 

If one status updater is accessing the table exclusively, the transaction on the query and insert is not needed, since there are no other concurrent inserts.

In any case, I think the only standard way is to query first to see whether the URL is already known.

rzo1:
Exactly - I agree.

That's why I proposed:

What about (1) keep a cache (2) check with a "SELECT EXISTS(SELECT url FROM table WHERE url = :url)" query in case we had no cache hit (3) insert the tuple

jnioche:
> But this means probably poor performance and poor parallelism.

definitely not an option. 

> If one status updater is accessing the table exclusively, the transaction on the query and insert is not needed, since there are no other concurrent inserts.

there won't be concurrent inserts for the same entry but anyway there could be more than one statusupdater

Again, let's discuss making the code SQL neutral separately. But in any case, if it means having such bad performance that it can't be used then it's not going to happen.
One approach would be to have subclasses to deal with the different flavours of "insert or update", most of the code would be in the super class.

Any objections to committing these changes?

rzo1:
Let's do it and test it ;) - we will see really fast, which subclass approach performs better with less errors :)

jnioche:
thanks for your comments and suggestions @cruftex and @rzo1

PR #670, opened 2018-12-03
669: Implement outlink limit per page

jcruzmartini:
Added a new property called `parser.emitOutlinks.max.per.page` in config file, to be used for limiting quantity of outlinks discovered in a page during parsing.
If `parser.emitOutlinks.max.per.page : -1` limit will be disabled. Default value of this property is -1.  
Change is really straight forward .

Thanks 
Juan

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/bolt/JSoupParserBolt.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/bolt/JSoupParserBolt.java
@@ -29,6 +29,7 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang.StringUtils;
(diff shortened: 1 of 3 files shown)


jnioche:
thanks @jcruzmartini

core/src/main/java/com/digitalpebble/stormcrawler/bolt/JSoupParserBolt.java
```diff
@@ -339,7 +345,10 @@ public void execute(Tuple tuple) {
         }
 
         if (emitOutlinks) {
-            for (Outlink outlink : parse.getOutlinks()) {
+            final List<Outlink> outlinksAfterLimit = (maxOutlinksPerPage == -1) ? parse
```
scrubious: I think i understand what is happening here. When i first looked at this i read "after limit" as the outlinks that exceeded the limit, not as the outlinks we are going to process.

PR #680, opened 2019-01-21
Optional timeout configurations for elasticsearch's RestClient

ndtreviv:
In this version it is possible to optionally configure overrides for elasticsearch's default RestClient timeouts.

Options are:
```
es.[BOLT TYPE].connect.timeout: // Timeout for establishing a connection (if the value is 0, the timeout is considered "infinity")
es.[BOLT TYPE].socket.timeout: // Timeout for receiving data (if the value is 0, the timeout is considered "infinity")
es.[BOLT TYPE].max.retry.timeout: // The time before a connection retry is considered to have timed out (must be > 0)
```

All values should be in millis.

If no values are supplied, the elasticsearch defaults are used.


jnioche:
thanks @ndtreviv

PR #681, opened 2019-02-06
add: support for extracting frameset URLs.

AlbertWeichselbraun:
StormCrawler currently ignores framesets, since only `<a href="..." />` tags are considered as outlinks.

### Steps to reproduce the problem
A crawl of the site http://www.zub.ch/, for example, only yields one page.

### Suggested fix
The enclosed patch modifies `JSoupParserBolt` to also extract `<frame src="..." />` links.


jnioche:
This should already be doable using the LinkParseFilter, see [WIKI page on parse filters]( https://github.com/DigitalPebble/storm-crawler/wiki/ParseFilters). It probably makes sense to add support for frames in the default configuration in the archetype.

jnioche:
committed in b8d6f12242d12a8e516ef566c76e6f81c5c38174

thanks @AlbertWeichselbraun

Issue #688, opened 2019-03-01
Add a ScrollSpout to read all the documents from a shard


jnioche:
See discussion on #684 

This would be useful for reindexing an entire index e..g when the list of domains has changed or a larger number of shards is needed. These spouts would then be connected to a StatusUpdaterBolt in a simple topology

jnioche:
The logic of the StatusUpdaterBolt might need changing as well. Even if the spout sends all the info as-is, the status updater bolt might make changes to it e.g. increment the counter for number of failures. Either we split the existing behavior into 2 separate bolts or we have a brand new simplified bolt .

PR #690, opened 2019-03-07
Add a ScrollSpout to read all the documents from a shard

### LINKED ISSUE #684: StatusUpdaterBolt to use provided domain name for routing
StatusUpdaterBolt if configured with routing `byDomain` should use the routing key from metadata (if provided in the field defined by `es.status.routing.fieldname`). Updates of the public suffix list (included in the crawler-commons dependency) may change the domain name and routing key, and may cause duplicate status records in the index and needless refetches of the same URL (cf. commoncrawl/news-crawl#28).

The simplest solution is just to use the provided routing key (similar as it's done for routing `byIP`). This would require only changes in URLPartitioner. Alternatively, StatusUpdaterBolt could check whether the routing key has changed and then send a deletion request using the original routing key and update the status document with the new routing key.

jnioche:
Implements  #688 and fixes #684

This adds a ScrollSpout for ES as well as a mechanism to the AbstractStatusUpdaterBolt so that it stores a tuple without modifying any of its content. 

The following Flux illustrates its use

```
name: "reindexer"

includes:
    - resource: true
      file: "/crawler-default.yaml"
      override: false

    - resource: false
      file: "crawler-conf.yaml"
      override: true

    - resource: false
      file: "es-conf.yaml"
      override: true

config:
  es.status2.addresses: "localhost"
  es.status2.index.name: "status2"
  es.status2.doc.type: "status"
  es.status2.routing: true
  es.status2.routing.fieldname: "key"
  es.status2.bulkActions: 500
  es.status2.flushInterval: "1s"
  es.status2.concurrentRequests: 5
  es.status2.settings:
    cluster.name: "elasticsearch"
  topology.max.spout.pending: 5000
  topology.workers: 1

spouts:
  - id: "spout"
    className: "com.digitalpebble.stormcrawler.elasticsearch.persistence.ScrollSpout"
    parallelism: 10

bolts:
  - id: "status"
    className: "com.digitalpebble.stormcrawler.elasticsearch.persistence.StatusUpdaterBolt"
    parallelism: 4
    constructorArgs:
      - "status2"

streams:
  - from: "spout"
    to: "status"
    grouping:
      streamId: "status"
      type: CUSTOM
      customClass:
        className: "com.digitalpebble.stormcrawler.util.URLStreamGrouping"
        constructorArgs:
          - "byDomain"
```

The target index (here 'status2)' has to be initialised just like the source one. It can of course live on a separate cluster.  

It is a good idea to set `"refresh_interval": "-1"` in the configuration of the target index to speed up the writes. This can be set to any value afterwards when crawling with the new index.

Copying the content of a status index can be useful e.g. when changing the number of shards or the way the documents are assigned to them - for instance if using domains with a different version of crawler-commons, see  #684

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/persistence/AbstractQueryingSpout.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/persistence/AbstractQueryingSpout.java
@@ -92,7 +92,7 @@ public abstract class AbstractQueryingSpout extends BaseRichSpout {
     protected Queue<Values> buffer = new LinkedList<>();
     protected HashSet<String> in_buffer = new HashSet<>();
 
-    private SpoutOutputCollector _collector;
+    protected SpoutOutputCollector _collector;
 
     /** Required for implementations doing asynchronous calls **/
     protected AtomicBoolean isInQuery = new AtomicBoolean(false);
@@ -136,13 +136,12 @@ public abstract class AbstractQueryingSpout extends BaseRichSpout {
     protected abstract void populateBuffer();
 
     /**
-     * Map to keep in-process URLs, ev. with additional information for URL /
-     * politeness bucket (hostname / domain etc.). The entries are kept in a
-     * cache for a configurable amount of time to avoid that some items are
-     * fetched a second time if new items are queried shortly after they have
-     * been acked.
+     * Map to keep in-process URLs, with the URL as key and optional value
+     * depending on the spout implementation. The entries are kept in a cache
+     * for a configurable amount of time to avoid that some items are fetched a
+     * second time if new items are queried shortly after they have been acked.
      */
-    protected InProcessMap<String, String> beingProcessed;
+    protected InProcessMap<String, Object> beingProcessed;
     private boolean active;
 
     /** Map which holds elements some additional time after the removal. */
--- a/core/src/main/java/com/digitalpebble/stormcrawler/persistence/AbstractStatusUpdaterBolt.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/persistence/AbstractStatusUpdaterBolt.java
@@ -177,6 +177,7 @@ public abstract class AbstractStatusUpdaterBolt extends BaseRichBolt {
             try {
                 store(url, status, metadata, nextFetch);
                 ack(tuple, url);
+                return;
             } catch (Exception e) {
                 LOG.error("Exception caught when storing", e);
                 _collector.fail(tuple);
--- a/core/src/main/java/com/digitalpebble/stormcrawler/persistence/AbstractStatusUpdaterBolt.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/persistence/AbstractStatusUpdaterBolt.java
@@ -175,7 +175,7 @@ public abstract class AbstractStatusUpdaterBolt extends BaseRichBolt {
         if (dateInMetadata != null) {
             Date nextFetch = Date.from(Instant.parse(dateInMetadata));
             try {
-                store(url, status, metadata, nextFetch);
+                store(url, status, mdTransfer.filter(metadata), nextFetch);
                 ack(tuple, url);
                 return;
             } catch (Exception e) {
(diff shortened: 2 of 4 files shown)


sebastian-nagel:
Thanks, @jnioche! I'll test it on Monday with a real index.

jnioche:
hi @sebastian-nagel, how are you getting on?

sebastian-nagel:
Just launched the Docker container and created the new status index and trying to add some docs with outdated routing. I'll let you know about the results today afternoon.

sebastian-nagel:
Hi @jnioche, I've run a test reindexing an existing index with 13,000 docs using the given flux:
- resharding (from 10 to 16) works
- all docs have been re-indexed
- however, the routing is preserved from the original index (see #684). It's probably only a matter of the configuration, I'll have a look what happened. But maybe you have an idea?

jnioche:
Did you set 
'es.%s.routing : true'

where %s is the value used in the constructor of StatusUpdaterBolt?

what should happen is that no routing is done if this is not set but it should not preserve the routing. If it is set, it will compute the partition key based on the config of URLPartitioner

sebastian-nagel:
Yes, the new index is aliased `status2`, so the routing was active. Actually, a closer look at the records shows that I also have to change the routing field name (`metadata.hostname` instead of `key`):
```
          "metadata" : {
            "status%2Estore%2Eas%2Eis%2Ewith%2Enextfetchdate" : [
              "2019-03-13T10:05:30.000Z"
            ],
            "hostname" : [
              "www.sportmediaset.med"
            ]
          },
          "key" : "sportmediaset.med",
          "nextFetchDate" : "2019-03-13T10:05:30.000Z"
```
Also, how to get rid of the nextfetchdate in metadata?

sebastian-nagel:
Ok, with `es.status2.routing.fieldname: "metadata.hostname"` the topology fails with:
```
2019-03-13 12:10:55.947 c.d.s.e.p.StatusUpdaterBolt elasticsearch[scheduler][T#1] [ERROR] Exception with bulk 276 - failing the whole lot 
com.fasterxml.jackson.core.JsonParseException: Duplicate field 'hostname'
```

jnioche:
> Actually, a closer look at the records shows that I also have to change the routing field name (`metadata.hostname` instead of `key`):

you shouldn't have to change it, it depends on the schema you used and can be the same as in your source index

it is set with 'es.%s.routing.fieldname'

> 
> ```
>           "metadata" : {
>             "status%2Estore%2Eas%2Eis%2Ewith%2Enextfetchdate" : [
>               "2019-03-13T10:05:30.000Z"
>             ],
>             "hostname" : [
>               "www.sportmediaset.med"
>             ]
>           },
>           "key" : "sportmediaset.med",
>           "nextFetchDate" : "2019-03-13T10:05:30.000Z"
> ```
> Also, how to get rid of the nextfetchdate in metadata?

good catch, thanks! we should call 

'metadata = mdTransfer.filter(metadata)'

to filter the metadata to store as done during a crawl.

sebastian-nagel:
Filtering the metadata before storing solves both points:
```
@@ -175,7 +175,7 @@ public abstract class AbstractStatusUpdaterBolt extends BaseRichBolt {
         if (dateInMetadata != null) {
             Date nextFetch = Date.from(Instant.parse(dateInMetadata));
             try {
-                store(url, status, metadata, nextFetch);
+                store(url, status, mdTransfer.filter(metadata), nextFetch);
                 ack(tuple, url);
                 return;
```
The metadata now contains the fixed hostname/routing value and no "status%2Estore%2Eas%2Eis%2Ewith%2Enextfetchdate":
```
          "metadata" : {
            "hostname" : "sportmediaset.med"
          },
          "nextFetchDate" : "2019-03-13T11:59:20.000Z"
```

sebastian-nagel:
Regarding the `Duplicate field 'hostname'` error: since "hostname" is not in `metadata.persist` it is filtered away and there is no collision.

jnioche:
the exception was due to the field not being filtered out and added forcibly later on.
I'll add the filtering

jnioche:
thanks Sebastian. OK to merge?

sebastian-nagel:
Yes, lgtm. Resharding and re-routing work. With the metadata filter applied it would even allow to clean up the metadata.

jnioche:
thanks for the review and tests @sebastian-nagel

Issue #509, opened 2017-10-18
Generate WARC request records


jnioche:
#508 allows to get the verbatim representation of the request, the WARCRecordFormatClass could therefore generate WARC request records

jnioche:
@sebastian-nagel this would be relevant for the news-crawl

PR #691, opened 2019-03-13
WARC format improvements

sebastian-nagel:
This PR combines a list of improvements to the WARC writer bolt and record formats but also to the okhttp-based protocol implementation:
- generate WARC request records (#509) by calling `withRequestRecords()` on an existing WARCHdfsBolt.
- use capture time (when fetching began) for WARC-Date not the time the WARC record is written. This also ensures that response and request record have the same `WARC-Date`
- add field `WARC-IP-Address (address commoncrawl/news-crawl#29)
- ports back a couple of improvements from Nutch's protocol-okhttp:
  - use okhttp's internal byte buffer, do not buffer content twice
  - optionally keep partially fetched content (e.g., if a single read fails with a socket timeout) if `http.content.partial.as.trimmed` is true
- track reason for trimmed content (content or time limit, read or connection failure) and add it as `WARC-Truncated` field (address commoncrawl/news-crawl#31)
-  mask HTTP headers (Content-Length, Content-Encoding and Transfer-Encoding) which may cause errors when processing WARC records because the content is stored uncompressed and unchunked and may be of different length than originally received. The prefix to mask the original headers is `X-Crawler`. Address commoncrawl/news-crawl#30.

I can split the PR into multiple one if required.

The result WARC files have been checked and validated using warcio's new check utility (webrecorder/warcio#68). An example request/response record pair with all changes applied:
```
WARC/1.0
WARC-Type: request
WARC-Record-ID: <urn:uuid:8dfc0435-7417-4b4c-8189-705d766a3b16>
Content-Length: 377
WARC-Date: 2019-03-13T13:56:00.953Z
WARC-Target-URI: https://www.theguardian.com/politics/2019/mar/13/theresa-may-confirms-she-will-vote-to-block-no-deal-brexit-pmqs-jeremy-corbyn
Content-Type: application/http; msgtype=request
WARC-Block-Digest: sha1:KVPT2QYXP5FMSI542GRPAG4VXOUVCNRC

GET /politics/2019/mar/13/theresa-may-confirms-she-will-vote-to-block-no-deal-brexit-pmqs-jeremy-corbyn 
User-Agent: CCBot/3.0 (http://commoncrawl.org/faq/; info@commoncrawl.org)
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
Accept-Language: en-us,en-gb,en;q=0.7,*;q=0.3
Host: www.theguardian.com
Connection: Keep-Alive
Accept-Encoding: gzip



WARC/1.0
WARC-Record-ID: <urn:uuid:203780a2-dcb8-49a7-973e-1d88b125f587>
WARC-Concurrent-To: <urn:uuid:8dfc0435-7417-4b4c-8189-705d766a3b16>
Content-Length: 1041736
WARC-Date: 2019-03-13T13:56:00.953Z
WARC-Type: response
WARC-IP-Address: 151.101.113.111
WARC-Target-URI: https://www.theguardian.com/politics/2019/mar/13/theresa-may-confirms-she-will-vote-to-block-no-deal-brexit-pmqs-jeremy-corbyn
WARC-Payload-Digest: sha1:YWXUX773QVKHQPO4MQXAFFBOH7IIYIA4
WARC-Block-Digest: sha1:DY6IIMGARYQGPERMPEUT6RUB3VZRQF76

HTTP/1.1 200 OK
X-Crawler-Content-Encoding: gzip
Content-Type: text/html; charset=utf-8
ETag: W/"hash-826299181980705017"
Link: <https://assets.guim.co.uk/stylesheets/ed6dfcab2e0ddab1c51867a93921f6ca/content.garnett.css>; rel=preload; as=style; nopush,<https://assets.guim.co.uk/polyfill.io/v3/polyfill.min.js?rum=0&features=es6,es7,es2017,default-3.6,HTMLPictureElement,IntersectionObserver,IntersectionObserverEntry&flags=gated&callback=guardianPolyfilled&unknown=polyfill>; rel=preload; as=script; nopush,<https://assets.guim.co.uk/javascripts/ff1648a1c5e6beec11e0/graun.standard.js>; rel=preload; as=script; nopush,<https://assets.guim.co.uk/javascripts/9a2a60d293fd119118cf/graun.commercial.js>; rel=preload; as=script; nopush
X-Crawler-Content-Length: 164346
Content-Length: 1040002
Accept-Ranges: bytes
Date: Wed, 13 Mar 2019 13:56:00 GMT
Age: 36
Connection: keep-alive
Set-Cookie: GU_mvt_id=186384; expires=Tue, 11 Jun 2019 13:56:00 GMT; path=/; domain=.theguardian.com; Secure
X-Timer: S1552485361.961381,VS0,VE0
Strict-Transport-Security: max-age=31536000; includeSubDomains; preload
X-Frame-Options: SAMEORIGIN
X-XSS-Protection: 1; mode=block
X-Content-Type-Options: nosniff
Content-Security-Policy: default-src https:; script-src https: 'unsafe-inline' 'unsafe-eval'; style-src https: 'unsafe-inline'; img-src https: data: blob:; media-src https: data: blob:; font-src https: data:; connect-src https: wss:
Referrer-Policy: no-referrer-when-downgrade
Feature-Policy: camera 'none'; microphone 'none'; midi 'none'
X-GU-Edition: int
Cache-Control: max-age=60, stale-while-revalidate=6, stale-if-error=864000, private
Vary: Accept-Encoding,User-Agent
Set-Cookie: GU_geo_continent=EU; path=/; Secure


<!DOCTYPE html>
<html id="js-context" class="js-off is-not-modern id--signed-out" lang="en" data-page-path="/politics/2019/mar/13/theresa-may-confirms-she-will-vote-to-block-no-deal-brexit-pmqs-jeremy-corbyn">
<head>
<!--
     __        __                      _     _      _
     \ \      / /__    __ _ _ __ ___  | |__ (_)_ __(_)_ __   __ _
      \ \ /\ / / _ \  / _` | '__/ _ \ | '_ \| | '__| | '_ \ / _` |
       \ V  V /  __/ | (_| | | |  __/ | | | | | |  | | | | | (_| |
        \_/\_/ \___|  \__,_|_|  \___| |_| |_|_|_|  |_|_| |_|\__, |
                                                            |___/
    Ever thought about joining us?
    https://workforus.theguardian.com/careers/digital-development/
     --->
<title>Theresa May confirms she will vote to block no-deal Brexit | Politics | The Guardian</title>
<meta charset="utf-8">
```

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
@@ -23,11 +23,8 @@ import java.net.MalformedURLException;
 import java.net.Proxy;
 import java.net.URL;
 import java.security.cert.CertificateException;
-import java.time.Instant;
 import java.time.ZoneId;
 import java.time.ZoneOffset;
-import java.time.ZonedDateTime;
-import java.time.format.DateTimeFormatter;
 import java.util.Base64;
 import java.util.LinkedList;
 import java.util.List;
@@ -84,11 +81,6 @@ public class HttpProtocol extends AbstractHttpProtocol {
 
     /** Accept partially fetched content as trimmed content */
     private boolean partialContentAsTrimmed = false;
-
-    private final static String VERBATIM_REQUEST_KEY = "_request.headers_";
-    private final static String VERBATIM_RESPONSE_KEY = "_response.headers_";
-    private final static String VERBATIM_RESPONSE_IP_KEY = "_response.ip_";
-    private final static String VERBATIM_REQUEST_TIME_KEY = "_request.time_";
 
     private final List<String[]> customRequestHeaders = new LinkedList<>();
 
@@ -261,11 +253,12 @@ public class HttpProtocol extends AbstractHttpProtocol {
                 String key = headers.name(i);
                 String value = headers.value(i);
 
-                if (key.equalsIgnoreCase(VERBATIM_REQUEST_KEY) || key.equalsIgnoreCase(VERBATIM_RESPONSE_KEY)) {
+                if (key.equals(ProtocolResponse.REQUEST_HEADERS_KEY)
+                        || key.equals(ProtocolResponse.RESPONSE_HEADERS_KEY)) {
                     value = new String(Base64.getDecoder().decode(value));
                 }
 
-                responsemetadata.addValue(key.toLowerCase(Locale.ROOT), value);
+                responsemetadata.addValue(key, value);
             }
 
             MutableObject trimmed = new MutableObject(TrimmedContentReason.NOT_TRIMMED);
@@ -274,8 +267,10 @@ public class HttpProtocol extends AbstractHttpProtocol {
                 if (!call.isCanceled()) {
                     call.cancel();
                 }
-                responsemetadata.setValue("http.trimmed", "true");
-                responsemetadata.setValue("http.trimmed.reason",
+                responsemetadata.setValue(ProtocolResponse.TRIMMED_RESPONSE_KEY,
+                        "true");
+                responsemetadata.setValue(
+                        ProtocolResponse.TRIMMED_RESPONSE_REASON_KEY,
                         trimmed.getValue().toString().toLowerCase(Locale.ROOT));
                 LOG.warn("HTTP content trimmed to {}", bytes.length);
             }
@@ -355,9 +350,7 @@ public class HttpProtocol extends AbstractHttpProtocol {
         @Override
         public Response intercept(Interceptor.Chain chain) throws IOException {
 
-            String startFetchTime = DateTimeFormatter.ISO_INSTANT
-                    .format(ZonedDateTime.ofInstant(Instant.now(),
-                            TIME_ZONE_UTC));
+            long startFetchTime = System.currentTimeMillis();
 
             Connection connection = chain.connection();
             String ipAddress = connection.socket().getInetAddress()
@@ -415,12 +408,13 @@ public class HttpProtocol extends AbstractHttpProtocol {
             // returns a modified version of the response
             return response
                     .newBuilder()
-                    .header(VERBATIM_REQUEST_KEY,
+                    .header(ProtocolResponse.REQUEST_HEADERS_KEY,
                             new String(encodedBytesRequest))
-                    .header(VERBATIM_RESPONSE_KEY,
+                    .header(ProtocolResponse.RESPONSE_HEADERS_KEY,
                             new String(encodedBytesResponse))
-                    .header(VERBATIM_RESPONSE_IP_KEY, ipAddress)
-                    .header(VERBATIM_REQUEST_TIME_KEY, startFetchTime).build();
+                    .header(ProtocolResponse.RESPONSE_IP_KEY, ipAddress)
+                    .header(ProtocolResponse.REQUEST_TIME_KEY,
+                            Long.toString(startFetchTime)).build();
         }
     }
 
--- a/core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
@@ -23,8 +23,6 @@ import java.net.MalformedURLException;
 import java.net.Proxy;
 import java.net.URL;
 import java.security.cert.CertificateException;
-import java.time.ZoneId;
-import java.time.ZoneOffset;
 import java.util.Base64;
 import java.util.LinkedList;
 import java.util.List;
--- a/core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
@@ -18,7 +18,6 @@
 package com.digitalpebble.stormcrawler.protocol.okhttp;
 
 import java.io.IOException;
-import java.io.InputStream;
 import java.net.InetSocketAddress;
 import java.net.MalformedURLException;
 import java.net.Proxy;
@@ -38,9 +37,8 @@ import javax.net.ssl.TrustManager;
 import javax.net.ssl.X509TrustManager;
 
 import org.apache.commons.lang.StringUtils;
-import org.apache.commons.lang.mutable.MutableBoolean;
+import org.apache.commons.lang.mutable.MutableObject;
 import org.apache.http.cookie.Cookie;
-import org.apache.http.util.ByteArrayBuffer;
 import org.apache.storm.Config;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
```diff
@@ -270,80 +285,105 @@ public ProtocolResponse getProtocolOutput(String url, final Metadata metadata) t
...
     }
 
     class HTTPHeadersInterceptor implements Interceptor {
 
+        final ZoneId TIME_ZONE_UTC = ZoneId.of(ZoneOffset.UTC.toString());
+
         @Override
         public Response intercept(Interceptor.Chain chain) throws IOException {
+
+            String startFetchTime = DateTimeFormatter.ISO_INSTANT
```
jnioche: isn't that the same as Instant.now().toString()? the formatter used is
https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_INSTANT

external/warc/src/main/java/com/digitalpebble/stormcrawler/warc/WARCRecordFormat.java
```diff
@@ -149,9 +268,13 @@ public static String getDigestSha1(byte[] bytes1, byte[] bytes2) {
         buffer.append("Content-Length").append(": ")
                 .append(Integer.toString(contentLength)).append(CRLF);
 
-        // TODO get actual fetch time from metadata if any
-        Date now = new Date();
-        buffer.append("WARC-Date").append(": ").append(WARC_DF.format(now))
+        // get actual fetch time from metadata if any
+        String captureTime = metadata.getFirstValue("_request.time_");
+        if (captureTime == null) {
+            Date now = new Date();
```
jnioche: use Instant instead of Date? Would mean changing in the rest of this class

external/warc/src/main/java/com/digitalpebble/stormcrawler/warc/WARCRecordFormat.java
```diff
@@ -166,9 +289,9 @@ public static String getDigestSha1(byte[] bytes1, byte[] bytes2) {
                 .append(CRLF);
 
         // "WARC-IP-Address" if present
-        String IP = metadata.getFirstValue("_ip_");
+        String IP = metadata.getFirstValue("_response.ip_");
         if (StringUtils.isNotBlank(IP)) {
-            buffer.append("WARC-IP-Address").append(": ").append("IP")
```
jnioche: looks like this was a bug

core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
sebastian-nagel: Looks like, although I don't like the toString() methods which often implicitly assumes some localization. It's also not the best design to do the date formatting defined by the WARC format in the protocol layer. But re-parsing the date from a String is also not ideal.

external/warc/src/main/java/com/digitalpebble/stormcrawler/warc/WARCRecordFormat.java
sebastian-nagel: Good idea. It would actually only change these lines.

external/warc/src/main/java/com/digitalpebble/stormcrawler/warc/WARCRecordFormat.java
sebastian-nagel: yep.

core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
jnioche: the protocol could store the value as a long and let the WARC code format it whichever way it wants e.g. using _Instant.ofEpochSecond(long epochSecond)_

jnioche:
@sebastian-nagel could you update the README for the WARC module e.g. show how to set withRequestRecords() in Java topology and Flux?

sebastian-nagel:
Yes, I'll do and also address the other comments regarding the fetch date. After a second check also this should be addressed:
- add IP address also to request records
- the WARC-Date should be only with a precision in seconds to be compatible with [v1.0 of the WARC standard](http://iipc.github.io/warc-specifications/specifications/warc-format/warc-1.0/#warc-date-mandatory) ([1.1](http://iipc.github.io/warc-specifications/specifications/warc-format/warc-1.1/#warc-date-mandatory) allows a higher precision)

sebastian-nagel:
Rebased PR and addressed all discussed points. Also tested the WARC bolt via Flux, including passing the warcinfo map and configuring the bolt to use the RawLocalFileSystem.

core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
```diff
@@ -270,80 +280,103 @@ public ProtocolResponse getProtocolOutput(String url, final Metadata metadata) t
...
+            bytesToCopy = maxContentBytes;
+        }
+        byte[] arr = new byte[bytesToCopy];
+        source.buffer().readFully(arr);
+        return arr;
     }
 
     class HTTPHeadersInterceptor implements Interceptor {
 
+        final ZoneId TIME_ZONE_UTC = ZoneId.of(ZoneOffset.UTC.toString());
```
jnioche: still needed?

sebastian-nagel: Yep, it's removed.

jnioche: thanks - could you remove the imports as well?

sebastian-nagel: Done. Thanks!

jnioche:
@sebastian-nagel could we have _http.content.partial.as.trimmed_ as well as any other new config in the default config file with some comments to document it? The default config file acts as documentation in a way.

sebastian-nagel:
Done. Btw, crawler-default.yaml mentions `http.store.responsetime` which seems unused (not implemented).

jnioche:
thanks @sebastian-nagel that was part of the initial protocol from Nutch, will remove it

jnioche:
thanks for this contrib @sebastian-nagel

core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
```diff
@@ -248,20 +251,25 @@ public ProtocolResponse getProtocolOutput(String url, final Metadata metadata) t
                 String key = headers.name(i);
                 String value = headers.value(i);
 
-                if (key.equalsIgnoreCase(VERBATIM_REQUEST_KEY) || key.equalsIgnoreCase(VERBATIM_RESPONSE_KEY)) {
+                if (key.equals(ProtocolResponse.REQUEST_HEADERS_KEY)
+                        || key.equals(ProtocolResponse.RESPONSE_HEADERS_KEY)) {
                     value = new String(Base64.getDecoder().decode(value));
                 }
 
-                responsemetadata.addValue(key.toLowerCase(Locale.ROOT), value);
```
jnioche: @sebastian-nagel not normalising the keys to lowercase has serious implications on the  subsequent steps and is also incompatible with what the other implementation produces. What problem were you trying to solve here?

sebastian-nagel: Looks like a mistake - given that this was together with moving custom headers to constants (being lowercase).

sebastian-nagel: After a closer look: that's a regression. I see: HTTP header names are required in lowercase everywhere else. Sorry.

jnioche: no problems, I should have spotted it when merging. Thanks for confirming!

sebastian-nagel: Yeah, it would break some other things, eg. last-modified handling. Maybe add a comment?
Btw., the `source.buffer()` calls are deprecated, should be `getBuffer()` now.

PR #716, opened 2019-05-03
ESSeedInjector: filter bolt to subscribe to spout's status stream, fixes #715

### LINKED ISSUE #715: ESSeedInjector: no URLs injected because URL filter does not subscribe to status stream
With #677 the FileSpout emits tuples to the "status" stream if called with param `withDiscoveredStatus`. However, the URL filter bolt does not subscribe to the "status" stream which causes that all injected URLs are lost. I'll open a PR soon.

sebastian-nagel:
- ESSeedInjector: explicitly set name of stream for source of URL filter bolt
- more verbose logging in FileSpout and URLFilterBolt - for easier debugging to exclude trivial reasons such as URLs have been filtered due to misconfigured filters
- add optional URL filter bolt config to es-injector.flux


jnioche:
thanks @sebastian-nagel

PR #732, opened 2019-05-24
Force English Locale for SimpleDateFormat in coockie converter.

mstrewe:
The coockie parser test will fail on german pc bacause the SimpleDateFormatter will use german day names at the expire check in CookieConverter


mstrewe:
I checked all other SimpleDateFormat usages. Because there is no use of words, but only numbers, the locale must not be given. On 2 other occurences Locale is set

jnioche:
thanks @mstrewe

PR #760, opened 2019-10-02
Fix NullPointerException in JSONResourceWrappers

JacobsonMT:
`JSONURLFilterWrapper` and `JSONResourceWrapper` will always throw a `NullPointerException` during their `Timer`'s run method as `JSONResource resource` is never initialized.

Also cleans up invalid calls to `LOG.error`.


jnioche:
Good catch @JacobsonMT, thanks!

PR #774, opened 2019-12-12
(Storm 2.x) Fix call of storm jar / storm local in README

sebastian-nagel:
The Storm 2.x command-line parser has been rewritten. Without any additional hint it also consumes the options and flags for the crawl topology class. The `--` tells storm that the remaining options and flags should be passed to the topology main class. Storm options need be given before.

Without the `--` when launching the topology via `storm jar` or `storm local` the `-conf` options are stripped and the Java command is logged:
```
Running: java ... package.CrawlTopology /path/es-conf.yaml /path/crawler-conf.yaml
```
Consequently, the crawl topology does not read the configuration files. By calling the topology as
```
storm {jar,local} [<storm-opts>] -- package.CrawlTopology [<crawl-topology-opts>]
```
the Java command correct:
```
Running: java ... package.CrawlTopology -conf /path/es-conf.yaml -conf /path/crawler-conf.yaml
```

Btw., successfully run a test crawl on Storm 2.1.0 and ES 7.5.0

### DIFF
--- a/archetype/src/main/resources/archetype-resources/README.md
+++ b/archetype/src/main/resources/archetype-resources/README.md
@@ -10,7 +10,7 @@ mvn clean package
 before submitting the topology using the storm command:
 
 ``` sh
-storm local target/${artifactId}-${version}.jar ${package}.CrawlTopology -conf crawler-conf.yaml --local-ttl 60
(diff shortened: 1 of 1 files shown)


jnioche:
thanks @sebastian-nagel; great to hear that you got it to run on 2.x
am hoping to do an initial release of the 2.0 branch soon. All the development effort of Storm has gone into 2.x so we might as well go with it.

PR #775, opened 2019-12-13
okhttp protocol: HTTP request header lacks protocol name and version

sebastian-nagel:
The verbatim HTTP request header saved in metadata by the okhttp protocol implementation does not contain the protocol name and version in the request message. The request message should be (instead of `GET /path `):
```
GET /path HTTP/1.1
```
See also commoncrawl/news-crawl#34.

This PR adds
- the HTTP protocol name and version to the request message
- normalizes the protocol name for HTTP/2 (okhttp's internal name is "H2")

Two notes:
- HTTP/2 is enabled by default but requires Java 9 upwards, e.g. when running with OpenJDK 11 request/response headers are now:
```
$> /usr/lib/jvm/java-11-openjdk-amd64/bin/java ... com.digitalpebble.stormcrawler.protocol.okhttp.HttpProtocol \
   -c ... https://www.google.com/
...
_request.headers_: GET / HTTP/2
...

_response.headers_: HTTP/2 200 
...
```
- request and response protocol version may differ. Verified using wget:
```
$> wget -d https://www.kp.ru/daily/27061/4129507/
...
---request begin---
GET /daily/27061/4129507/ HTTP/1.1
...
---response begin---
HTTP/1.0 200 OK
...
```

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
@@ -57,6 +57,7 @@ import okhttp3.Headers;
 import okhttp3.Interceptor;
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
+import okhttp3.Protocol;
 import okhttp3.Request;
 import okhttp3.Request.Builder;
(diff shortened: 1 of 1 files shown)


jnioche:
fab, thanks @sebastian-nagel

Issue #777, opened 2019-12-13
status index / Adaptive scheduler: get rid of last-modified dates in non-ISO date format


sebastian-nagel:
There are still quite a few "last-modified" dates not in ISO date format, e.g.:
``` json
"_source" : {
  "url" : "https://zz.te.ua/feed/",
  "status" : "FETCHED",
  "metadata" : {
    "last-modified" : [
      "Thu, 12 Dec 2019 20:00:57 GMT"
    ],
    "signature" : [
      "9fde2d107d3339060a178fa8a59a55da"
    ],
    "fetchInterval" : [
      "432"
    ],
    "isFeed" : [
      "true"
    ],
    "fetch%2EstatusCode" : [
      "200"
    ],
    "signatureChangeDate" : [
      "2019-12-12T21:44:57.948Z"
    ],
    "hostname" : "zz.te.ua"
  },
  "nextFetchDate" : "2019-12-13T13:02:41.000Z"
}
```

Protocol implementations expect the last-modified time to be in ISO format (see #673) - if the ISO format is not followed, they do not send the "If-not-modified-since" header.

With #496 the adaptive scheduler uses the ISO date-time format to store the last-modified time.
However, the adaptive scheduler only sets the last-modified time when the comparison of the current and the previous signature indicates a content change. It never sets the last-modified time
1. for the initial fetch of a page (no signatures to compare) or
2. if the pages has not changed (HTTP 304 or same signature)
3. if another scheduler is used

In these situations, if the server sends a "Last-modified" header with the response, this value overwrites the existing one. There two issues with the value from the "Last-modified" header:
- the date format
- we cannot trust the date:
  - it could be a time in the future or even implausible far in the past
  - or just off few hours because of time zone issues or just the clock going wrong

"If-modified-since" requests should always use the date of the last fetch (if successful).

To overcome the issue:
1. use a different metadata field than "last-modified" for either AdaptiveScheduler or to hold the response metadata (see #776)
2. let AdaptiveScheduler set "last-modified" for initial fetches (or first successful fetch after a fetch error or redirect status)

sebastian-nagel:
Alternative solution without changes of metadata field names:
-  AdaptiveScheduler
   - sets "signatureChangeDate" also at the initial fetch
   - always overwrites "last-modified" with "signatureChangeDate" so that the "last-modified" field holds a well-formatted date or is empty after non-successful fetches
[...]

PR #789, opened 2020-02-20
Prefix protocol metadata 

### LINKED ISSUE #776: Prefix protocol metadata to avoid that internal metadata fields are accidentally overwritten
HTTP protocol implementations add all HTTP response header fields to the response metadata. The FetcherBolt merges the response metadata into the metadata which is passed forward in the topology as part of the tuples <url, content, metadata>. Existing key-value(s) pairs (persisted in the status index) are overwritten and later eventually stored in the status index (if listed in "metadata.persist") or even transferred to outlinks ("metadata.transfer").

This allows to easily use the response header values for requests, e.g. cookies or "ETag".

However, webadmins are free to send any header back. This may cause unwanted collisions with metadata keys used by crawler-internal classes. E.g. if the server responds with non-standard headers "HostName" or "Depth". Or even standard headers such as "Last-modified" which require to follow a specific format for internal use.

To avoid collisions: Why not prefix protocol metadata: `protocol.content-type` or `http.content-type`? This would also make it clear which component sets the metadata - similar to the prefixes `fetch.` and `parse.` already used. The draw-back would be that users are required to update the configuration.

jnioche:
fixes #776
@sebastian-nagel is that what you had in mind?

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/Metadata.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/Metadata.java
@@ -67,6 +67,37 @@ public class Metadata {
         md.putAll(m.md);
     }
 
+    /**
+     * Puts all prefixed metadata into the current instance
+     * 
+     * @param m
+     *            metadata to be added
+     * @param prefix
+     *            string to prefix keys in m before adding them to the current
+     *            metadata. No separator is inserted between prefix and original
+     *            key, so the prefix must include any separator (eg. a dot)
+     * @param cleanup
+     *            if true and prefix is not empty, the map is cleaned up by
+     *            removing any keys starting with prefix except those
+     *            overwritten from m
+     * 
+     */
+    public void putAll(Metadata m, String prefix, boolean cleanup) {
+        if (prefix == null || prefix.isEmpty()) {
+            md.putAll(m.md);
+            return;
+        }
+        checkLockException();
+
+        Map<String, String[]> ma = m.asMap();
+        if (cleanup) {
+            md.keySet().removeIf(k -> (k.startsWith(prefix) && !ma.containsKey(k)));
+        }
+        ma.forEach((k, v) -> {
+            setValues(prefix + k, v);
+        });
+    }
+
     /** @return the first value for the key or null if it does not exist **/
     public String getFirstValue(String key) {
         String[] values = md.get(key);
--- a/core/src/main/java/com/digitalpebble/stormcrawler/Metadata.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/Metadata.java
@@ -76,24 +76,14 @@ public class Metadata {
      *            string to prefix keys in m before adding them to the current
      *            metadata. No separator is inserted between prefix and original
      *            key, so the prefix must include any separator (eg. a dot)
-     * @param cleanup
-     *            if true and prefix is not empty, the map is cleaned up by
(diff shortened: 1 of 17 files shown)


sebastian-nagel:
Looks good. What about the ETag header? Should also get prefixed by default. See 582ead1 ?

sebastian-nagel:
One remark (haven't verified whether this is a problem or not): what happens if there is already some metadata stored in the status index from a prior fetch (eg. "protocol.etag" or just "etag") and now a fetch is made without the "ETag" header in the response? Is the prior value left unchanged? Should it be cleared/removed? - The prefix would allow to safely clear old protocol metadata.

sebastian-nagel:
Handling of cookies also requires changes regarding the prefixed metadata. Here cleaning up the metadata would break handling of cookies: the set-cookie value must be preserved unless a new one is sent by the server. I'll update branch 776 regarding ETag and cookies.

sebastian-nagel:
Also, the WARC writer requires changes as it reads the HTTP headers stored in response metadata.

sebastian-nagel:
Also the parsers required changes to read HTTP Content-Type and status whether content is trimmed.

jnioche:
thanks @sebastian-nagel , will have a look tomorrow

jnioche:
instead of changing all the key names (which makes the code a bit more difficult to read) - what about adding _getFirstValue(key, prefix)_ to metadata?

jnioche:
as a general observation, this is quite a few changes for an issue which is largely hypothetical, but let's see if we can get the right balance

sebastian-nagel:
> getFirstValue(key, prefix) 

To retrieve the cookies also getValues(key, prefix) is required.

I totally agree about the amount of changes. I also didn't expect that many. However, it isn't hypothetical regarding the last-modified date format which is still an open issue (#777) - all because the field is (also) filled from an external server and there are some limitations on the date format (see #492).

jnioche:
@sebastian-nagel have added the 2 methods to Metadata. 
Need to get my head around  **_putAll(Metadata m, String prefix, boolean cleanup)_**

sebastian-nagel:
> use getValue methods with prefix

+1 looks good

jnioche:
@sebastian-nagel do we really need  _public void putAll(Metadata m, String prefix, boolean cleanup)?

we're only using it iin 2 places and with _false_.

I'll add a default value for protocol.md.prefix in the default config of 'protocol.'; anyone upgrading on an existing crawl will have to override it to an empty string - I'll make this clear in the release notes. Leaving it to an empty string by default means no one will set it even for new crawls. What do you think?

sebastian-nagel:
Not necessarily. The the discussion above why I've added the cleanup option and why it cannot be used. The other motivation was to avoid unnecessary String concatenations if the prefix is empty.

I'm in course to run a test crawl with the changes applied, I'll let you know about the results later today.

jnioche:
cleanup -> found your previous comment. Let's remove that method then.

test-> thanks, I'll wait to hear from you before I merge

sebastian-nagel:
Hi @jnioche, the test crawl went well. Passing of protocol metadata to parsers, WARC writers seems to work. Unfortunately, I forgot to set http.use.cookies = true in the config - but cookies are saved in ES status index in metadata ("protocol%2Eset-cookie").

jnioche:
Merged. Thanks @sebastian-nagel

Issue #755, opened 2019-09-23
Implement WARC spout


sebastian-nagel:
A "WARC spout" could read records from WARC files and emit tuples <URL, content, metadata> into the topology, similar as done by FetcherBolt. This would allow
- to reproducibely test and benchmark the topology (parser, indexer and status update bolts) with the same "frozen" data
- to use parts of a crawl topology to process historic data from WARC files using the same parsers and indexers
- to easily rewrite or (re)index WARC files, cf. #567

kkrugler:
Just FYI, there's a Common Crawl fetcher that's part of the flink-crawler project, which we used extensively during testing as a way of having a large-scale, reproducible crawl. See https://github.com/ScaleUnlimited/flink-crawler/tree/master/src/main/java/com/scaleunlimited/flinkcrawler/fetcher/commoncrawl

sebastian-nagel:
Thanks, @kkrugler: a fetcher which gets the content from WARC files via a CDX index would be a further step. For now, my idea was simpler: just consume a list of WARC files and emit all records into the topology. It's just a "replay" of an already recorded crawl.

Issue #567, opened 2018-05-05
wish: WARCHdfsBolt with CDX index


dportabella:
StormCrawler allows to filter web pages and archive them into WARC archives, as follows:
```
WARCHdfsBolt warcbolt = (WARCHdfsBolt) new WARCHdfsBolt().withFileNameFormat(fileNameFormat);

TopologyBuilder builder = new TopologyBuilder();

builder.setBolt("warc", warcbolt, numWorkers)
  .localOrShuffleGrouping("parse", WarcStreamName)
  .localOrShuffleGrouping("tika",  WarcStreamName);
```
Would it be possible to create a [CDX index](https://iipc.github.io/warc-specifications/specifications/cdx-format/cdx-2015/) (or JCDX index) for the WARC archives at the same time?

sebastian-nagel:
Should be possible:
- cf. [WarcCdxWriter](https://github.com/commoncrawl/nutch/blob/cc/src/java/org/commoncrawl/warc/WarcCdxWriter.java) for Nutch
- should ev. exclude most of the dependencies of [webarchive-commons](https://mvnrepository.com/artifact/org.netpreserve.commons/webarchive-commons) which is required to get the SURT key

jnioche:
Doable indeed. 

@dportabella the warc bolt is usually connected to the fetcher, not the parsers

`        builder.setBolt("warc", warcbolt).localOrShuffleGrouping("fetch");
`

sebastian-nagel:
Alternatively, the WARC bolt could add WARC file name, record offset and length to the metadata. An indexer (CDX or anything else) then could store it directly which obsoletes the need to index the CDX files in a separate step.

PR #799, opened 2020-05-27
WARC spout to emit captures into topology (implements #755)

sebastian-nagel:
A simple WARC spout:
- inherits from FileSpout to read paths/URLs from local files
- based on [jwarc](//github.com/iipc/jwarc)
- mimics the behavior of FetcherBolt, i.e.
  - all captures are sent to status stream
  - only successful fetches (HTTP status 200) will emit content to default stream for parsing, indexing, etc.

Tested so far only with a few WARC files, both local and Common Crawl files referenced by URL `http://commoncrawl.s3.amazonaws.com/crawl-data/...`

### DIFF
--- a/external/warc/src/main/java/com/digitalpebble/stormcrawler/warc/WARCSpout.java
+++ b/external/warc/src/main/java/com/digitalpebble/stormcrawler/warc/WARCSpout.java
@@ -1,3 +1,20 @@
+/**
+ * Licensed to DigitalPebble Ltd under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * DigitalPebble licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package com.digitalpebble.stormcrawler.warc;
 
 import java.io.IOException;
@@ -25,6 +42,7 @@ import org.netpreserve.jwarc.HttpRequest;
 import org.netpreserve.jwarc.HttpResponse;
 import org.netpreserve.jwarc.MediaType;
 import org.netpreserve.jwarc.MessageBody;
+import org.netpreserve.jwarc.ParsingException;
 import org.netpreserve.jwarc.WarcPayload;
 import org.netpreserve.jwarc.WarcReader;
 import org.netpreserve.jwarc.WarcRecord;
@@ -38,13 +56,12 @@ import com.digitalpebble.stormcrawler.Constants;
 import com.digitalpebble.stormcrawler.Metadata;
 import com.digitalpebble.stormcrawler.persistence.Status;
 import com.digitalpebble.stormcrawler.protocol.ProtocolResponse;
-import com.digitalpebble.stormcrawler.protocol.ProtocolResponse.TrimmedContentReason;
 import com.digitalpebble.stormcrawler.spout.FileSpout;
 import com.digitalpebble.stormcrawler.util.ConfUtils;
 
 /**
  * Read WARC files from the local files system and emit the WARC captures as
- * tuples into the topology same ways as done by
+ * tuples into the topology same way as done by
  * {@link com.digitalpebble.stormcrawler.bolt.FetcherBolt}.
  */
 @SuppressWarnings("serial")
@@ -61,6 +78,7 @@ public class WARCSpout extends FileSpout {
     private WarcReader warcReader;
     private String warcFileInProgress;
     private WarcRequest precedingWarcRequest;
+    private Optional<WarcRecord> record;
 
     private MultiCountMetric eventCounter;
 
@@ -70,16 +88,6 @@ public class WARCSpout extends FileSpout {
 
     public WARCSpout(String dir, String filter) {
         super(dir, filter, false);
-    }
-
-    public static class RecordHolder {
-        WarcRecord record;
-        byte[] content;
-
-        public RecordHolder(WarcRecord record, byte[] content) {
-            this.record = record;
-            this.content = content;
-        }
     }
 
     /**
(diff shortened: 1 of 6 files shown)


jnioche:
Nice one @sebastian-nagel! 

Would some of the existing components of the WARC module benefit from **jwarc**?

Our code to generate WARCs leverages **storm-hdfs** (and can of course deal with local files as well), why not have a spout based on HdfsSpout instead? This way we could deal with remote and or larger input data?

sebastian-nagel:
> Would some of the existing components of the WARC module benefit from **jwarc**?

Maybe, it's on my list to try to rewrite the WARCHdfsBolt using jwarc.

> why not have a spout based on HdfsSpout instead?

After a look on [HDFS spout](https://github.com/apache/storm/blob/master/docs/storm-hdfs.md#hdfs-spout): the way files are marked before and moved after processing seems not really applicable for web archives - those usually stay in place. We would need the same mechanism as in the first version of WarcSpout: write paths/URLs pointing to WARCs into files and put these in the input folder of HdfsSpout. I'll think about it. Maybe it's the next step.

jnioche:
makes sense

Could you please update the README for the WARC module as part of the PR? Good to go otherwise

thanks

sebastian-nagel:
Yes, I'll update the README. Give me a few days to do some more testing - I've found already one open point: WARC records with payload still compressed using Content-Encoding.

sebastian-nagel:
Rebased to current master and addressed the following points:
- added info about warc spout to README
- store WARC record location to metadata (warc.file.name, warc.record.offset and warc.record.length), useful for debugging if the content caused a parser failure, related to #567 
- upgrade jwarc dependency to 0.12.0
- add configurable sleep after fetches to work around OOMs due to large overflow queues

Open points:
- HTTP Content-Encoding is not decoded. Waiting whether a method for this could be made available in jwarc, see iipc/jwarc#48
- a single WARC spout is able to emit 2,000-3,000 "fetches" per sec. which makes any topology consuming the content reliably fail by an OOM caused by >20k tuples in the overflow queue. I've tried without success: shrink `topology.max.spout.pending`, backpressure enabled/disabled and various values for topology.message.timeout.secs (30-300 sec.)


Testing:
- "dummy" topology in local mode: WARCSpout ( -> DummyIndexer ) -> StdOutStatusUpdater
  - processed 70 WARC files
  - average: 2,400 WARC records per sec.
  - no sleep
- cluster mode, with 10 ms. sleep after "fetches":
  - rewrite WARC files: WARCSpout -> WARCHdfsBolt
  - parse and index content

external/warc/src/main/java/com/digitalpebble/stormcrawler/warc/WARCSpout.java
```diff
@@ -0,0 +1,530 @@
...
+                populateBuffer();
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+        }
+
+        if (warcReader == null && buffer.isEmpty()) {
+            // input exhausted
+            try {
+                Thread.sleep(10);
```
jnioche, line 379: don't need it -  Storm will handle sleeping between calls to nextTuple if you don't emit anything during a call

external/warc/src/main/java/com/digitalpebble/stormcrawler/warc/WARCSpout.java
```diff
@@ -0,0 +1,530 @@
...
+        }
+
+        byte[] head = buffer.removeFirst();
+        List<Object> fields = _scheme.deserialize(ByteBuffer.wrap(head));
+        warcFileInProgress = (String) fields.get(0);
+        if (warcFileInProgress == null)
+            return;
+
+        LOG.info("Reading WARC file {}", warcFileInProgress);
+        ReadableByteChannel warcChannel = null;
```
jnioche, line 136: try with resource to simplify the code and remove the close section below?

external/warc/src/main/java/com/digitalpebble/stormcrawler/warc/WARCSpout.java
```diff
@@ -0,0 +1,530 @@
...
+            if (warcType == null) {
+                LOG.warn("No type for {}", record.get().getClass());
+            } else {
+                eventCounter.scope(
+                        "warc_skipped_record_of_type_" + record.get().type())
+                        .incr();
+                LOG.debug("Skipped WARC record of type {}",
+                        record.get().type());
+            }
+            if (storeHTTPHeaders && "request".equals(warcType)) {
```
jnioche, line 399: that string "request" is used in other places in the module - have it as a constant of enum somewhere?

jnioche:
Implementing our own timeout mechanism is not right, max spout pending should be the way to go.

Is there more than one _emit_ per call to _nextTuple()_? if so this would explain why _topology.max.spout.pending_ doesn't have any effect. 

Looking at the code this does not seem to be the case though, any idea?

external/warc/src/main/java/com/digitalpebble/stormcrawler/warc/WARCSpout.java
sebastian-nagel: Ok, will remove it - was following the API doc of [ISpout::nextTuple](https://storm.apache.org/releases/1.2.3/javadocs/org/apache/storm/spout/ISpout.html#nextTuple--), obviously outdated.

jnioche: Let me double check in Storm's code

external/warc/src/main/java/com/digitalpebble/stormcrawler/warc/WARCSpout.java
sebastian-nagel: Good point. Thanks!

external/warc/src/main/java/com/digitalpebble/stormcrawler/warc/WARCSpout.java
jnioche: see http://storm.apache.org/releases/current/Performance.html -> wait strategy

default values 

_topology.sleep.spout.wait.strategy.time.ms | 1
topology.spout.wait.strategy | "org.apache.storm.spout.SleepSpoutWaitStrategy"_

external/warc/src/main/java/com/digitalpebble/stormcrawler/warc/WARCSpout.java
sebastian-nagel: Actually, it does not work: the channel needs to stay, it'll be closed together by `warcReader.close()`.

external/warc/src/main/java/com/digitalpebble/stormcrawler/warc/WARCSpout.java
sebastian-nagel: Ok. But maybe better use `instanceof WarcRequest` as the literal "request" is an internal of jwarc.

external/warc/src/main/java/com/digitalpebble/stormcrawler/warc/WARCSpout.java
jnioche: ok!

external/warc/src/main/java/com/digitalpebble/stormcrawler/warc/WARCSpout.java
sebastian-nagel: Ok: sleep removed.

sebastian-nagel:
Hi @jnioche, I've addressed all your comments. Some example topologies which push WARC content into the topology are in [sebastian-nagel/warc-crawler](/sebastian-nagel/warc-crawler). The "devnull" topology reads 1.2 million WARC records in 9 min. on my laptop. However, the workers of the "rewarc" topo repeatedly crash with an OOM exception. An analysis of the heap dump shows that almost the entire memory is occupied by the overflow queue (cf. [here](http://mail-archives.apache.org/mod_mbox/storm-user/201705.mbox/%3C913EBAD0-2431-469E-B1D9-B2E07DFF9725@hortonworks.com%3E)). Some 20k tuples are emitted but not yet acked:
Screenshot_2020-06-23StormUI-rewarc
The WARCHdfsBolt is considerably slow (checksums and gzip are computationally expensive), but looks like that max. pending is not applied (maybe due to a misconfiguration).

jnioche:
Thanks @sebastian-nagel I can reproduce the issue and will look into it.

jnioche:
You need to emit with a tuple ID, otherwise, Storm won't track the acks and fails and the number of tuples being processed.

`            _collector.emit(new Values(url, content, metadata), url);
`
you might need to set   `topology.backpressure.enable: false` for `max.spout.pending` to be effective

The change above seems to have fixed the issue in my case.

external/warc/src/main/java/com/digitalpebble/stormcrawler/warc/WARCSpout.java
```diff
@@ -0,0 +1,520 @@
...
+ * {@link com.digitalpebble.stormcrawler.bolt.FetcherBolt}.
+ */
+@SuppressWarnings("serial")
+public class WARCSpout extends FileSpout {
+
+    private static final Logger LOG = LoggerFactory.getLogger(WARCSpout.class);
+
+    private int maxContentSize = -1;
+    private int contentBufferSize = 8192;
+    private long sleepEmitFetched = 0;
```
jnioche, line 74: not used - remove?

sebastian-nagel: Done.

sebastian-nagel:
Thanks, @jnioche! I've added the missing ID to the emit call. I've run only a short test: looks good, memory consumption is low.

jnioche:
Thanks @sebastian-nagel, this is a great addition to our WARC module.

Issue #809, opened 2020-06-30
AggregationSpout error due SimpleDateFormat not thread safe 


jcruzmartini:
sometimes when the crawl is finishing and we only have few URLs pending, the `nextTuple()` in the aggregation spout is being called steadily (totally expected). If you have the property `es.status.concurrentRequests` in a number greater than 1 and your property `spout.min.delay.queries` is too low, you may get this error

```
java.io.IOException: Unable to parse response body for Response{requestLine=POST /status*/_search?typed_keys=true&ignore_unavailable=false&expand_wildcards=open&allow_no_indices=true&preference=_shards%3A8&ignore_throttled=true&search_type=query_then_fetch&batched_reduce_size=512&ccs_minimize_roundtrips=true HTTP/1.1, host=https://elasticsearch-coordinating:9200, response=HTTP/1.1 200 OK}
	at org.elasticsearch.client.RestHighLevelClient$1.onSuccess(RestHighLevelClient.java:1665) [stormjar.jar:?]
	at org.elasticsearch.client.RestClient$FailureTrackingResponseListener.onSuccess(RestClient.java:590) [stormjar.jar:?]
	at org.elasticsearch.client.RestClient$1.completed(RestClient.java:333) [stormjar.jar:?]
	at org.elasticsearch.client.RestClient$1.completed(RestClient.java:327) [stormjar.jar:?]
	at org.apache.http.concurrent.BasicFuture.completed(BasicFuture.java:122) [stormjar.jar:?]
	at org.apache.http.impl.nio.client.DefaultClientExchangeHandlerImpl.responseCompleted(DefaultClientExchangeHandlerImpl.java:181) [stormjar.jar:?]
	at org.apache.http.nio.protocol.HttpAsyncRequestExecutor.processResponse(HttpAsyncRequestExecutor.java:448) [stormjar.jar:?]
 	at org.apache.http.nio.protocol.HttpAsyncRequestExecutor.inputReady(HttpAsyncRequestExecutor.java:338) [stormjar.jar:?]
 	at org.apache.http.impl.nio.DefaultNHttpClientConnection.consumeInput(DefaultNHttpClientConnection.java:265) [stormjar.jar:?]
 	at org.apache.http.impl.nio.client.InternalIODispatch.onInputReady(InternalIODispatch.java:81) [stormjar.jar:?]
[...]

PR #810, opened 2020-06-30
#809 - Fixing issue with SDF not thread safe

jcruzmartini:
Fixing issue related due SDF not thread-safe

### DIFF
--- a/external/elasticsearch/src/main/java/com/digitalpebble/stormcrawler/elasticsearch/persistence/AggregationSpout.java
+++ b/external/elasticsearch/src/main/java/com/digitalpebble/stormcrawler/elasticsearch/persistence/AggregationSpout.java
@@ -19,8 +19,6 @@ package com.digitalpebble.stormcrawler.elasticsearch.persistence;
 
 import static org.elasticsearch.index.query.QueryBuilders.boolQuery;
 
-import java.text.ParseException;
-import java.text.SimpleDateFormat;
 import java.time.Instant;
 import java.util.Calendar;
 import java.util.Date;
@@ -78,9 +76,6 @@ public class AggregationSpout extends AbstractSpout implements
     private static final String ESStatusSampleParamName = "es.status.sample";
     private static final String ESMostRecentDateIncreaseParamName = "es.status.recentDate.increase";
     private static final String ESMostRecentDateMinGapParamName = "es.status.recentDate.min.gap";
-
-    private static final SimpleDateFormat formatter = new SimpleDateFormat(
-            "yyyy-MM-dd'T'HH:mm:ss.SSSX");
     
     private boolean sample = false;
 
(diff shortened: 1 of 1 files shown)


external/elasticsearch/src/main/java/com/digitalpebble/stormcrawler/elasticsearch/persistence/AggregationSpout.java
```diff
@@ -255,8 +256,9 @@ public void onResponse(SearchResponse response) {
                 if (hitsForThisBucket == 1 && !iterator.hasNext()) {
                     String strDate = (String) keyValues.get("nextFetchDate");
                     try {
-                        mostRecentDateFound = formatter.parse(strDate);
-                    } catch (ParseException e) {
+                        LocalDateTime ldt = LocalDateTime.parse(strDate, formatter);
+                        mostRecentDateFound = Date.from(ldt.atZone(ZoneId.systemDefault()).toInstant());
```
jcruzmartini, line 260: using `ZoneId.systemDefault()`, since when we are creating the calendar using Calendar.getInstance() few lines bellow, we are system default too

jnioche: Do we need to go via a Date?

Instant instant = ldt.toInstant(ZoneOffset.UTC);

with the default instead of ZOneOffset?

Or even better, if the ISO format matches the pattern we have above, wouldn't 

https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html#parse-java.lang.CharSequence-

do the trick?

jcruzmartini: I will try, I like your suggestion...

external/elasticsearch/src/main/java/com/digitalpebble/stormcrawler/elasticsearch/persistence/AggregationSpout.java
```diff
@@ -255,8 +250,8 @@ public void onResponse(SearchResponse response) {
                 if (hitsForThisBucket == 1 && !iterator.hasNext()) {
                     String strDate = (String) keyValues.get("nextFetchDate");
                     try {
-                        mostRecentDateFound = formatter.parse(strDate);
-                    } catch (ParseException e) {
+                        mostRecentDateFound = Instant.parse(strDate);
```
jcruzmartini, line 253: @jnioche it seems working with your proposal... I will be running some test in my cluster to guarantee that everything is working as expected... will add the updates here

jnioche: @jcruzmartini how is it going?

jcruzmartini: @jnioche I run an end to end test this morning... everything is looking great. this PR is ready to be merged. thanks

jnioche:
thanks @jcruzmartini

PR #812, opened 2020-07-07
AdaptiveScheduler to set last-modified time in metadata, fixes #777

### LINKED ISSUE #777: status index / Adaptive scheduler: get rid of last-modified dates in non-ISO date format

sebastian-nagel:
- this PR assumes that the HTTP last-modified field is kept separately, see `protocol.md.prefix` and #776
- set `last-modified` in metadata
  - for the initial successful fetch (also after error status)
  - and if a change of the content signature indicates a modification
- clear `last-modified` for permanent failures (status ERROR)
- set `fetchInterval` in metadata for initial fetches scheduled by DefaultScheduler

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/persistence/AdaptiveScheduler.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/persistence/AdaptiveScheduler.java
@@ -17,6 +17,7 @@
 
 package com.digitalpebble.stormcrawler.persistence;
 
+import java.time.Duration;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.Locale;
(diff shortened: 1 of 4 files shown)


jnioche:
brilliant, thanks @sebastian-nagel

sebastian-nagel:
I'll also update the wiki, later today or tomorrow.

Issue #749, opened 2019-09-09
Enable extension parsing for SitemapParser 


jnioche:
https://github.com/crawler-commons/crawler-commons/pull/218/ introduced support for sitemap extensions. 
These are not active by default and should be made configurable. The extension data found (if any) would be added to the outlink metadata and could therefore be used by a ParseFilter.

evanhalley:
I'm curious about the implementation for this one.  It looks the sitemap extension attributes are regular java objects (instead of a Map<String,String> for example).  This means we would essentially need a helper class that can convert an extension attributes class from POJO to Strings that we could then store in the metadata object attached to the Outlink.  You could then access them by the keys as specified in the Sitemap XML.  

For example, using the key **video:duration** to get the video duraiton out of the metadata object.

Would an implementation like this be a solution for this issue?

sebastian-nagel:
Hi @evanhalley, yes and some of the attributes are itself complex objects, eg. "video:price".

Maybe it's better to split the implementation into two parts:
1. add `asMap()` methods to the attribute objects in [crawler-commons](https://github.com/crawler-commons/crawler-commons/). That seems better maintainable than implementing it in storm-crawler. There are already `toString()` methods for the attributes. The `asMap()` method should return a Map<String, String[]> because some attributes are multi-valued, eg. video:price which can be specified using multiple currencies.
2. in storm-crawler: add the code to enable the sitemap extensions by a configuration property and to add the attributes to the metadata. Maybe it's worth to discuss the following points:
   - enable all extensions or make it configurable which extensions
   - (optionally) prefix the metadata keys - `sitemap.video:price`, cf. #776 

What's your opinion?

jnioche:
thanks @evanhalley and @sebastian-nagel 

I had a first prototype to deal with images only and looked like that 

```
  Map<Extension, ExtensionMetadata[]> attributes = smurl.getAttributes();
                if (attributes != null) {
                    // handle image only for now
                    ExtensionMetadata[] emds = attributes.getOrDefault(Extension.IMAGE,
                            new ExtensionMetadata[] {});
                    for (ExtensionMetadata emd : emds) {
                        ImageAttributes e = (ImageAttributes) emd;
                        String val = e.getCaption();
                        if (val != null) {
                            customKeyVal.add(Extension.IMAGE.name()+".caption");
                            customKeyVal.add(val);
                        }
                        val = e.getGeoLocation();
                        if (val != null) {
[...]

PR #815, opened 2020-07-16
Implemented Sitemap extension parsing (issue #749)

evanhalley:
Implemented Sitemap extension parsing (issue #749)
- extracting the extension attributes out of each extension object for a given Sitemap URL added adding it to the outlink metadata object (prefixed by the extension type)

Somewhat related, while implementing this, I noticed that crawler-commons isn't adding the `duration` attribute from the `VideoExtension` to the map generated by `VideoExtension.asMap()` (oversight on my part).  I wrote this up as an issue, [crawler-commons/300](https://github.com/crawler-commons/crawler-commons/issues/300) and submitted a PR.  A future release will include duration, however, it will be missing from StormCrawler until a new version of `crawler-commons` is published with the fix.


jnioche:
Fab, thanks @evanhalley

PR #823, opened 2020-09-23
WARCSpout: add metadata field "fetch.statusCode" (HTTP status code)

sebastian-nagel:
Same as (Simple)FetcherBolt WARCSpout should add "fetch.statusCode" which holds HTTP status code to metadata. It's required by AdaptiveScheduler which will fail with a NPE if the field is not provided.


jnioche:
Thanks @sebastian-nagel

PR #826, opened 2020-10-01
WARCSpout/FileSpout: ClassCastException if message ID of failed tuples is not of type byte[]

### LINKED ISSUE #825: WARCSpout/FileSpout: ClassCastException if message ID of failed tuples is not of type byte[]
When a tuple emitted by WARCSpout fails downstream, the method `fail(Object msgId)` inherited from FileSpout throws a ClassCastException:
```
2020-09-30 12:50:16.983 o.a.s.util Thread-16-spout-executor[5 5] [ERROR] Async loop died!
java.lang.RuntimeException: java.lang.ClassCastException: java.lang.String cannot be cast to [B
        at org.apache.storm.utils.DisruptorQueue.consumeBatchToCursor(DisruptorQueue.java:522) ~[storm-core-1.2.3.jar:1.2.3]
        ...
Caused by: java.lang.ClassCastException: java.lang.String cannot be cast to [B
        at com.digitalpebble.stormcrawler.spout.FileSpout.fail(FileSpout.java:257) ~[stormjar.jar:?]
        ...
        ... 7 more
2020-09-30 12:50:17.113 o.a.s.util Thread-16-spout-executor[5 5] [ERROR] Halting process: ("Worker died")
```

Also FileSpout itself (not an extending class) may run into this issue, in case `withDiscoveredStatus` is true which makes FileSpout to use the URL string as message ID.

sebastian-nagel:
fixes #825

- FileSpout: check for type of message ID
    byte[] : replay
    String : log error, cannot replay
    (anything else) : log error and throw Exception
- WARCSpout: log error, cannot replay

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/spout/FileSpout.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/spout/FileSpout.java
@@ -254,8 +254,19 @@ public class FileSpout extends BaseRichSpout {
 
     @Override
     public void fail(Object msgId) {
-        String msg = new String((byte[]) msgId);
-        LOG.error("Failed - adding back to the queue: {}", msg);
-        buffer.add((byte[]) msgId);
+        if (msgId instanceof byte[]) {
+            String msg = new String((byte[]) msgId);
+            LOG.error("Failed - adding back to the queue: {}", msg);
+            buffer.add((byte[]) msgId);
+        } else if (withDiscoveredStatus && msgId instanceof String) {
+            // messageID is the injected URL
+            LOG.error("Failed - cannot replay URL without metadata: {}", msgId);
+        } else {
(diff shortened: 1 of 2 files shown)


jnioche:
we should really replace in FileSpout

```
 this._collector.emit(Constants.StatusStreamName, fields, fields
                    .get(0).toString());
```

with 

```
            this._collector.emit(Constants.StatusStreamName, fields, head);
```
to make the behaviour consistent in the ack / fail. also this way we'll be able to replay with the metadata correctly

WarcSpout -> indeed it should handle its own failures to match whatever it does with  _collector.emit

sebastian-nagel:
> we should really replace in FileSpout

Ok, I'll update the PR.

jnioche:
thanks Sebastian

Issue #827, opened 2020-10-05
HTTP protocol implementation: allow to configure which protocol version(s) to use


sebastian-nagel:
It should be possible to configure which HTTP protocol versions (HTTP/1.1, HTTP/2, etc.) are used by a protocol implementation:
- WARC readers may fail on yet unknown HTTP protocol versions (see [discussion "Is there an option to use HTTP/1.0 or 1.1 for the Warc File"](https://groups.google.com/g/digitalpebble/c/VgoLRAnQDV0/m/bUrod9jxAAAJ), commoncrawl/news-crawl#42)
- but could be also useful for debugging and testing

jnioche:
fixed in #829

PR #829, opened 2020-10-05
HTTP protocol implementation: allow to configure which protocol version(s) to use

sebastian-nagel:
implements #827
- configuration key `http.protocol.versions` holds a list of protocols in order of preference
- implement selection of protocols in okhttp protocol implementation (httpclient should pick this once HTTP/2 is supported)
- fix loading of YAML lists when configuration isn't loaded via Storm (eg. by main method of AbstractHttpProtocol)
- add notice about protocol version to warc module README


jnioche:
nice one thanks @sebastian-nagel 
could you please document it in the wiki? thanks

sebastian-nagel:
Done: updated https://github.com/DigitalPebble/storm-crawler/wiki/Protocols, the warc module README is already updated by this PR.

PR #831, opened 2020-10-06
WARCSpout to add `_request.time_` to metadata

sebastian-nagel:
Add the capture time from the WARC header to metadata field `_request.time_` same as done by the okhttp protocol. Of course, the capture time is useful when indexing the archive content, it's even required if WARC files are re-archived.


jnioche:
Thanks @sebastian-nagel

PR #853, opened 2021-01-27
WARCSpout doesn't handle http.content.limit -1 correctly, fixes #850

### LINKED ISSUE #850: WARCSpout doesn't handle http.content.limit -1 correctly
-1 is the value it takes by default if not set in the config and it means that the content should not be trimmed, just like the Fetcher Bolts do, however it doesn't handle it correctly and tries to trim to a negative value.
A temporary workaround is to configure it with a large value

`  http.content.limit: 99999999`

sebastian-nagel:
- if http.content.limit is -1 trim content to the max. possible array size

The following issues where fixes to allow testing with 3 WARC files holding content of [2^31-8 (safe)](https://github.com/DigitalPebble/storm-crawler/files/5880994/test-size-int-max-safe.warc.gz), [2^31-1 (int_max)](https://github.com/DigitalPebble/storm-crawler/files/5880999/test-size-int-max.warc.gz) and [2^31 bytes](https://github.com/DigitalPebble/storm-crawler/files/5881003/test-size-int-max-overflow.warc.gz) and to fix bugs detected while testing these edge cases:
- when reading into multiple buffers: do not read more bytes than necessary
- avoid an integer overflow when reporting length of truncation
- implement gzip and deflate Content-Type decompression

### DIFF
--- a/external/warc/src/main/java/com/digitalpebble/stormcrawler/warc/WARCSpout.java
+++ b/external/warc/src/main/java/com/digitalpebble/stormcrawler/warc/WARCSpout.java
@@ -40,8 +40,8 @@ import org.apache.storm.tuple.Values;
 import org.netpreserve.jwarc.HttpMessage;
 import org.netpreserve.jwarc.HttpRequest;
 import org.netpreserve.jwarc.HttpResponse;
+import org.netpreserve.jwarc.IOUtils;
 import org.netpreserve.jwarc.MediaType;
-import org.netpreserve.jwarc.MessageBody;
 import org.netpreserve.jwarc.ParsingException;
 import org.netpreserve.jwarc.WarcPayload;
 import org.netpreserve.jwarc.WarcReader;
@@ -96,7 +96,7 @@ public class WARCSpout extends FileSpout {
      */
     public static class TruncationStatus {
         boolean isTruncated = false;
-        int originalSize = -1;
+        long originalSize = -1;
 
         public void set(boolean isTruncated) {
             this.isTruncated = isTruncated;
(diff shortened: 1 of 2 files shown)


external/warc/src/main/java/com/digitalpebble/stormcrawler/warc/WARCSpout.java
```diff
@@ -59,6 +59,8 @@
 import com.digitalpebble.stormcrawler.spout.FileSpout;
 import com.digitalpebble.stormcrawler.util.ConfUtils;
 
+import parquet.it.unimi.dsi.fastutil.Arrays;
```
jnioche, line 62: I presume this is already in our dependencies but I'd rather just have a static final somewhere in our code

kkrugler: Is this a shaded class? It's `it.unimi.dsi.fastutil.Arrays` in my code. If it's shaded, you don't want to be using it in your code.

sebastian-nagel: Ok, I'll change it. Also to be used for #854.

sebastian-nagel: Ok, it's now defined in the class Constants:
```java
/** Maximum array size, safe value on any JVM */
public static final int MAX_ARRAY_SIZE = Integer.MAX_VALUE - 8;
```

jnioche:
thanks @sebastian-nagel and @kkrugler

Issue #861, opened 2021-03-02
Do not generate a nextFetchDate at all if the scheduling is set to NEVER


jnioche:
Setting a date in a far distant future is a way of preventing URLs to be ever returned, however, the same result would be achieved by not setting a fetch date at all. The benefit of the latter would be a gain in storage space and possibly faster queries (to be tested).
An URL can always be updated with a nextFetchDate later on if necessary.

jnioche:
The scheduler interface could return an _Optional<Date>_ or an integer for a timestamp, with a value of -1 in case a URL doesn't need refetching. Relying on  _DefaultScheduler.NEVER_ to determine whether a _nextFetchDate_ should not be written is not ideal; there is no guarantee that the DefaultScheduler is used.

PR #866, opened 2021-03-19
Scheduler implementations return an Optional<Date>

jnioche:
implements #861

Using Optional<Date> would have worked but require a big change in the API whereas _null_ allows to keep it as it is.

@sebastian-nagel any thoughts?

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/persistence/AbstractStatusUpdaterBolt.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/persistence/AbstractStatusUpdaterBolt.java
@@ -22,6 +22,7 @@ import java.util.Calendar;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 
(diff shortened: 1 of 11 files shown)


jnioche:
Using _null_ is dirty. Went for Optional<Date> instead.

sebastian-nagel (changes requested):
Looks good.

sebastian-nagel:
One question: so far there's no way to specify a rule for DefaultScheduler so that match items are never refetched. E.g.
```yaml
# never re-fetch spammy pages
fetchInterval.FETCHED.isSpam=true: -1
```

jnioche:
> One question: so far there's no way to specify a rule for DefaultScheduler so that match items are never refetched, right? E.g.
> 
> ```yaml
> # never re-fetch spammy pages
> fetchInterval.FETCHED.isSpam=true: -1
> ```

Good catch! I have fixed that as part of the latest commit.

jnioche:
thanks for the review @sebastian-nagel - please have a look at the latest commit

sebastian-nagel (approved):

Issue #892, opened 2021-07-14
Bug: StackOverFlow issue in CharsetIdentification 


vitos73:
If a website has erroneous content we've got a crash of crawler with a java.lang.StackOverflowError in com.digitalpebble.stormcrawler.util.CharsetIdentification.getCharsetFromMeta(CharsetIdentification.java:192)

Example site: http://gurpreetkaur.co.in/
Example content (exact)
```
<!DOCTYPE html>
--
  | <!--[if IE 7]><html class="ie ie7 ltie8 ltie9" lang="en-US"><![endif]-->
  | <!--[if IE 8]><html class="ie ie8 ltie9" lang="en-US"><![endif]-->
  | <!--[if !(IE 7) \| !(IE 8)  ]><!-->
  | <html lang="en-US">
  | <!--<![endif]-->
  |  
  | <head>
  | <meta charset="UTF-8
```
Problem with comparison in line 191 CharsetIdentification.java:
was: 
```
            if (end == -1) {
                return getCharsetFromMeta(buffer, maxlength + 10);
            }
```
should be:
```
             if (end == -1 && ((maxlength +10) < buffer.length)) {
                return getCharsetFromMeta(buffer, maxlength + 10);
            } 
```
We've to ensure that were chars left before EOF.

vitos73:
full solution of fast detection via tag looks like:
 ```
int start = html.indexOf("<meta charset=\"");
        if (start != -1) {
            int end = html.indexOf('"', start + 15);
            // https://github.com/DigitalPebble/storm-crawler/issues/870
            // try on a slightly larger section of text if it is trimmed
            if (end == -1 && ((maxlength +10) < buffer.length)) {
                return getCharsetFromMeta(buffer, maxlength + 10);
            }
            if (end == -1) {
                // there is open tag meta but not closed = we have broken content!
                return null;
            }
            return validateCharset(html.substring(start + 15, end));
        }
```

jnioche:
thanks @vitos73. Do you want to / can you contribute a PR to fix it?

vitos73:
@jnioche 
Done: #895 
Thank you!

PR #895, opened 2021-07-15
Fixed bug with StackOverflowError in fast charset identification

vitos73:
Fixed issue #892 
We have to ensure that were enough content for a recursive call with a bit large window.
If we have found broken the meta tag we should return null.


jnioche:
thanks @vitos73

PR #916, opened 2021-10-26
ES Status configuration : Can override discovered links

Mikwiss:
Add a parameter to override a link in status index with ES StatusUpdaterBolt. 

Context : our topologies _create_ in a filter's ParserBolt some links. Those links aren't in the DOM page, so the classic discover cannot be used. Also, they must not be refetch automatically, only when the filter create it, so the StatusUpdaterBolt have to override the previous fetched one.

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

* You've applied the formatting used by the project with `mvn java-formatter:format`

Thanks!

### DIFF
--- a/external/elasticsearch/archetype/src/main/resources/archetype-resources/es-conf.yaml
+++ b/external/elasticsearch/archetype/src/main/resources/archetype-resources/es-conf.yaml
@@ -30,6 +30,8 @@ config:
   es.status.bulkActions: 500
   es.status.flushInterval: "5s"
   es.status.concurrentRequests: 1
+  # Override discovered status
+  es.status.override: false
   
   ################
   # spout config #
--- a/external/elasticsearch/src/main/java/com/digitalpebble/stormcrawler/elasticsearch/persistence/StatusUpdaterBolt.java
+++ b/external/elasticsearch/src/main/java/com/digitalpebble/stormcrawler/elasticsearch/persistence/StatusUpdaterBolt.java
@@ -73,8 +73,11 @@ public class StatusUpdaterBolt extends AbstractStatusUpdaterBolt implements
     private static final String ESStatusIndexNameParamName = "es.%s.index.name";
     private static final String ESStatusRoutingParamName = "es.%s.routing";
     private static final String ESStatusRoutingFieldParamName = "es.%s.routing.fieldname";
+    private static final String ESStatusOverrideParamName = "es.%s.override";
 
     private boolean routingFieldNameInMetadata = false;
+
+    private boolean overrideStatus = false;
 
     private String indexName;
 
@@ -121,6 +124,9 @@ public class StatusUpdaterBolt extends AbstractStatusUpdaterBolt implements
 
         doRouting = ConfUtils.getBoolean(stormConf, String.format(
                 StatusUpdaterBolt.ESStatusRoutingParamName, ESBoltType), false);
+
+        overrideStatus = ConfUtils.getBoolean(stormConf, String.format(
+                StatusUpdaterBolt.ESStatusOverrideParamName, ESBoltType), false);
 
         partitioner = new URLPartitioner();
         partitioner.configure(stormConf);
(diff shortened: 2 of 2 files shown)


jnioche:
hi @Mikwiss and thanks for the PR. Just to make sure I correctly understand what you are trying to achieve: you are creating custom outlinks in a ParseFilter and as these get a default status of DISCOVERED  they won't overwrite any existing value in ES. Is that right?
The trouble is that neither the Scheduling classes nor the Metadata Transfer ones allow you to modify the status for a URL.
So, yes you might need to do it the way you are suggesting, however, instead of one config that would apply to all the discovered URLs. I would rather have that triggered on a per-URL basis based on some arbitrary key in the metadata, just like we do for instance for specifying a custom _nextFetchDate_. 
I can see 2 options:
1. we do it in ES StatusUpdaterBolt only e.g. we use a key like 'es.status.override'
2. we do it in AbstractStatusUpdaterBolt by changing the status value to FETCHED based on an arbitrary key in the MD. You will need to control the _nextFetchDate_ explicitly, rely on the default setting for it or use a custom scheduler.

I am not sure which one of these 2 approaches I like best: the 2nd one is more generic but then it feels a bit wrong to fiddle with the status; the first one is more straightforward - you'd need to modify AbstractStatusUpdater so we bypass the caching.

What do you think?
Can you share a concrete example of the sort of URLs that you would need this mechanism for? There are often more than one way of doing things with StormCrawler.

Thanks

Mikwiss:
Hi @jnioche ! Thanks for reply !

> Just to make sure I correctly understand what you are trying to achieve: you are creating custom outlinks in a ParseFilter and as these get a default status of DISCOVERED they won't overwrite any existing value in ES. Is that right? 

Exactly ! 

> Can you share a concrete example of the sort of URLs that you would need this mechanism for ?

In order to check our e-commerce frontend, we have to fetch and parse product pages (> 100k). For each product, the ParseFilter have to find a specific _<div>_. If the _<div>_ is found, we have to fetch another URL (the marketplace page). The marketplace URL may or not be emitted as an outlink (btw this link is not in the page as it, we build it), depends on the product and the moment. The default fetch interval for a marketplace URL is set to -1.

> you'd need to modify AbstractStatusUpdater so we bypass the caching.

We disable the cache because we control all the input (product page). In another word, we don't discover any links in the page.

> The trouble is that neither the Scheduling classes nor the Metadata Transfer ones allow you to modify the status for a URL. So, yes you might need to do it the way you are suggesting, however, instead of one config that would apply to all the discovered URLs. I would rather have that triggered on a per-URL basis based on some arbitrary key in the metadata, just like we do for instance for specifying a custom _nextFetchDate_. I can see 2 options:
> 
> 1. we do it in ES StatusUpdaterBolt only e.g. we use a key like 'es.status.override'
> 2. we do it in AbstractStatusUpdaterBolt by changing the status value to FETCHED based on an arbitrary key in the MD. You will need to control the _nextFetchDate_ explicitly, rely on the default setting for it or use a custom scheduler.
> 
> I am not sure which one of these 2 approaches I like best: the 2nd one is more generic but then it feels a bit wrong to fiddle with the status; the first one is more straightforward - you'd need to modify AbstractStatusUpdater so we bypass the caching.
> 
> What do you think? 

The second one is more generic but, as you said, fiddle with the status seems strange and not StromCrawler compliant, juste to workaround ESIndexerbolt behaviour.

jnioche:
thanks for the explanation. Any chance you could amend the PR so that the override is done on  a per URL basis with a key like 'es.status.override' ?

Mikwiss:
Sure, let me try it and i'll feedback you 👍

jnioche:
> Sure, let me try it and i'll feedback you +1

any luck with it @Mikwiss ?

Mikwiss:
Hi @jnioche ! I didn't notice your reply, I'm really sorry. This feature is on my backlog and I do a new pull request as soon as I can treat it !

PR #918, opened 2021-10-29
OkHttp protocol: make connection pool configurable

sebastian-nagel:
[OkHttp's ConnectionPool](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-connection-pool/) by default "holds up to 5 idle connections which will be evicted after 5 minutes of inactivity."  A pool of this size is suitable for site crawls but not for larger crawls over many different sites.

Note: in the current version (4.9.2) the connection pool is implemented as a [linked queue](https://github.com/square/okhttp/blob/parent-4.9.2/okhttp/src/main/kotlin/okhttp3/internal/connection/RealConnectionPool.kt#L83) and searching for a pooled connection does not scale up. In order to scale beyond pool sizes exceeding 1000 a set of clients must be used each with its own connection pool.

Notes:
- so far, only partially tested: need to increase the pool size and run a test crawl to measure the impact
- proxied connections are unchanged, that is, for every fetch a client is created anew and no connection pool is used.  Depending on the proxy manager, it could make sense to define a connection pool ahead and pass it to the [client builder](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/-builder/connection-pool/). Since proxy information is included in the [okhttp address](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-address/) (stored in the connection pool) it should be possible to pool proxied connections.
- the documentation could be moved from crawler-default.yaml to the wiki

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
@@ -240,8 +240,8 @@ public class HttpProtocol extends AbstractHttpProtocol {
             LOG.info(
                     "Using connection pool with max. {} idle connections "
                             + "and {} sec. connection keep-alive time",
-                    time,
-                    size);
+                    size,
+                    time);
         }
 
         client = builder.build();
--- a/core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
@@ -232,7 +232,7 @@ public class HttpProtocol extends AbstractHttpProtocol {
         builder.addInterceptor(BrotliInterceptor.INSTANCE);
 
         Map<String, Object> connectionPoolConf =
-                (Map<String, Object>) conf.get("http.protocol.connection.pool");
+                (Map<String, Object>) conf.get("okhttp.protocol.connection.pool");
         if (connectionPoolConf != null) {
             int size = ConfUtils.getInt(connectionPoolConf, "max.idle.connections", 5);
             int time = ConfUtils.getInt(connectionPoolConf, "connection.keep.alive", 300);
--- a/core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
@@ -45,6 +45,7 @@ import javax.net.ssl.TrustManager;
 import javax.net.ssl.X509TrustManager;
 import okhttp3.Call;
 import okhttp3.Connection;
+import okhttp3.ConnectionPool;
 import okhttp3.Credentials;
 import okhttp3.EventListener;
 import okhttp3.EventListener.Factory;
@@ -229,6 +230,19 @@ public class HttpProtocol extends AbstractHttpProtocol {
 
         // enable support for Brotli compression (Content-Encoding)
         builder.addInterceptor(BrotliInterceptor.INSTANCE);
+
+        Map<String, Object> connectionPoolConf =
+                (Map<String, Object>) conf.get("http.protocol.connection.pool");
+        if (connectionPoolConf != null) {
+            int size = ConfUtils.getInt(connectionPoolConf, "max.idle.connections", 5);
+            int time = ConfUtils.getInt(connectionPoolConf, "connection.keep.alive", 300);
+            builder.connectionPool(new ConnectionPool(size, time, TimeUnit.SECONDS));
+            LOG.info(
+                    "Using connection pool with max. {} idle connections "
+                            + "and {} sec. connection keep-alive time",
+                    time,
+                    size);
+        }
 
         client = builder.build();
     }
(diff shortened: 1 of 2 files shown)


sebastian-nagel:
(rebased to master to resolve merge conflicts)

jnioche:
Looks good, thanks @sebastian-nagel. Will give it a try

jnioche:
I need to test it more. I haven't been able to reproduce the situation where a lot of CPU is used in the first place.
An easy fix could also be to simply declare more instances of the Fetcher bolts - it should have the same effect. I will try to find some time to run a large crawl and see if I can reproduce the problem. thanks @sebastian-nagel

jnioche:
@sebastian-nagel there is at least another [protocol implementation](https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/java/com/digitalpebble/stormcrawler/protocol/selenium/RemoteDriverProtocol.java)  that requires multiple instances and one for Playwright currently in a PR.

Instead of handling this in one implementation, why not add that at [ProtocolFactory](https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/java/com/digitalpebble/stormcrawler/protocol/ProtocolFactory.java) level? We'd go from     

`private final HashMap<String, Protocol> cache = new HashMap<>();
`
to something like

`private final HashMap<String, Protocol[]> cache = new HashMap<>();
`
what do you think?

sebastian-nagel:
> simply declare more instances of the Fetcher bolts

Well, yes and no. A Storm bolt has less overhead than a MapReduce task which runs in a separate JVM. But there's still some overhead. See also below.

> at least another protocol implementation that requires multiple instances and one for Playwright currently in a PR.

But for the connection pools it's required that the same pool is used for the same host (implemented via hash partitioning). Do you want to change this for the Selenium/Playwright protocols? Would make sense in case browser caches aren't shared between protocol instances.

While looking into this, I became aware of one difference between Nutch and StormCrawler:
- Nutch has exactly one protocol instance (because it's a plugin and every plugin is instantiated only once)
- StormCrawler has
  - 1 ProtocolFactory per FetcherBolt and
  - (if URL filters include RobotsFilter) 1 ProtocolFactory in every bolt which filters URLs (URL filter, fetcher, parsers). Of course, if RobotsFilter is configured `fromCacheOnly`, the protocol instances are only used to access the static robots.txt cache.
  - every ProtocolFactory holds (usually) 2 protocol instances (http and https)

This may result in 20 or more protocol instances per worker depending on the complexity and parallelism of the topology. It would get worse if more protocol instances are kept per factory. Make ProtocolFactory a singleton? Efficiently utilizing a connection pool requires that the same pool is used for all requests which may potentially share a connection.

jnioche:
> But for the connection pools it's required that the same pool is used for the same host (implemented via hash partitioning). Do you want to change this for the Selenium/Playwright protocols? Would make sense in case browser caches aren't shared between protocol instances.

yes, I can't think of a reason not to hash for those as well
 
> While looking into this, I became aware of one difference between Nutch and StormCrawler:
> 
> * Nutch has exactly one protocol instance (because it's a plugin and every plugin is instantiated only once)
> * StormCrawler has
>   
>   * 1 ProtocolFactory per FetcherBolt and
>   * (if URL filters include RobotsFilter) 1 ProtocolFactory in every bolt which filters URLs (URL filter, fetcher, parsers). Of course, if RobotsFilter is configured `fromCacheOnly`, the protocol instances are only used to access the static robots.txt cache.
>   * every ProtocolFactory holds (usually) 2 protocol instances (http and https)
> 
> This may result in 20 or more protocol instances per worker depending on the complexity and parallelism of the topology. It would get worse if more protocol instances are kept per factory. Make ProtocolFactory a singleton? Efficiently utilizing a connection pool requires that the same pool is used for all requests which may potentially share a connection.

Totally! Having a singleton would make sense.

sebastian-nagel:
Hi @jnioche, so
- I'll shrink this PR to only make OkHttp's connection pool configurable (max. idle connections and keep-alive time), removing the multiple pool/client instances
- remaining tasks (singleton factory and "pooled" protocol instances) are moved to separate issues/PRs

Ok?

jnioche:
Yes, makes sense. Thanks. I will have a go at the pooled protocol instances.

sebastian-nagel:
Ok, simplified the config as discussed - a single but configurable connection pool is now used. Rebased to master.

core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
```diff
@@ -230,6 +231,19 @@ public EventListener create(Call call) {
         // enable support for Brotli compression (Content-Encoding)
         builder.addInterceptor(BrotliInterceptor.INSTANCE);
 
+        Map<String, Object> connectionPoolConf =
+                (Map<String, Object>) conf.get("http.protocol.connection.pool");
```
jnioche, line 235: change the config key to 'okhttp.protocol.connection.pool' to make it obvious that it applies only to it?

jnioche:
thanks @sebastian-nagel, see my very minor comment about the config key. +1 to merge whichever way you choose to handle it

sebastian-nagel:
Done: renamed config property, also fixed the log message.

jnioche:
Thanks @sebastian-nagel

PR #919, opened 2021-10-29
OkHttp protocol: add support for Brotli compression (Content-Encoding)

sebastian-nagel:
enables OkHttp's interceptor to handle [Brotli](https://en.wikipedia.org/wiki/Brotli) Content-Encoding


jnioche:
thanks @sebastian-nagel

PR #920, opened 2021-11-01
Upgrade to Tika 2.1.0

sebastian-nagel:
based on "tika-parsers-standard-package" (no extended and scientific parsers)

Open questions:
- add a tika-config.xml and disable OCR on images? Tika 2.x enables OCR by default Which may slow down the tika parser significantly in case Tesseract is installed on the system, see https://cwiki.apache.org/confluence/display/tika/tikaocr

See also:
  https://cwiki.apache.org/confluence/display/TIKA/Migrating+to+Tika+2.0.0

### DIFF
--- a/archetype/src/main/resources/archetype-resources/crawler-conf.yaml
+++ b/archetype/src/main/resources/archetype-resources/crawler-conf.yaml
@@ -90,6 +90,9 @@ config:
    - application/.+powerpoint.*
    - application/.*pdf.*
 
+  # Tika parser configuration file
+  parse.tika.config.file: "tika-config.xml"
+
   # custom fetch interval to be used when a document has the key/value in its metadata
   # and has been fetched successfully (value in minutes)
   # fetchInterval.FETCH_ERROR.isFeed=true: 30
--- a/core/src/main/java/com/digitalpebble/stormcrawler/bolt/JSoupParserBolt.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/bolt/JSoupParserBolt.java
@@ -42,6 +42,7 @@ import org.apache.storm.tuple.Tuple;
 import org.apache.storm.tuple.Values;
 import org.apache.tika.config.TikaConfig;
 import org.apache.tika.detect.Detector;
+import org.apache.tika.metadata.TikaCoreProperties;
 import org.apache.tika.mime.MediaType;
 import org.jsoup.nodes.Element;
 import org.jsoup.parser.Parser;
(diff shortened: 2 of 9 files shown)


jnioche:
thanks @sebastian-nagel 
Yes, I think OCR should be disabled by default

sebastian-nagel:
Added a configuration file disabling OCR.

archetype/src/main/resources/archetype-resources/crawler-conf.yaml
```diff
@@ -90,6 +90,9 @@ config:
    - application/.+powerpoint.*
    - application/.*pdf.*
 
+  # Tika parser configuration file
+  parse.tika.config.file: "tika-config.xml"
```
jnioche, line 94: This will be changed so rarely that I wonder whether it should be put in the default config instead. This would help the generated crawler-conf.yaml as minimal as possible.

jnioche (approved):
Looks great thanks @sebastian-nagel 
Just one minor comment
Did you use it on a crawl?

archetype/src/main/resources/archetype-resources/crawler-conf.yaml
sebastian-nagel: Yes, of course. My thought was that a configuration used in and deployed via a module shouldn't be defined in core.

jnioche: I had exactly the same thought. I don't mind too much, let's leave it where it is. 
Did you run a crawl with it?

sebastian-nagel:
> Did you use it on a crawl?

I used [warc-crawler](//github.com/sebastian-nagel/warc-crawler) but only looked whether the configuration is loaded and parsing generally works. But I'll compare it with a run with SC 1.18 (Tika 1.26) on the same input to look for regressions.

jnioche:
Fab, thanks!

sebastian-nagel:
Results from 450 documents parsed by the Tika parser bolt comparing SC 1.18 / Tika 1.26 and SC 2.2-SNAPSHOT / Tika 2.1.0 :
- 2 documents not successfully parsed with 2.1.0:
  - a Dublin Core XML file: [asked for clarification on the Tika user mailing list](https://lists.apache.org/thread/1p0jld73jm7q7dxk29whhowgxovvctog)
  - a [application/x-hdf](https://en.wikipedia.org/wiki/Hierarchical_Data_Format) file: expected as the parser is not included in the standards parser package
- the parsefilters required a [tweak](https://github.com/sebastian-nagel/warc-crawler/commit/1790689b4c4339c8a71525642ba9022682a1f784) to include `dc:description` and `dc:subject` due to changes in the Tika metadata, see the [Tika 2.x migration guide](https://cwiki.apache.org/confluence/display/TIKA/Migrating+to+Tika+2.0.0)

jnioche:
Perfect thanks @sebastian-nagel

PR #922, opened 2021-11-08
Allow to set http.content.limit per page in metadata

### LINKED ISSUE #921: http.content.limit has effect on robots.txt
Since we need only some header information in our project we set »http.content.limit: 0« and we just realized that this setting concerns also the robots.txt, so that f.e. a crawl-delay set there isn't respected anymore. 

If it's not a bug, it might be reasonable to point out the effect of the setting on the processing of robots.txt

sebastian-nagel:
Add property `http.robots.content.limit` to configure the robots.txt content limit independent from `http.content.limit`
- fixes #921 "http.content.limit has effect on robots.txt"
- default is 512 kiB to follow the [robots.txt RFC draft](https://datatracker.ietf.org/doc/html/draft-rep-wg-topic-00#section-2.5)

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/protocol/HttpRobotRulesParser.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/protocol/HttpRobotRulesParser.java
@@ -41,6 +41,8 @@ public class HttpRobotRulesParser extends RobotRulesParser {
 
     protected boolean allowForbidden = false;
 
+    protected Metadata fetchRobotsMd;
+
     HttpRobotRulesParser() {
     }
 
@@ -53,6 +55,12 @@ public class HttpRobotRulesParser extends RobotRulesParser {
         super.setConf(conf);
         allowForbidden = ConfUtils.getBoolean(conf, "http.robots.403.allow",
                 true);
+        fetchRobotsMd = new Metadata();
+        /* http.content.limit for fetching the robots.txt */
+        int robotsTxtContentLimit = ConfUtils.getInt(conf,
+                "http.robots.content.limit", 524288);
+        fetchRobotsMd.addValue("http.content.limit",
+                Integer.toString(robotsTxtContentLimit));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/digitalpebble/stormcrawler/protocol/HttpRobotRulesParser.java
```diff
@@ -53,6 +55,12 @@ public void setConf(Config conf) {
         super.setConf(conf);
         allowForbidden = ConfUtils.getBoolean(conf, "http.robots.403.allow",
                 true);
+        fetchRobotsMd = new Metadata();
+        /* http.content.limit for fetching the robots.txt */
+        int robotsTxtContentLimit = ConfUtils.getInt(conf,
+                "http.robots.content.limit", 524288);
```
jnioche, line 61: default value should match what we have for http.content.limit and the default conf file => -1

core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
```diff
@@ -392,7 +403,7 @@ public ProtocolResponse getProtocolOutput(String url, final Metadata metadata) t
     }
 
     private final byte[] toByteArray(final ResponseBody responseBody,
-            MutableObject trimmed) throws IOException {
+            int maxContent, MutableObject trimmed) throws IOException {
```
jnioche, line 406: there is a field _maxContent_ already, which stores the value to be applied in the absence of a per-URL override. To avoid confusion, it would be better to rename that field into something like _globalMaxContent_.

core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
```diff
@@ -306,6 +306,8 @@ public ProtocolResponse getProtocolOutput(String url, final Metadata metadata) t
             rb.header(k.getKey(), k.getValue());
         });
 
+        int pageMaxContent = maxContent;
```
jnioche, line 309: see comment below about renaming the field maxContent for more clarity

jnioche (commented):
thanks @sebastian-nagel, this is an elegant approach. See comments

sebastian-nagel:
Hi @jnioche, addressed your comments and rebased to recent master.

jnioche:
That's great. Thanks @sebastian-nagel

Issue #485, opened 2017-07-11
Configure HTTP protocol to use HEAD method only


jnioche:
This can be done on a per URL basis for cases where we only want to know the HTTP headers returned by the server without getting the actual content e.g. is this URL redirected and of so where to?

This can be done on a per URL basis by giving a tuple an arbitrary key/value such as `http.method.head`

jnioche:
Add this to #486

PR #923, opened 2021-11-08
Support HEAD method in okhttp protocol

sebastian-nagel:
- send HEAD request if `http.method.head` is in page metadata and is true
- ports the behavior of #485 to the okhttp-based protocol implementation
- fix deprecation warning in POST method


jnioche:
thanks @sebastian-nagel

Issue #925, opened 2021-11-16
JSoupParserBolt cannot configure more than one JSoupFilters per worker


Mikwiss:
Hi all !

I'm playing with the apache storm parallelism. With the following configuration, I encounter an issue : 

Conf yaml :
```
config:
    topology.workers: 1
```

Flux file :
```
-  id: "parser"
   className: "com.digitalpebble.stormcrawler.bolt.JSoupParserBolt"
   parallelism: 2
```

In this case, only the first JSoupParserBolt instantiated has a not empty list of JSoupFilters. With some investigation, I found a variable should be not static : 

`    public static String configFile = "jsoup.filters.config.file"; `

With the keyword static, all the bolt on the same worker share the variable. After the first instantiation, the JouspFilters looses the key config value _jsoup.filters.config.file_. 

JsoupFilters line 93 : 
```
public JSoupFilters(Map stormConf, String configFile) throws IOException {
		this.configFile = configFile;
}
```

I'll run some test and I'll make a pull request !

PR #927, opened 2021-11-17
Remove static keyword on JSoupFilters field

Mikwiss:
Fixed issue #925 

Removed the static field into JSoupFilters. Also, removed the never used _configFile_ value into ParserFilters.


jnioche:
I can see what happened here. There was some confusion between static fields used for the name of the config key and the use of the same field to store the value associated with them i.e. the actual name of the file.
Good catch! thanks for reporting AND fixing @Mikwiss !

PR #936, opened 2021-12-21
Upgrade to log4j 2.17.0, fixes #935

### LINKED ISSUE #935: Log4J vulnerability - CVE-2021-44228
What kind of issue is this?

 - [] Question. This issue tracker is not the best place for questions. If you want to ask how to do
       something, or to understand why something isn't working the way you expect it to, use StackOverflow
       instead with the label 'stormcrawler': https://stackoverflow.com/questions/tagged/stormcrawler 

 - [ ] Bug report. If you’ve found a bug, please include a test if you can, it makes it a lot easier to fix things. Use the label 'bug' on the issue.
 
 - [x] Feature request. Please use the label 'wish' on the issue.

Thanks!

Adding this issue, in order to let you know that the stormcrawler project is being affected by the vulnerability documented here
https://nvd.nist.gov/vuln/detail/CVE-2021-44228 related with log4j.

image

```
 ✘ jmartini@cacute-xps  ~/workspace/storm-crawler-fork   master  mvn dependency:tree | grep log4j
[INFO] |  +- org.apache.logging.log4j:log4j-api:jar:2.11.2:provided
[INFO] |  +- org.apache.logging.log4j:log4j-core:jar:2.11.2:provided

```
we can create a PR for fixing this on our end , overriding the dependencies that are coming from apache storm dependency

sebastian-nagel:
Enforce usage of 2.17.0 as minimum log4j-core version in all modules. If storm-client is upgraded and depends on a safe log4j version, we can remove the log4j dependency from sc.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -96,6 +96,22 @@
 			<groupId>org.apache.storm</groupId>
 			<artifactId>storm-client</artifactId>
 		</dependency>
+		<!-- dependency overrides for vulnerable log4j versions -->
+		<dependency>
+			<groupId>org.apache.logging.log4j</groupId>
+			<artifactId>log4j-core</artifactId>
+			<version>${log4j2.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.logging.log4j</groupId>
+			<artifactId>log4j-api</artifactId>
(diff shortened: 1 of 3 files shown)


jnioche:
thanks @sebastian-nagel

PR #934, opened 2021-12-11
Apply various bug-fixes, code cleaning and optimizations

FelixEngl:
Hello everyone,

i wrote some bugfixes and did some code cleaning + code optimizing.
For an easier overview I attached a markdown file to this pull request listing all relevant changes.  (see [ChangelogStormcrawler.md](https://github.com/DigitalPebble/storm-crawler/files/7697090/ChangelogStormcrawler.md))

Signed-off-by: Felix Engl <felix.engl@hotmail.com>

Developer Certificate of Origin
===============================

By contributing to StormCrawler, you accept and agree to the following terms and conditions (the *Developer Certificate of Origin*) for your present and future contributions submitted to StormCrawler. 
Please refer to the *Developer Certificate of Origin* section in [`CONTRIBUTING.md`](CONTRIBUTING.md) for details.

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

- [x] You've squashed your commits into a single 
- [x] You've described what the PR does or at least point to a related issue
- [x] You've signed-ff your commits with 'git commit -s'

PR #937, opened 2021-12-21
General Code Refactoring and Good Practices

FelixEngl:
As described in #934: A collection of contract improvements and interface enhancements. They don't change anything of the program logic and are mostly for stability, performance, contract and/or documentation reasons.

Changes a lot of files due to the nature of the modified interfaces/abstract classes.

Signed-off-by: Felix Engl <felix.engl@hotmail.com>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

Thanks!

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/Metadata.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/Metadata.java
@@ -159,15 +159,15 @@ public class Metadata {
         if (values == null || values.size() == 0) return;
         String[] existingvals = md.get(key);
         if (existingvals == null) {
-            md.put(key, values.toArray(new String[values.size()]));
+            md.put(key, values.toArray(new String[0]));
             return;
         }
 
         ArrayList<String> existing = new ArrayList<>(existingvals.length + values.size());
-        for (String v : existingvals) existing.add(v);
+        Collections.addAll(existing, existingvals);
 
         existing.addAll(values);
-        md.put(key, existing.toArray(new String[existing.size()]));
+        md.put(key, existing.toArray(new String[0]));
     }
 
     /** @return the previous value(s) associated with <tt>key</tt> */
@@ -243,7 +243,7 @@ public class Metadata {
     }
 
     /** @since 1.16 */
-    private final void checkLockException() {
+    private void checkLockException() {
         if (locked)
             throw new ConcurrentModificationException(
                     "Attempt to modify a metadata after it has been sent to the serializer");
--- a/core/src/main/java/com/digitalpebble/stormcrawler/bolt/FeedParserBolt.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/bolt/FeedParserBolt.java
@@ -50,7 +50,6 @@ import org.slf4j.LoggerFactory;
 import org.xml.sax.InputSource;
 
 /** Extracts URLs from feeds */
-@SuppressWarnings("serial")
 public class FeedParserBolt extends StatusEmitterBolt {
 
     public static final String isFeedKey = "isFeed";
@@ -72,7 +71,7 @@ public class FeedParserBolt extends StatusEmitterBolt {
 
         LOG.debug("Processing {}", url);
 
-        boolean isfeed = Boolean.valueOf(metadata.getFirstValue(isFeedKey));
+        boolean isfeed = Boolean.parseBoolean(metadata.getFirstValue(isFeedKey));
         // doesn't have the metadata expected
         if (!isfeed) {
             if (sniffWhenNoMDKey) {
@@ -163,7 +162,7 @@ public class FeedParserBolt extends StatusEmitterBolt {
             throws Exception {
         List<Outlink> links = new ArrayList<>();
 
-        SyndFeed feed = null;
+        SyndFeed feed;
         try (ByteArrayInputStream is = new ByteArrayInputStream(content)) {
             SyndFeedInput input = new SyndFeedInput();
             feed = input.build(new InputSource(is));
@@ -200,7 +199,7 @@ public class FeedParserBolt extends StatusEmitterBolt {
                         LOG.info(
                                 "{} has a published date {} which is more than {} hours old",
                                 targetURL,
-                                publishedDate.toString(),
+                                publishedDate,
                                 filterHoursSincePub);
                         continue;
                     }
@@ -220,8 +219,8 @@ public class FeedParserBolt extends StatusEmitterBolt {
     }
 
     @Override
-    @SuppressWarnings({"rawtypes", "unchecked"})
-    public void prepare(Map stormConf, TopologyContext context, OutputCollector collect) {
+    public void prepare(
+            Map<String, Object> stormConf, TopologyContext context, OutputCollector collect) {
         super.prepare(stormConf, context, collect);
         sniffWhenNoMDKey = ConfUtils.getBoolean(stormConf, "feed.sniffContent", false);
         filterHoursSincePub = ConfUtils.getInt(stormConf, "feed.filter.hours.since.published", -1);
--- a/core/src/main/java/com/digitalpebble/stormcrawler/bolt/FetcherBolt.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/bolt/FetcherBolt.java
@@ -64,7 +64,6 @@ import org.slf4j.LoggerFactory;
  * A multithreaded, queue-based fetcher adapted from Apache Nutch. Enforces the politeness and
  * handles the fetching threads itself.
  */
-@SuppressWarnings("serial")
 public class FetcherBolt extends StatusEmitterBolt {
 
     private static final org.slf4j.Logger LOG = LoggerFactory.getLogger(FetcherBolt.class);
@@ -236,8 +235,8 @@ public class FetcherBolt extends StatusEmitterBolt {
      * provides items eligible for fetching from any queue.
      */
     private static class FetchItemQueues {
-        Map<String, FetchItemQueue> queues =
-                Collections.synchronizedMap(new LinkedHashMap<String, FetchItemQueue>());
+        final Map<String, FetchItemQueue> queues =
+                Collections.synchronizedMap(new LinkedHashMap<>());
 
         AtomicInteger inQueues = new AtomicInteger(0);
 
@@ -318,7 +317,7 @@ public class FetcherBolt extends StatusEmitterBolt {
                 // custom maxThread value?
                 for (Entry<Pattern, Integer> p : customMaxThreads.entrySet()) {
                     if (p.getKey().matcher(id).matches()) {
-                        customThreadVal = p.getValue().intValue();
+                        customThreadVal = p.getValue();
                         break;
                     }
                 }
@@ -393,7 +392,7 @@ public class FetcherBolt extends StatusEmitterBolt {
         // whether the default delay is used even if the robots.txt
         // specifies a shorter crawl-delay
         private final boolean crawlDelayForce;
-        private int threadNum;
+        private final int threadNum;
 
         private long timeoutInQueues = -1;
 
@@ -404,7 +403,7 @@ public class FetcherBolt extends StatusEmitterBolt {
             this.setDaemon(true); // don't hang JVM on exit
             this.setName("FetcherThread #" + num); // use an informative name
 
-            this.maxCrawlDelay = ConfUtils.getInt(conf, "fetcher.max.crawl.delay", 30) * 1000;
+            this.maxCrawlDelay = ConfUtils.getInt(conf, "fetcher.max.crawl.delay", 30) * 1000L;
             this.maxCrawlDelayForce =
                     ConfUtils.getBoolean(conf, "fetcher.max.crawl.delay.force", false);
             this.crawlDelayForce = ConfUtils.getBoolean(conf, "fetcher.server.delay.force", false);
@@ -486,13 +485,16 @@ public class FetcherBolt extends StatusEmitterBolt {
 
                     // check in the metadata if discovery setting has been
                     // overridden
-                    boolean smautodisco = sitemapsAutoDiscovery;
+
                     String localSitemapDiscoveryVal =
                             metadata.getFirstValue(SITEMAP_DISCOVERY_PARAM_KEY);
-                    if ("true".equalsIgnoreCase(localSitemapDiscoveryVal)) {
+                    boolean smautodisco;
+                    if (Boolean.parseBoolean(localSitemapDiscoveryVal)) {
                         smautodisco = true;
                     } else if ("false".equalsIgnoreCase(localSitemapDiscoveryVal)) {
                         smautodisco = false;
+                    } else {
+                        smautodisco = sitemapsAutoDiscovery;
                     }
 
                     if (!fromCache && smautodisco) {
@@ -746,9 +748,9 @@ public class FetcherBolt extends StatusEmitterBolt {
         }
     }
 
-    @SuppressWarnings({"rawtypes", "unchecked"})
     @Override
-    public void prepare(Map stormConf, TopologyContext context, OutputCollector collector) {
+    public void prepare(
+            Map<String, Object> stormConf, TopologyContext context, OutputCollector collector) {
 
         super.prepare(stormConf, context, collector);
 
@@ -823,7 +825,7 @@ public class FetcherBolt extends StatusEmitterBolt {
 
         maxNumberURLsInQueues = ConfUtils.getInt(conf, "fetcher.max.urls.in.queues", -1);
 
-        /**
+        /*
          * If set to a valid path e.g. /tmp/fetcher-dump-{port} on a worker node, the content of the
          * queues will be dumped to the logs for debugging. The port number needs to match the one
          * used by the FetcherBolt instance.
--- a/core/src/main/java/com/digitalpebble/stormcrawler/bolt/FetcherBolt.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/bolt/FetcherBolt.java
@@ -488,8 +488,10 @@ public class FetcherBolt extends StatusEmitterBolt {
 
                     String localSitemapDiscoveryVal =
                             metadata.getFirstValue(SITEMAP_DISCOVERY_PARAM_KEY);
+
                     boolean smautodisco;
-                    if (Boolean.parseBoolean(localSitemapDiscoveryVal)) {
+
+                    if ("true".equalsIgnoreCase(localSitemapDiscoveryVal)) {
                         smautodisco = true;
                     } else if ("false".equalsIgnoreCase(localSitemapDiscoveryVal)) {
                         smautodisco = false;
--- a/core/src/main/java/com/digitalpebble/stormcrawler/bolt/JSoupParserBolt.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/bolt/JSoupParserBolt.java
@@ -68,7 +68,6 @@ import org.w3c.dom.DocumentFragment;
  * Parser for HTML documents only which uses ICU4J to detect the charset encoding. Kindly donated to
  * storm-crawler by shopstyle.com.
  */
-@SuppressWarnings("serial")
 public class JSoupParserBolt extends StatusEmitterBolt {
 
     /** Metadata key name for tracking the anchors */
@@ -82,7 +81,7 @@ public class JSoupParserBolt extends StatusEmitterBolt {
 
     private JSoupFilter jsoupFilters = null;
 
-    private Detector detector = TikaConfig.getDefaultConfig().getDetector();
+    private final Detector detector = TikaConfig.getDefaultConfig().getDetector();
 
     private boolean detectMimeType = true;
 
@@ -119,9 +118,9 @@ public class JSoupParserBolt extends StatusEmitterBolt {
 
     private boolean ignoreMetaRedirections;
 
-    @SuppressWarnings({"rawtypes", "unchecked"})
     @Override
-    public void prepare(Map conf, TopologyContext context, OutputCollector collector) {
+    public void prepare(
+            Map<String, Object> conf, TopologyContext context, OutputCollector collector) {
 
         super.prepare(conf, context, collector);
 
@@ -237,7 +236,7 @@ public class JSoupParserBolt extends StatusEmitterBolt {
         }
 
         Map<String, List<String>> slinks;
-        String text = "";
+        String text;
         final org.jsoup.nodes.Document jsoupDoc;
 
         try {
@@ -301,9 +300,7 @@ public class JSoupParserBolt extends StatusEmitterBolt {
             }
 
             Element body = jsoupDoc.body();
-            if (body != null) {
-                text = textExtractor.text(body);
-            }
+            text = textExtractor.text(body);
 
         } catch (Throwable e) {
             String errorMessage = "Exception while parsing " + url + ": " + e;
@@ -489,7 +486,7 @@ public class JSoupParserBolt extends StatusEmitterBolt {
             // the URL is valid
             LOG.error("MalformedURLException on {}", url);
             eventCounter.scope("error_invalid_source_url").incrBy(1);
-            return new LinkedList<Outlink>();
+            return new LinkedList<>();
         }
 
         for (Map.Entry<String, List<String>> linkEntry : slinks.entrySet()) {
@@ -519,6 +516,6 @@ public class JSoupParserBolt extends StatusEmitterBolt {
             }
         }
 
-        return new LinkedList<Outlink>(outlinks.values());
+        return new LinkedList<>(outlinks.values());
     }
 }
--- a/core/src/main/java/com/digitalpebble/stormcrawler/bolt/SiteMapParserBolt.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/bolt/SiteMapParserBolt.java
@@ -64,7 +64,6 @@ import org.slf4j.LoggerFactory;
  * stream, whereas any URLs extracted from the sitemaps are sent to the 'status' field with a
  * 'DISCOVERED' status.
  */
-@SuppressWarnings("serial")
 public class SiteMapParserBolt extends StatusEmitterBolt {
 
     public static final String isSitemapKey = "isSitemap";
@@ -106,14 +105,10 @@ public class SiteMapParserBolt extends StatusEmitterBolt {
 
         String isSitemap = metadata.getFirstValue(isSitemapKey);
 
-        boolean treatAsSM = false;
-
-        if ("true".equalsIgnoreCase(isSitemap)) {
-            treatAsSM = true;
-        }
+        boolean treatAsSM = Boolean.parseBoolean(isSitemap);
 
         // doesn't have the key and want to rely on the clue
-        else if (isSitemap == null && looksLikeSitemap) {
+        if (isSitemap == null && looksLikeSitemap) {
             LOG.info("{} detected as sitemap based on content", url);
             treatAsSM = true;
         }
@@ -222,7 +217,7 @@ public class SiteMapParserBolt extends StatusEmitterBolt {
                             LOG.info(
                                     "{} has a modified date {} which is more than {} hours old",
                                     target,
-                                    lastModified.toString(),
+                                    lastModified,
                                     filterHoursSinceModified);
                             continue;
                         }
@@ -337,8 +332,8 @@ public class SiteMapParserBolt extends StatusEmitterBolt {
     }
 
     @Override
-    @SuppressWarnings({"rawtypes", "unchecked"})
-    public void prepare(Map stormConf, TopologyContext context, OutputCollector collector) {
+    public void prepare(
+            Map<String, Object> stormConf, TopologyContext context, OutputCollector collector) {
         super.prepare(stormConf, context, collector);
         parser = new SiteMapParser(false);
         filterHoursSinceModified =
@@ -373,15 +368,12 @@ public class SiteMapParserBolt extends StatusEmitterBolt {
      * Examines the first bytes of the content for a clue of whether this document is a sitemap,
      * based on namespaces. Works for XML and non-compressed documents only.
      */
-    private final boolean sniff(byte[] content) {
+    private boolean sniff(byte[] content) {
         byte[] beginning = content;
         if (content.length > maxOffsetGuess && maxOffsetGuess > 0) {
             beginning = Arrays.copyOfRange(content, 0, maxOffsetGuess);
         }
         int position = Bytes.indexOf(beginning, clue);
-        if (position != -1) {
-            return true;
-        }
-        return false;
+        return position != -1;
     }
 }
--- a/core/src/main/java/com/digitalpebble/stormcrawler/filtering/URLFilter.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/filtering/URLFilter.java
@@ -17,6 +17,8 @@ package com.digitalpebble.stormcrawler.filtering;
 import com.digitalpebble.stormcrawler.Metadata;
 import com.digitalpebble.stormcrawler.util.Configurable;
 import java.net.URL;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * Unlike Nutch, URLFilters can normalise the URLs as well as filtering them. URLFilter instances
@@ -31,6 +33,12 @@ public interface URLFilter extends Configurable {
      * @param sourceUrl the URL of the page where the URL was found. Can be null.
      * @param sourceMetadata the metadata collected for the page
      * @param urlToFilter the URL to be filtered
+     * @return null if the url is to be removed or a normalised representation which can correspond
+     *     to the input URL
      */
-    public String filter(URL sourceUrl, Metadata sourceMetadata, String urlToFilter);
+    @Nullable
+    String filter(
+            @Nullable URL sourceUrl,
+            @Nullable Metadata sourceMetadata,
+            @NotNull String urlToFilter);
 }
--- a/core/src/main/java/com/digitalpebble/stormcrawler/filtering/sitemap/SitemapFilter.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/filtering/sitemap/SitemapFilter.java
@@ -18,6 +18,8 @@ import com.digitalpebble.stormcrawler.Metadata;
 import com.digitalpebble.stormcrawler.bolt.SiteMapParserBolt;
 import com.digitalpebble.stormcrawler.filtering.URLFilter;
 import java.net.URL;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * URLFilter which discards URLs discovered in a page which is not a sitemap when sitemaps have been
@@ -40,15 +42,17 @@ import java.net.URL;
 public class SitemapFilter implements URLFilter {
 
     @Override
-    public String filter(URL sourceUrl, Metadata sourceMetadata, String urlToFilter) {
-        if (sourceMetadata != null) {
-            if ("false"
-                            .equalsIgnoreCase(
-                                    sourceMetadata.getFirstValue(SiteMapParserBolt.isSitemapKey))
-                    && "true"
-                            .equalsIgnoreCase(
-                                    sourceMetadata.getFirstValue(
-                                            SiteMapParserBolt.foundSitemapKey))) return null;
+    public @Nullable String filter(
+            @Nullable URL sourceUrl,
+            @Nullable Metadata sourceMetadata,
+            @NotNull String urlToFilter) {
+
+        if (sourceMetadata != null
+                && !Boolean.parseBoolean(
+                        sourceMetadata.getFirstValue(SiteMapParserBolt.isSitemapKey))
+                && Boolean.parseBoolean(
+                        sourceMetadata.getFirstValue(SiteMapParserBolt.foundSitemapKey))) {
+            return null;
         }
         return urlToFilter;
     }
--- a/core/src/main/java/com/digitalpebble/stormcrawler/parse/JSoupFilter.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/parse/JSoupFilter.java
@@ -15,21 +15,26 @@
 package com.digitalpebble.stormcrawler.parse;
 
 import com.digitalpebble.stormcrawler.util.Configurable;
+import org.jetbrains.annotations.NotNull;
 
 /**
  * Implementations of ParseFilter are responsible for extracting custom data from the crawled
  * content. They are used exclusively by {@link
  * com.digitalpebble.stormcrawler.bolt.JSoupParserBolt}.
  */
-public interface JSoupFilter extends Configurable {
+public abstract class JSoupFilter implements Configurable {
 
     /**
      * Called when parsing a specific page
      *
-     * @param URL the URL of the page being parsed
+     * @param url the URL of the page being parsed
      * @param content the content being parsed
      * @param doc document produced by JSoup's parsingF
      * @param parse the metadata to be updated with the resulting of the parsing
      */
-    void filter(String URL, byte[] content, org.jsoup.nodes.Document doc, ParseResult parse);
+    public abstract void filter(
+            @NotNull String url,
+            byte[] content,
+            @NotNull org.jsoup.nodes.Document doc,
+            @NotNull ParseResult parse);
 }
--- a/core/src/main/java/com/digitalpebble/stormcrawler/parse/JSoupFilter.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/parse/JSoupFilter.java
@@ -21,7 +21,7 @@ import com.digitalpebble.stormcrawler.util.Configurable;
  * content. They are used exclusively by {@link
  * com.digitalpebble.stormcrawler.bolt.JSoupParserBolt}.
  */
-public abstract class JSoupFilter implements Configurable {
+public interface JSoupFilter extends Configurable {
 
     /**
      * Called when parsing a specific page
@@ -31,6 +31,5 @@ public abstract class JSoupFilter implements Configurable {
      * @param doc document produced by JSoup's parsingF
      * @param parse the metadata to be updated with the resulting of the parsing
      */
-    public abstract void filter(
-            String URL, byte[] content, org.jsoup.nodes.Document doc, ParseResult parse);
+    void filter(String URL, byte[] content, org.jsoup.nodes.Document doc, ParseResult parse);
 }
--- a/core/src/main/java/com/digitalpebble/stormcrawler/persistence/AdaptiveScheduler.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/persistence/AdaptiveScheduler.java
@@ -145,9 +145,7 @@ public class AdaptiveScheduler extends DefaultScheduler {
 
     /**
      * Name of key to hold previous signature: a copy, not overwritten by {@link
-     * MD5SignatureParseFilter}, is added by {@link
-     * com.digitalpebble.stormcrawler.parse.filter.SignatureCopyParseFilter} . This key is a
-     * temporary copy, not necessarily persisted in metadata.
+     * MD5SignatureParseFilter}.
      */
     public static final String SIGNATURE_OLD_KEY = "signatureOld";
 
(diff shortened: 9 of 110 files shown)


jnioche:
why remove the _@SuppressWarnings("serial")?_

core/src/main/java/com/digitalpebble/stormcrawler/bolt/JSoupParserBolt.java
```diff
@@ -301,9 +300,7 @@ public void execute(Tuple tuple) {
             }
 
             Element body = jsoupDoc.body();
-            if (body != null) {
```
jnioche, line 304: what guarantee do you have that the _body_ returned is not null?

FelixEngl:
Hello @jnioche,

i hope you had a merry christmas and a good start into the new year.

> why remove the _@SuppressWarnings("serial")?_

IComponent implements Serializeable, therefore suppressing that warning is a violation to the Serializeable contract.

core/src/main/java/com/digitalpebble/stormcrawler/bolt/JSoupParserBolt.java
FelixEngl: The documentation of Document.body() states, that it'll create a body element if none is existing.

https://jsoup.org/apidocs/org/jsoup/nodes/Document.html#body()

jnioche:
Hi @FelixEngl 

> i hope you had a merry christmas and a good start into the new year.

Thanks! I hope you had a  good Christmas too and wish you a very happy new year!

More questions / comments. Not finished by all means...

core/src/main/java/com/digitalpebble/stormcrawler/bolt/FeedParserBolt.java
```diff
@@ -72,7 +71,7 @@ public void execute(Tuple tuple) {
 
         LOG.debug("Processing {}", url);
 
-        boolean isfeed = Boolean.valueOf(metadata.getFirstValue(isFeedKey));
+        boolean isfeed = Boolean.parseBoolean(metadata.getFirstValue(isFeedKey));
```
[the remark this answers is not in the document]
FelixEngl: Yes, that is a redundant boxing.

core/src/main/java/com/digitalpebble/stormcrawler/bolt/FetcherBolt.java
```diff
@@ -318,7 +317,7 @@ public synchronized FetchItemQueue getFetchItemQueue(String id) {
                 // custom maxThread value?
                 for (Entry<Pattern, Integer> p : customMaxThreads.entrySet()) {
                     if (p.getKey().matcher(id).matches()) {
-                        customThreadVal = p.getValue().intValue();
+                        customThreadVal = p.getValue();
```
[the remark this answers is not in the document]
FelixEngl: Because that is an uneccessary unboxing. That is done automatically by java.
https://docs.oracle.com/javase/tutorial/java/data/autoboxing.html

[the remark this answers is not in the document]
FelixEngl: The difference between the situations is, that we can not store a primitive in a Map, therefore we need that `Integer`-object and the JVM calls the `intValue` automatically to unbox, but here we would only call `parseBoolean` and wrap it with an additional If-Else-Statement to return one of the `Boolean`-constants followed by a call of `booleanValue` to get the value of the `Boolean`-instance for our `boolean` variable.

```java
// Code from Boolean

public static boolean parseBoolean(String s) {
        return "true".equalsIgnoreCase(s);
}

public static final Boolean TRUE = new Boolean(true);

public static final Boolean FALSE = new Boolean(false);

public static Boolean valueOf(boolean b) {
    return (b ? TRUE : FALSE);
}
```

core/src/main/java/com/digitalpebble/stormcrawler/bolt/FetcherBolt.java
```diff
@@ -486,13 +485,16 @@ public void run() {
 
                     // check in the metadata if discovery setting has been
                     // overridden
-                    boolean smautodisco = sitemapsAutoDiscovery;
+
                     String localSitemapDiscoveryVal =
                             metadata.getFirstValue(SITEMAP_DISCOVERY_PARAM_KEY);
-                    if ("true".equalsIgnoreCase(localSitemapDiscoveryVal)) {
+                    boolean smautodisco;
+                    if (Boolean.parseBoolean(localSitemapDiscoveryVal)) {
```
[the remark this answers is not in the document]
FelixEngl: That is true, this is just a personal preference of mine. But when i did this i was so `in the zone` that i just did whatever i found. ^^'

core/src/main/java/com/digitalpebble/stormcrawler/bolt/JSoupParserBolt.java
```diff
@@ -489,7 +486,7 @@ public String guessMimeType(String URL, String httpCT, byte[] content) {
             // the URL is valid
             LOG.error("MalformedURLException on {}", url);
             eventCounter.scope("error_invalid_source_url").incrBy(1);
-            return new LinkedList<Outlink>();
+            return new LinkedList<>();
```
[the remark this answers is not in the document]
FelixEngl: I left out the explicit type argument of the LinkedList because the return type is already defined as `List<Outlink>`, therefore the compiler infers the parametertype from the signature via the diamond operator <>. It works basically the same as declaring a list like that: `List<String> myList = new ArrayList<>()`.

The other reason is, that the explicit type argument of a generic is resolved during compile time and then erased. Therefore the type argument of a generic doesn't add anything to the class itself during runtime. 
That's the same reason for the fact that you could return a raw type like `new LinkedList()` and it would still compile and work the same as returning `LinkedList<>` or `LinkedList<Outlink>` as long as your return type is `List<Outlink>`. Basically every generic class that you use, loses most\* of it's information about its generic types during compile time and replaces them in the bytecode (conceptually, the real results of the compiler might differ due to optimisations and other whatnots) like this:

```java
List<String> = new ArrayList<>();
list.add("A value!");
String s = list.get(0);

List list = new ArrayList();
list.add("A value!");
String s = (String) list.get(0);
```

And the explicit declaration of generics in a signature like `Map<String, Object>` has multiple advantages over raw types:
- The compiler creates and optimizes the conversion code of the values for you and you don't have to do that by yourself. There is also the possibility, that the compiler doesn't introduce any type checking during runtime at all.
- The interface is clearer for the reader because the reader sees what's expected from the generic objects like a Map or List without knowing every inch of the code. Like in the following example, where it is easier to implement something when the interface is declared like in framework B:
```java

// Framework A
public abstract class Event {
    public abstract String eventMessage();
}

@FunctionalInterface
public interface EventConsumer {
    public void consume(List events);
}

public interface EventSource {
    public void register(EventConsumer event);
}

// Usage in code
public class NetworkListener implements EventSource {
    // This is the only Event sent by this class.
    public static class NetworkEvent extends Event {
      //...
    }

    public void register(EventConsumer event) {
      //...
    }
}

// Framework B
public abstract class Event {
    abstract String eventMessage();
}

@FunctionalInterface
public interface EventConsumer<T extends Event> {
    void consume(List<T> events);
}

public interface EventSourcee<T extends Event> {
    public void register(EventConsumer<T> event);
}

// Usage in code
public class NetworkListener implements EventSource<NetworkListener.NetworkEvent> {
     // This is the only Event sent by this class.
    public static class NetworkEvent extends Event {
        //...
    }

    public void register(EventConsumer<NetworkEvent> event) {
        //...
    }
}

```
- That you can declare raw types is a legacy of old java code prior to jdk 5.0 when there was no "generic". Therefore when you use generics you get essentially pre generic behaviour and that is sometimes really strange and not wanted or even dangerous.

Sources: 
- https://www.baeldung.com/java-type-erasure
- https://docs.oracle.com/javase/tutorial/java/generics/rawTypes.html
- https://docs.oracle.com/javase/tutorial/java/generics/erasure.html

\* I don't want to say "all" because I'm not deep enough in the "java type erasure during compilation"-process to know if java really erases all informations about the generic type from the class instance.

core/src/main/java/com/digitalpebble/stormcrawler/bolt/SiteMapParserBolt.java
```diff
@@ -373,15 +368,12 @@ public void declareOutputFields(OutputFieldsDeclarer declarer) {
      * Examines the first bytes of the content for a clue of whether this document is a sitemap,
      * based on namespaces. Works for XML and non-compressed documents only.
      */
-    private final boolean sniff(byte[] content) {
+    private boolean sniff(byte[] content) {
```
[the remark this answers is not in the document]
FelixEngl: Declaring a private method `final` is redundant because you can not override a private function and are therefore handled like being `final` functions. This is basically like declaring the methods of an public interface as public.

core/src/main/java/com/digitalpebble/stormcrawler/filtering/URLFilter.java
```diff
@@ -17,6 +17,8 @@
 import com.digitalpebble.stormcrawler.Metadata;
 import com.digitalpebble.stormcrawler.util.Configurable;
 import java.net.URL;
+import org.jetbrains.annotations.NotNull;
```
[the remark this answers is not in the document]
FelixEngl: As far as i know there is no "official" annotation, so I'll just write why I used the Jetbrains one. (see [[1]](#sources) and [[2]](#sources))

There was once the javax.annotation thing, but as far as i know that one is basically dead. 

An other framework is SpotBugs the successor of FindBugs, but I have never used it because of trust issues (FindBugs was already abandoned, why shouldn't happen the same gain?) and there are more "standardly" feeling frameworks.

Another framework is Lombok, but this one is modifying the code to add real nullchecks to the code during runtime. But I don't like the idea of having a framework for annotating null-handling modifying written code, that only causes problems in the long run when the framework ceases to exist.

The next Framework is the Checker Framework. This is also the framework that Guava uses and can be seen as one of the more "standardly" feeling frameworks. But CF is quite big and does a lot of different checks that are far beyond simple documentation of "you can give me null" and properly using it seems quite bothersome because it has many annotations and most features are not supported by other checker frameworks and IDEs. (But I haven't used it so far, so I can only make assumptions about that support point. But I haven't found anything that says what is supported.)


The reasons why I chose `org.jetbrains.annotations` over the others are as follows:
- They only exist during compile time and are more part of the documentation like part of the code. So adding them like this is enough and we don't have to add any weird dependencies on compiler-plugins or whatnot:
```xml
<!-- https://mvnrepository.com/artifact/org.jetbrains/annotations -->
<dependency>
    <groupId>org.jetbrains</groupId>
    <artifactId>annotations</artifactId>
    <version>23.0.0</version>
    <scope>compile</scope>
</dependency>
```
- They are supported by CF, Eclipse, FindBugs, SpotBugs etc. (Regarding my research this annotations are basically supported by every big IDE and checking framework), giving them (at least for me) a very "standardly" feeling. (see [[5]](#sources) and [[6]](#sources) for an example how to use it in eclipse.)
- The amount of annotations is small (14) and are quite straight forward regarding their meaning [[3]](#sources).  Even if the framework has to be removed on a later date, there won't be much fuss about it because there are only 14 of them and you usually only use `@Nullable`, `@NotNull`, `@TestOnly` and `@Contract`.
- Some Apache Projects like Geode and Sling, and other opensource projects like testcontainer already use these annotations. Apache Spring explains it's usage of these annotations like this [[4]](#sources): 
    > The annotations used within Sling are based on the [Jetbrains Annotations](https://www.jetbrains.com/help/idea/nullable-and-notnull-annotations.html). Although introduced by the company that offers the IntelliJ IDEA IDE, those annotations are understood by most of the tools and used by other Apache Projects like Apache Oak.


## Sources
1. https://jcp.org/en/jsr/detail?id=305
2. An exhaustive list of nullness checking frameworks: https://checkerframework.org/manual/#nullness-related-work
3. https://github.com/JetBrains/java-annotations
4. https://sling.apache.org/documentation/development/null-analysis.html
5. https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.jdt.doc.user%2Ftasks%2Ftask-using_external_null_annotations.htm
6. https://sling.apache.org/documentation/development/null-analysis.html#use-with-eclipse

core/src/main/java/com/digitalpebble/stormcrawler/Metadata.java
```diff
@@ -243,7 +243,7 @@ public Metadata unlock() {
     }
 
     /** @since 1.16 */
-    private final void checkLockException() {
```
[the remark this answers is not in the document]
FelixEngl: Declaring a private method `final` is redundant because you can not override a private function and therefore private functions are already handled like being `final`. This is basically like declaring the methods of an public interface as public.

core/src/main/java/com/digitalpebble/stormcrawler/bolt/FetcherBolt.java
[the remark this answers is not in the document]
jnioche: OK

core/src/main/java/com/digitalpebble/stormcrawler/bolt/FetcherBolt.java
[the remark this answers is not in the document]
jnioche: can we remove this modification and keep it as it is?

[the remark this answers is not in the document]
FelixEngl: Yes, your old solution is even clearer than the new proposal because the new one it hides the behaviour if `localSitemapDiscoveryVal` is `null`.

core/src/main/java/com/digitalpebble/stormcrawler/bolt/JSoupParserBolt.java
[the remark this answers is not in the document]
jnioche: thanks for the detailed explanation. Makes sense

core/src/main/java/com/digitalpebble/stormcrawler/filtering/URLFilter.java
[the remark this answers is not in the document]
jnioche: OK, you  convinced me

core/src/main/java/com/digitalpebble/stormcrawler/parse/JSoupFilter.java
```diff
@@ -21,7 +21,7 @@
  * content. They are used exclusively by {@link
  * com.digitalpebble.stormcrawler.bolt.JSoupParserBolt}.
  */
-public abstract class JSoupFilter implements Configurable {
+public interface JSoupFilter extends Configurable {
```
jnioche, line 24: I would rather keep it as an abstract class.

core/src/main/java/com/digitalpebble/stormcrawler/persistence/AdaptiveScheduler.java
```diff
@@ -145,9 +145,7 @@
 
     /**
      * Name of key to hold previous signature: a copy, not overwritten by {@link
-     * MD5SignatureParseFilter}, is added by {@link
```
jnioche, line 148: why removing part of the Javadoc?

core/src/main/java/com/digitalpebble/stormcrawler/protocol/HttpHeaders.java
```diff
@@ -27,7 +27,9 @@
  * @see <a href="http://rfc-ref.org/RFC-TEXTS/2616/">Hypertext Transfer Protocol -- HTTP/1.1 (RFC
  *     2616)</a>
  */
-public interface HttpHeaders {
+public final class HttpHeaders {
```
jnioche, line 30: What's wrong with having an interface for this?

jnioche (changes requested):
Have gone through the changes. As discussed previously, it is neither practical nor reassuring to have 109 files changes in one go: spotting a problem in so many modifications is rather difficult. Could we please not have that in the future.
Some points are still open for discussion (see comments). Some changes need to be reverted.
Thanks!

core/src/main/java/com/digitalpebble/stormcrawler/protocol/Protocol.java
```diff
@@ -20,7 +20,7 @@
 
 public interface Protocol {
 
-    public void configure(Config conf);
+    void configure(Config conf);
```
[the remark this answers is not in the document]
rzo1: Interface = always `public`

core/src/main/java/com/digitalpebble/stormcrawler/parse/JSoupFilter.java
FelixEngl: > Have gone through the changes. As discussed previously, it is neither practical nor reassuring to have 109 files changes in one go: spotting a problem in so many modifications is rather difficult. Could we please not have that in the future. Some points are still open for discussion (see comments). Some changes need to be reverted. Thanks!

`Could we please not have that in the future.`: Definetly. I'm sorry to putting you through that. Future PRs will be in a reasonable size. Thank you for putting up with this PR.

I'll revert the changes you marked when I find some time to do it.

FelixEngl:
> Have gone through the changes. As discussed previously, it is neither practical nor reassuring to have 109 files changes in one go: spotting a problem in so many modifications is rather difficult. Could we please not have that in the future. Some points are still open for discussion (see comments). Some changes need to be reverted. Thanks!

`Could we please not have that in the future.`: Definetly. I'm sorry to putting you through that. Future PRs will be in a reasonable size. Thank you for putting up with this PR.

I'll revert the changes you marked when I find some time to do it.

jnioche:
> > Have gone through the changes. As discussed previously, it is neither practical nor reassuring to have 109 files changes in one go: spotting a problem in so many modifications is rather difficult. Could we please not have that in the future. Some points are still open for discussion (see comments). Some changes need to be reverted. Thanks!
> 
> `Could we please not have that in the future.`: Definetly. I'm sorry to putting you through that. Future PRs will be in a reasonable size. Thank you for putting up with this PR.
> 
> I'll revert the changes you marked when I find some time to do it.

thanks for your contribution @FelixEngl, much appreciated

core/src/main/java/com/digitalpebble/stormcrawler/protocol/HttpHeaders.java
FelixEngl: `HttpHeaders` is a utility class, meaning that it only has static, stateless functions without any other dependencies. But if you use an interface or abstract class for a utility class you communicate to the framework user that implementing this interface or inheriting from this abstract class is supposed to happen because the interface/abstract class is needed somewhere.

When you look at the java standard library they declare utility-classes like I do. (Except for the `final` declaration, thats just my personal taste to make it 100% clear, that noone can ever use this class for something else than it's static functions.)

```java
package java.util;
....

/**
 * This class contains various methods for manipulating arrays (such as
 * sorting and searching). This class also contains a static factory
 * that allows arrays to be viewed as lists.
 ...
 * @since  1.2
 */
public class Arrays {

    /**
    ...
     */
    private static final int MIN_ARRAY_SORT_GRAN = 1 << 13;

    // Suppresses default constructor, ensuring non-instantiability.
    private Arrays() {}

    ...
```

There are some use-cases for static methods in interfaces (see `java.util.stream.Stream`) but as long as an interface/abstract class doesn't declare any non-static methods and/or the static functions do not take/return the containing interface/abstract class we should implement utility classes like shown in the standard library for a clear intention communication.

core/src/main/java/com/digitalpebble/stormcrawler/spout/FileSpout.java
```diff
@@ -138,9 +132,6 @@ protected void populateBuffer() throws IOException {
                                     StandardCharsets.UTF_8));
         }
 
-        // no more files to read from
-        if (currentBuffer == null) return;
```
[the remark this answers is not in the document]
FelixEngl: We set the buffer in line 128 if it is null by a non-null-operation. Therefore the current buffer can never be `null` between line 128 and 145.
Or is `populateBuffer` of a single object instance called by multiple threads? Then we have to do something about threadsafety.

core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
```diff
@@ -29,13 +29,7 @@
...
 import java.net.URL;
 import java.security.cert.CertificateException;
-import java.util.ArrayList;
-import java.util.Base64;
-import java.util.HashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
+import java.util.*;
```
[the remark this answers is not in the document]
FelixEngl: I think I called that source formatting hook, but I'll check that up when I revert the other changes.
Thank you.

core/src/main/java/com/digitalpebble/stormcrawler/protocol/file/FileResponse.java
```diff
@@ -80,7 +80,6 @@ public FileResponse(String u, Metadata md, FileProtocol fileProtocol) throws IOE
             getFileAsHttpResponse(file);
         } else {
             statusCode = HttpStatus.SC_INTERNAL_SERVER_ERROR;
-            return;
```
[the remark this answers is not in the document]
FelixEngl: Yes that makes sense, I'll revert it.

core/src/main/java/com/digitalpebble/stormcrawler/parse/JSoupFilter.java
FelixEngl: That class declares only a single abstract function with no further logic. Declaring it as an interface would not change much but give more possibilities for inheritance/implementation.

A possible use-case is that you could use a configurable base-class that manages the connection for some kind of database for implementing a custom JSoupFilter.

core/src/main/java/com/digitalpebble/stormcrawler/persistence/AdaptiveScheduler.java
FelixEngl: Because `com.digitalpebble.stormcrawler.parse.filter.SignatureCopyParseFilter` does not exist. The rest of the comment was deleted by a mistake. I'll revert the rest.

core/src/main/java/com/digitalpebble/stormcrawler/parse/JSoupFilter.java
FelixEngl: Okay, forget what I said about the use case. You could get the same result by using a object for that.

Are you planning on extending that JsoupFilter at some point? Or is it just your personal taste?
I'm okay with both, but I personally think that an "abstract only" class should always be an interface.

jnioche: The class could possibly extended in the future + as a personal preference, I find abstract classes more future-proof.

core/src/main/java/com/digitalpebble/stormcrawler/persistence/AdaptiveScheduler.java
jnioche: ah, makes sense. Thanks

core/src/main/java/com/digitalpebble/stormcrawler/protocol/HttpHeaders.java
jnioche: OK, my perception of it probably dates back to when doing this as an interface was in fashion (whenever that was)

core/src/main/java/com/digitalpebble/stormcrawler/parse/JSoupFilter.java
FelixEngl: Ok, I'll revert it.

core/src/main/java/com/digitalpebble/stormcrawler/spout/FileSpout.java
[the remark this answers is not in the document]
jnioche: there is no worry about thread-safety here. My concern was that it could still be null if 

            _currentBuffer =
                    new BufferedReader(
                            new InputStreamReader(
                                    new FileInputStream(inputPath.toFile()),
                                    StandardCharsets.UTF_8));_
                                    
did not resolve i.e. the input path was incorrect but then we'd get an Exception and probably would not make it to line 142 anyway

core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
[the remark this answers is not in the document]
jnioche: Thanks for checking. That rewrite is probably done by Eclipse, not the formatter. No big deal either way

FelixEngl:
I found the reason for the compile error in urlfrontier. That was a reference that was accidentally changed from the refractor.

core/src/main/java/com/digitalpebble/stormcrawler/persistence/AdaptiveScheduler.java
```diff
@@ -237,7 +235,7 @@ public void init(Map stormConf) {
             // - old signature not copied
             // fall-back to DefaultScheduler
             LOG.debug("No signature for FETCHED page: {}", metadata);
-            if (setLastModified && signature != null && oldSignature == null) {
```
jnioche, line 240: this should not be removed

FelixEngl: `&& oldSignature == null` is redundant at this point and only costs an additional nullcheck because we can only get in this branch if either signature or oldSignature is null or both are null:

```java
if (Objects.equals(metadata.getFirstValue("fetch.statusCode"), "304")) {
...
} else if (signature == null || oldSignature == null) {
    // possible states: 
    // 1. signature IS NULL         AND oldSignature IS NOT NULL
    // 2. signature IS NOT NULL AND oldSignature IS NULL
    // 3. signature IS NULL         AND oldSignature IS NULL

    if(setLastModified && signature != null) {
        // signature is not null
        // Possible states left from above:
        // 2. signature IS NOT NULL AND oldSignature IS NULL
    }
}
```

core/src/main/java/com/digitalpebble/stormcrawler/bolt/FeedParserBolt.java
```diff
@@ -50,7 +50,6 @@
 import org.xml.sax.InputSource;
 
 /** Extracts URLs from feeds */
-@SuppressWarnings("serial")
```
rzo1, line 53: `BaseRichBolt` implements `Serializable`, so should'nt we add a serial version uuid to honor the contract? At least, if we remove the suppression ;)

The same is true for other sub-classes of the `BaseRichBolt` in the project.

core/src/main/java/com/digitalpebble/stormcrawler/filtering/sitemap/SitemapFilter.java
```diff
@@ -40,15 +42,17 @@
...
-                            .equalsIgnoreCase(
-                                    sourceMetadata.getFirstValue(
-                                            SiteMapParserBolt.foundSitemapKey))) return null;
+    public @Nullable String filter(
+            @Nullable URL sourceUrl,
+            @Nullable Metadata sourceMetadata,
+            @NotNull String urlToFilter) {
+
+        if (sourceMetadata != null
+                && !Boolean.parseBoolean(
```
rzo1, line 51: Should't we be consistent in the project? Other sources use the explicit form

core/src/main/java/com/digitalpebble/stormcrawler/persistence/DefaultScheduler.java
```diff
@@ -17,12 +17,7 @@
 import com.digitalpebble.stormcrawler.Constants;
 import com.digitalpebble.stormcrawler.Metadata;
 import com.digitalpebble.stormcrawler.util.ConfUtils;
-import java.util.Calendar;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
rzo1, line 20: Start import intended?

core/src/main/java/com/digitalpebble/stormcrawler/persistence/urlbuffer/SchedulingURLBuffer.java
```diff
@@ -27,8 +27,8 @@
 import java.util.Queue;
 import java.util.concurrent.TimeUnit;
 import org.apache.storm.tuple.Values;
-import org.checkerframework.checker.nullness.qual.NonNull;
```
rzo1, line 30: Do we know, who brings `checkerframework` in the project? Is it even reqired, after adding jetbrain annotations?

core/src/main/java/com/digitalpebble/stormcrawler/protocol/httpclient/HttpProtocol.java
```diff
@@ -197,7 +198,7 @@ public ProtocolResponse getProtocolOutput(String url, Metadata md) throws Except
 
         if (md != null) {
             String useHead = md.getFirstValue("http.method.head");
-            if ("true".equalsIgnoreCase(useHead)) {
+            if (Boolean.parseBoolean(useHead)) {
```
rzo1, line 201: see above

core/src/main/java/com/digitalpebble/stormcrawler/protocol/selenium/NavigationFilter.java
```diff
@@ -27,7 +27,7 @@
      * @param stormConf The Storm configuration used for the parsing bolt
      * @param filterParams the filter specific configuration. Never null
      */
-    public void configure(@SuppressWarnings("rawtypes") Map stormConf, JsonNode filterParams) {}
+    public void configure(Map stormConf, JsonNode filterParams) {}
```
rzo1, line 30: Why not `Map<String,Object>` here (was changed on other places accordingly) ?

core/src/main/java/com/digitalpebble/stormcrawler/protocol/selenium/NavigationFilters.java
```diff
@@ -55,7 +55,6 @@ public ProtocolResponse filter(RemoteWebDriver driver, Metadata metadata) {
      * Loads and configure the NavigationFilters based on the storm config if there is one otherwise
      * returns an emptyNavigationFilters.
      */
-    @SuppressWarnings("rawtypes")
     public static NavigationFilters fromConf(Map stormConf) {
```
rzo1, line 58: see above?

core/src/main/java/com/digitalpebble/stormcrawler/protocol/selenium/NavigationFilters.java
```diff
@@ -77,7 +76,6 @@ public static NavigationFilters fromConf(Map stormConf) {
      *
      * @throws IOException
      */
-    @SuppressWarnings("rawtypes")
     public NavigationFilters(Map stormConf, String configFile) throws IOException {
```
rzo1, line 79: see above?

core/src/main/java/com/digitalpebble/stormcrawler/protocol/selenium/NavigationFilters.java
```diff
@@ -99,7 +97,6 @@ public NavigationFilters(Map stormConf, String configFile) throws IOException {
         configure(stormConf, confNode);
     }
 
-    @SuppressWarnings("rawtypes")
     @Override
     public void configure(Map stormConf, JsonNode filtersConf) {
```
rzo1, line 101: see above?

core/src/main/java/com/digitalpebble/stormcrawler/spout/FileSpout.java
```diff
@@ -27,10 +27,7 @@
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Queue;
+import java.util.*;
```
rzo1, line 30: Star import intended?

core/src/main/java/com/digitalpebble/stormcrawler/util/Configurable.java
```diff
@@ -32,13 +33,13 @@
...
+    default void configure(
+            @NotNull Map<String, Object> stormConf, @NotNull JsonNode filterParams) {}
 
     /**
      * Used by classes URLFilters and ParseFilters classes to load the configuration of filters from
      * JSON
      */
-    @SuppressWarnings("rawtypes")
     public static <T extends Configurable> List<T> configure(
             Map stormConf, JsonNode filtersConf, Class<T> filterClass, String callingClass) {
```
rzo1, line 44: see above

core/src/test/java/com/digitalpebble/stormcrawler/parse/filter/CollectionTaggerTest.java
```diff
@@ -12,7 +13,7 @@
     public void testCollectionTagger() throws IOException {
 
         CollectionTagger tagger = new CollectionTagger();
-        tagger.configure(new HashMap(), null);
+        tagger.configure(new HashMap<>(), NullNode.instance);
```
rzo1, line 16: Looks odd to me replacing `null` with some instance from jackson?

external/sql/src/main/java/com/digitalpebble/stormcrawler/sql/SQLUtil.java
```diff
@@ -25,7 +25,6 @@
 
     private SQLUtil() {}
 
-    @SuppressWarnings({"rawtypes", "unchecked"})
     public static Connection getConnection(Map stormConf) throws SQLException {
```
rzo1, line 28: see above

external/tika/src/main/java/com/digitalpebble/stormcrawler/tika/ParserBolt.java
```diff
@@ -93,7 +92,6 @@
 
     private String protocolMDprefix;
 
-    @SuppressWarnings({"rawtypes", "unchecked"})
     @Override
     public void prepare(Map conf, TopologyContext context, OutputCollector collector) {
```
rzo1, line 96: see above

pom.xml
```diff
@@ -271,6 +272,14 @@
 				<artifactId>jackson-databind</artifactId>
 				<version>${jackson-databind.version}</version>
 			</dependency>
+
+			<!-- https://mvnrepository.com/artifact/org.jetbrains/annotations -->
+			<dependency>
+				<groupId>org.jetbrains</groupId>
+				<artifactId>annotations</artifactId>
+				<version>${jetbrains.annotations.version}</version>
+				<scope>compile</scope>
```
rzo1, line 281: Can we add `optional` here? I guess, that this dependency is not required at runtime, so it would be save to add an `optional` flag here...

rzo1 (commented):
Hi @FelixEngl and @jnioche 

I scanned the changes and added some comments / questions :)

Gruß
Richard

core/src/main/java/com/digitalpebble/stormcrawler/persistence/AdaptiveScheduler.java
jnioche: duhh, of course you are right. Need more coffee

core/src/main/java/com/digitalpebble/stormcrawler/persistence/DefaultScheduler.java
FelixEngl: It was not corrected by the formatter. so i think thats ok.

core/src/main/java/com/digitalpebble/stormcrawler/persistence/urlbuffer/SchedulingURLBuffer.java
FelixEngl: That was an oversight by me, I'll get to it.

core/src/main/java/com/digitalpebble/stormcrawler/filtering/sitemap/SitemapFilter.java
FelixEngl: What do you mean with explicit form?
As far as i know that is the "standard java way" or at least that what you should use in this situation.

FelixEngl: In the other situation from before, we have a distinction between "true", "false" and null/invalid. Therefore we had to distinct them. Here we have that it is either "true" or whatever.

rzo1: Sometimes the PR uses `true`.equalsIgnoreCase(...) and sometimes `Boolean.parseBoolean(...)`  ;-) - so i was thinking about consistency

rzo1: > In the other situation from before, we have a distinction between "true", "false" and null/invalid. Therefore we had to distinct them. Here we have that it is either "true" or whatever.

Yup - I can live with both versions. It was just about consistency :)

FelixEngl: Yes, but at some places I left out many useful/correct changes by intent.
Because this PR is already so big that it is practically a crime.

If you take a look at my fork (https://github.com/FelixEngl/storm-crawler/tree/local_version) you can see what I plan to add PR by PR. (except the changes of the pom.xml for installing it to my local repo.)

If you want/have the muse for it you can also give me feedback there if you see something you don't like or want changed. :)

core/src/main/java/com/digitalpebble/stormcrawler/protocol/httpclient/HttpProtocol.java
FelixEngl: see https://github.com/DigitalPebble/storm-crawler/pull/937#discussion_r801502015

core/src/main/java/com/digitalpebble/stormcrawler/protocol/selenium/NavigationFilter.java
FelixEngl: see https://github.com/DigitalPebble/storm-crawler/pull/937#discussion_r801502015

core/src/main/java/com/digitalpebble/stormcrawler/protocol/selenium/NavigationFilters.java
FelixEngl: [see ](https://github.com/DigitalPebble/storm-crawler/pull/937#discussion_r801502015)

core/src/main/java/com/digitalpebble/stormcrawler/protocol/selenium/NavigationFilters.java
FelixEngl: see https://github.com/DigitalPebble/storm-crawler/pull/937#discussion_r801502015

core/src/main/java/com/digitalpebble/stormcrawler/protocol/selenium/NavigationFilters.java
FelixEngl: see https://github.com/DigitalPebble/storm-crawler/pull/937#discussion_r801502015

core/src/main/java/com/digitalpebble/stormcrawler/util/Configurable.java
FelixEngl: see https://github.com/DigitalPebble/storm-crawler/pull/937#discussion_r801502015

core/src/test/java/com/digitalpebble/stormcrawler/parse/filter/CollectionTaggerTest.java
FelixEngl: That's a null node singleton and the tests always failed due to the fact, that the implementation wasn't supposed to consume null.

This was easier than changing the interface from `@NotNull` to `@Nullable`.

FelixEngl: Plus the nullcheck would be only needed for this test. Nowhere else the value becomes `null`.

external/sql/src/main/java/com/digitalpebble/stormcrawler/sql/SQLUtil.java
FelixEngl: see https://github.com/DigitalPebble/storm-crawler/pull/937#discussion_r801502015

external/tika/src/main/java/com/digitalpebble/stormcrawler/tika/ParserBolt.java
FelixEngl: see https://github.com/DigitalPebble/storm-crawler/pull/937#discussion_r801502015

pom.xml
FelixEngl: I don't know, I only saw it with this configuration. What do you think @jnioche?

jnioche: No idea. it is not really optional since it is needed a compilation time. Am not that bothered either way as long as it compiles

core/src/main/java/com/digitalpebble/stormcrawler/persistence/urlbuffer/SchedulingURLBuffer.java
FelixEngl: Looking at the tree it is from caffeine:
```
[INFO] +- com.github.ben-manes.caffeine:caffeine:jar:2.9.2:compile
[INFO] |  +- org.checkerframework:checker-qual:jar:3.10.0:compile
[INFO] |  \- com.google.errorprone:error_prone_annotations:jar:2.5.1:compile
```
But I'm not that accustomed to maven and pom.xml, what is the proper way of removing it? Just excluding like this?

```xml
<dependency>
	<groupId>com.github.ben-manes.caffeine</groupId>
	<artifactId>caffeine</artifactId>
	<version>${caffeine.version}</version>
	<exclusions>
		<exclusion>  <!-- declare the exclusion here -->
			<groupId>org.checkerframework</groupId>
			<artifactId>checker-qual</artifactId>
		</exclusion>
	</exclusions>
</dependency>
```

jnioche:
BTW @FelixEngl the name of the PR doesnt reflect the content. It is not just about contracts of interfaces but is a bit more general. What about "general code refactoring and good practices?"

core/src/main/java/com/digitalpebble/stormcrawler/persistence/urlbuffer/SchedulingURLBuffer.java
jnioche: if it is needed by Caffeine, don't exclude it. Just let it be

FelixEngl:
> BTW @FelixEngl the name of the PR doesnt reflect the content. It is not just about contracts of interfaces but is a bit more general. What about "general code refactoring and good practices?"

But that is only a part of all refactorings and good practices.

Maybe "partial general code refactoring and good practices" or "general code refactoring and good practices - part 1"?

jnioche:
> > BTW @FelixEngl the name of the PR doesnt reflect the content. It is not just about contracts of interfaces but is a bit more general. What about "general code refactoring and good practices?"
> 
> @jnioche But that is only a part of all refactorings and good practices.
> 
> Maybe "partial general code refactoring and good practices" or "general code refactoring and good practices - part 1"?

saying that it is a part doesn't really describe it any better + I am hoping that the next PRs will be more focused and thus will have a more specific description. it doesn't matter that much

FelixEngl:
> > > BTW @FelixEngl the name of the PR doesnt reflect the content. It is not just about contracts of interfaces but is a bit more general. What about "general code refactoring and good practices?"
> > 
> > 
> > @jnioche But that is only a part of all refactorings and good practices.
> > Maybe "partial general code refactoring and good practices" or "general code refactoring and good practices - part 1"?
> 
> saying that it is a part doesn't really describe it any better + I am hoping that the next PRs will be more focused and thus will have a more specific description. it doesn't matter that much

Yes, the other PRs will be more focused and less general by either replacing/adding a feature, fixing bugs/dangerous code or further refactoring single packages like `com.digitalpebble.stormcrawler.bold`.

pom.xml
rzo1: Importing projects will need to declare it if optional=true. 

We can leave it as is and importing projects can exclude it - these annotations are mit read at runtime, so should be save but we can test/change in another PR.

jnioche (approved):

jnioche:
thanks @FelixEngl and @rzo1

core/src/main/java/com/digitalpebble/stormcrawler/filtering/sitemap/SitemapFilter.java
rzo1: > If you take a look at my fork (https://github.com/FelixEngl/storm-crawler/tree/local_version) you can see what I plan to add PR by PR. (except the changes of the pom.xml for installing it to my local repo.)
> 
> If you want/have the muse for it you can also give me feedback there if you see something you don't like or want changed. :)

I have run SC for 370 days 24/7 with only a few downtimes for gathering PhD related data. The code per-se (and storm) is pretty stable but I am interested in the development of SC for further research questions. So yes - I am happy to have a look but we would need to enable some sort of communication on your fork. If you like, you can also drop a mail at my university mail address (should be obvious to find), so we can get in touch ;)

PR #947, opened 2022-02-09
Improve Metadata, add Tests and stabilize serialisation

FelixEngl:
Hello @jnioche,

sorry for the first PR regarding the metadata, that PR was a mistake. I thought that a draft was only visible for me.

I improved the class Metadata in the following ways:
- Add annotations like `@Nullable` for easier utilisation and protection from NPEs because Metadata is a central class for the whole project/usage.
- Replace the old serialisation for Kyro with an explicit implementation to keep the information about the empty metadata instance, even if we serialize and deserialize it.
- Rename `asMap` to `getMap` because as implies a cast or conversion but the method `asMap` is implemented as a simple getter.
- Improve `addValues` by removing the usage of collections.
- Implement `equals` for metadata for easier comparison.
- Implement a basic `hashCode` because it would be unusual to override equals but not hashCode. (Maybe create hash based on key values?)
- Add basic tests for Metadata
- Add a check if an instance of Metadata is an empty map with `isReadOnly` (maybe rename to `isEmptyInstance()`?)
- Add `deepCopy` methods for Metadata and the underlying Map to ensure that the original is not modified if not wanted.
- Deprecate `empty` in favor of `EMPTY_METADATA` and `emptyMetadata` in order to be more streamlined with `Collections.emptyList` and `Collections.emptyMap`. Furthermore I added explicit annotations to indicate that it is unmodifiable.

Best regards,

Felix

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/util/MetadataKryoSerializer.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/util/MetadataKryoSerializer.java
@@ -1,0 +1,64 @@
+/**
+ * Licensed to DigitalPebble Ltd under one or more contributor license agreements. See the NOTICE
+ * file distributed with this work for additional information regarding copyright ownership.
+ * DigitalPebble licenses this file to You under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy of the
+ * License at
+ *
+ * <p>http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * <p>Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.digitalpebble.stormcrawler.util;
+
+import static java.util.Collections.EMPTY_MAP;
+
+import com.digitalpebble.stormcrawler.Metadata;
+import com.esotericsoftware.kryo.Kryo;
+import com.esotericsoftware.kryo.Serializer;
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
+import com.esotericsoftware.kryo.serializers.DefaultArraySerializers;
+import com.esotericsoftware.kryo.serializers.DefaultSerializers;
+import java.util.Map;
+import org.apache.storm.serialization.types.HashMapSerializer;
+
+/** A Kryo serializer for Metadata */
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/digitalpebble/stormcrawler/util/MetadataKryoSerializer.java
```diff
@@ -0,0 +1,64 @@
...
+import com.esotericsoftware.kryo.io.Output;
+import com.esotericsoftware.kryo.serializers.DefaultArraySerializers;
+import com.esotericsoftware.kryo.serializers.DefaultSerializers;
+import java.util.Map;
+import org.apache.storm.serialization.types.HashMapSerializer;
+
+/** A Kryo serializer for Metadata */
+// We need that serializer to make sure, that we do protect an empty metadata instance.
+// Otherwise it'll be converted to an HashMap.
+public class MetadataKryoSerializer extends Serializer<Metadata> {
```
rzo1, line 32: Do we need to explicitly register this serializer to be used in the context of a `flux`-based topology? If so, we would need to adjust some config files.

FelixEngl: It is not explicitly registered (at least when I understand the term "explicit" in the kryo-context). That serializer is set as default serializer via annotation at the `Metadata`-class. 
So flux shouldn't have a problem with that. Even the registration in the topology is the same as before: see core/src/main/java/com/digitalpebble/stormcrawler/ConfigurableTopology.java

And we need that serializer, otherwise the EMPTY_MAP wont be handled correctly, and then we get that what @jnioche had to correct with the "remove empty metadata from ...."

FelixEngl: > It is not explicitly registered (at least what I understand by the term "explicit" in the kryo-context). That serializer is set as default serializer via annotation at the `Metadata`-class. So flux shouldn't have a problem with that. Even the registration in the topology is the same as before: see core/src/main/java/com/digitalpebble/stormcrawler/ConfigurableTopology.java
> 
> And we need that serializer, otherwise the EMPTY_MAP wont be handled correctly, and then we get that what @jnioche had to correct with the "remove empty metadata from ...."

Ups, didn't mean to write the @ that autocompletion got me off guard. :'D

FelixEngl:
Maybe we should finalize `Metdadata`? Because either in the old nor the new implementation it is possible to inherit from the metadata in a meanigful way.

jnioche:
This PR is probably too big and hasn't had any activity for nearly a year so I am closing it.

PR #974, opened 2022-06-14
Add test for Metadata.remove

lacinoire:
Hey 😊
I want to contribute a test.
Curious to hear what you think!

(I wrote this test as part of a research study at TU Delft. [Find out more](https://github.com/lacinoire/lacinoire/blob/main/README.md))

### DIFF
--- a/core/src/test/java/com/digitalpebble/stormcrawler/util/MetadataTransferTest.java
+++ b/core/src/test/java/com/digitalpebble/stormcrawler/util/MetadataTransferTest.java
@@ -14,6 +14,9 @@
  */
 package com.digitalpebble.stormcrawler.util;
 
+import static com.digitalpebble.stormcrawler.util.MetadataTransfer.trackDepthParamName;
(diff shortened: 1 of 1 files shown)


jnioche (commented):
thanks @lacinoire 
I am not sure that I understand what you are trying to test here. Your assert check that when removing a key from an empty metadata, the result is null. I don't think there is much value in it.
Also, this particular test class is about the metadata transfer mechanism but your test does not do anything with the metadata of the outlink. So if you want to test metadata removal, this is probably not the right place for it.
what do you think?

jnioche:
Closing for now, @lacinoire feel free to reopen or comment on it

lacinoire:
Thank you for the critical look @jnioche ! 🙂

I was taking inspiration from another existing test, and you're right the tracking depth and url are unnecessary for what I'm trying to test.

You say there is not much value for the project to test for a null result on the empty metadata:
Would you not test the path returning null in general,
or would a non-empty metadata (and a not-existing key) improve this test?

PR #980, opened 2022-07-01
Overhaul urlfrontier-bolts, add tests, find problem with URLFrontier

FelixEngl:
Hi @jnioche,

we switched our URL and Status handling from a custom bolt to URL-frontier. But I recognized, that the Status-Bold is not acking any tuple. After going into the code, adding some log-events and cleaning up the async-code and improving the state-management, my unit test shows the following log:

```
12:35:05.187 [Time-limited test] INFO  c.d.s.u.StatusUpdaterBolt - Initialisation of connection to URLFrontier service on localhost:53770
12:35:05.187 [Time-limited test] INFO  c.d.s.u.StatusUpdaterBolt - Allowing up to 100000 message in flight
12:35:05.194 [Time-limited test] ERROR c.d.s.u.PartitionUtil - Unknown partition mode : null - forcing to byHost
12:35:05.194 [Time-limited test] INFO  c.d.s.u.URLPartitioner - Using partition mode : QUEUE_MODE_HOST
12:35:05.263 [Time-limited test] TRACE c.d.s.u.StatusUpdaterBolt - Added to waitAck https://www.url.net/something with ID https://www.url.net/something total 1 - sent to localhost:53770
12:35:05.751 [grpc-default-executor-1] WARN  c.d.s.u.StatusUpdaterBolt - Could not find unacked tuple for blank id ``. (Ack: )
12:35:05.752 [grpc-default-executor-1] TRACE c.d.s.u.StatusUpdaterBolt - Trace for unpacked tuple for blank id: 
12:35:10.787 [Time-limited test] INFO  c.d.s.u.ChannelManager - Shutting down channel ManagedChannelOrphanWrapper{delegate=ManagedChannelImpl{logId=1, target=localhost:53770}}
```

It looks like URL-Frontier does not provide an ID when responding to a put, this is an error that can not be fixed on the SC side.

List of changes:
- Prevent starving of tuple by using a fair RW-lock for ack-queue reads and writes
- Prevent starving by only locking what needs to be locked
- Use semaphore instead of busy-waiting for limiting the number of URLItems to be sent. 
- Remove starving busy-waiting for acquiring a permit
- Add URL-Frontier test container (To be removed when the complete container mapped and added to the testcontainers repo)
- Add single Unit-Test for testing if acking in status-bold works. (Adding further Unit tests not necessary until URL-Frontier is fixed)
- Extend LOG-messaging

Best Regards
Felix

Signed-off-by: Felix Engl <felix.engl@uni-bamberg.de>

[...]

PR #982, opened 2022-07-05
Add ChannelManager for local channel management and constants to Spout.java

FelixEngl:
Add constants for Spout & ChannelManager. Modify pom.xml.

(see https://github.com/DigitalPebble/storm-crawler/pull/980)

Signed-off-by: Felix Engl <felix.engl@uni-bamberg.de>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

Thanks!


external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/Spout.java
```diff
@@ -71,25 +71,27 @@ public void open(
             int nodeIndex = context.getThisTaskIndex();
             Collections.sort(addresses);
             address = addresses.get(nodeIndex);
+        } else if (addresses.size() == 1) {
```
jnioche, line 74: why not be flexible and let users specify a single address? Won't do any harm

external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/Spout.java
```diff
@@ -121,7 +124,7 @@ protected void populateBuffer() {
         final long start = System.currentTimeMillis();
 
         StreamObserver<URLInfo> responseObserver =
-                new StreamObserver<URLInfo>() {
+                new StreamObserver<>() {
```
jnioche, line 127: why?

external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/Spout.java
```diff
@@ -177,7 +196,6 @@ public void fail(Object msgId) {
     @Override
     public void close() {
         super.close();
-        LOG.info("Shutting down connection to URLFrontier service");
```
jnioche, line 180: why remove this log? it is useful

pom.xml
```diff
@@ -76,6 +76,8 @@
 		<mockito.version>4.6.1</mockito.version>
 		<jetbrains.annotations.version>23.0.0</jetbrains.annotations.version>
 		<log4j2.version>2.17.2</log4j2.version>
+		<testcontainer.version>1.17.3</testcontainer.version>
+		<urlfrontier.version>2.2</urlfrontier.version>
```
jnioche, line 80: having module specific variables in the top most pom defeats the objects of putting things into modules

external/urlfrontier/pom.xml
```diff
@@ -16,11 +16,49 @@
...
+			<artifactId>storm-crawler-core</artifactId>
+			<version>${project.version}</version>
+			<type>test-jar</type>
+			<scope>test</scope>
+		</dependency>
+
+		<dependency>
+			<groupId>org.testcontainers</groupId>
+			<artifactId>testcontainers</artifactId>
+			<version>${testcontainer.version}</version>
```
jnioche, line 60: I see, we're using the variable because it is not exactly the same dependency we are using here and in the es module and therefore can't set the version once and for all in the parent pom. Correct?

jnioche (changes requested):
thanks! apart from the use of constants, which is great, the main addition by this PR is the use of the ChannelManager. Can you please explain what it does and why it is needed?

external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/Spout.java
FelixEngl: That was a mistake. I'll put it back.

FelixEngl:
> thanks! apart from the use of constants, which is great, the main addition by this PR is the use of the ChannelManager. Can you please explain what it does and why it is needed?

A managed channel provides Livecycle Management of a connection to a server. If we have the Spout and Status on the same  JVM we should use the same ManagedChannel in both nodes.

https://grpc.github.io/grpc-java/javadoc/io/grpc/ManagedChannel.html

(Unless I misunderstood the documentation. But as far as I understood it, best practice would be one instance per Frontier-Host, per JVM for connection management etc.)

external/urlfrontier/pom.xml
FelixEngl: The problem is, that I am Implementing a GenericContainer and I don't know, how often these APIs change.
But I think we could put it in the externals-pom.xml

external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/Spout.java
jnioche: OK, thanks. If it isn't working then let's not allow it as per your suggestion. We can always fix it later

external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/Spout.java
jnioche: yes you are right. it is tautological

pom.xml
jnioche: yes, that would be perfect

jnioche:
> > thanks! apart from the use of constants, which is great, the main addition by this PR is the use of the ChannelManager. Can you please explain what it does and why it is needed?
> 
> A managed channel provides livecycle management of a connection to a server. If we have the Spout and Status on the same JVM we should use the same ManagedChannel in both nodes.
> 
> https://grpc.github.io/grpc-java/javadoc/io/grpc/ManagedChannel.html
> 
> (Unless I misunderstood the documentation. But as far as I understood it, best practice would be one instance per Frontier-Host, per JVM)

Thanks for the explanation. You probably don't want it per Frontier host as multiple instance could be running on that host with different ports. 

There is an opposite argument for having separate channels, in fact, this is why I recently allowed the StatusUpdaterBolt instances to be a multiple of the number of target frontiers. See https://grpc.io/docs/guides/performance/. Since both the spout and the statusupdater operations are streaming ones, the server uses a single thread to deal with the sequence of events. If you have multiple channels, the Frontier runs the operations in parallel. If you run a large crawl hitting a single Frontier and look at the CPU usage, you will notice that it doesn't get very high - that's because often it won't be using all the threads it can use. 

See https://github.com/crawler-commons/url-frontier/issues/41 for a related issue. A similar one could be made for the putting of the URLs, so that when dealing with a stream, it would use a pool of threads to deal with each update one by one.

FelixEngl:
> 

Thank you, I didn't know that page. (I'll read into this a little bit more, the Grpc protocol really seems convenient. They even got Kotlin support.)
When I understood the paragraph `(Special topic)` correctly, Grpc has (at the moment) a problem in high load/long long-lived areas (like Spouts/Status) and that's why you decided to use multiple channel instances in the bolts like they said. For the time being I'll remove the ChannelManager from this PR and the associated ones.

What do you think about writing a comment or long term issue relating to that point? Because at some point this Grpc-problem will be fixed and then we should start reusing channels as described in best practices.

jnioche:
> What do you think about writing a comment or long term issue relating to that point? Because at some point this Grpc-problem will be fixed and then we should start reusing channels as described in best practices.

yes, a comment to that effect would be good

external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/Spout.java
```diff
@@ -50,56 +51,67 @@ public class Spout extends AbstractQueryingSpout {
...
+                                    stormConf, URLFRONTIER_HOST_KEY, URLFRONTIER_DEFAULT_HOST),
+                            ConfUtils.getInt(
+                                    stormConf, URLFRONTIER_PORT_KEY, URLFRONTIER_DEFAULT_PORT));
+        } else {
+            channel = ManagedChannelToolkit.createChannel(address);
         }
 
-        channel = ManagedChannelBuilder.forTarget(address).usePlaintext().build();
-        frontier = URLFrontierGrpc.newStub(channel);
+        frontier = URLFrontierGrpc.newStub(channel).withWaitForReady();
```
jnioche, line 113: what does the waitForReady do?

pom.xml
```diff
@@ -76,6 +76,7 @@
 		<mockito.version>4.6.1</mockito.version>
 		<jetbrains.annotations.version>23.0.0</jetbrains.annotations.version>
 		<log4j2.version>2.17.2</log4j2.version>
+		<testcontainer.version>1.17.3</testcontainer.version>
```
jnioche, line 79: same as above, this change is about a different PR. Doesn't matter too much though

jnioche (commented):
A few minor points, main one being the name of ManagedChannelToolkit but this can be changed later on. Thanks

external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/Spout.java
FelixEngl: `waitForReady` ensures that the channel is in the state `READY` before sending an RPC. Due to the fact, that we are in an asynchronous inviroment we could have the case, that we try to send an RPC before the connection is properly initialized.

> gRPC implementations MAY provide a per-RPC option to not fail RPCs as a result of the channel being in TRANSIENT_FAILURE state. Instead, the implementation queues the RPCs until the channel is READY. This is known as "wait for ready." The RPCs SHOULD still fail before READY if there are unrelated reasons, such as the channel is SHUTDOWN or the RPC's deadline is reached.

https://github.com/grpc/grpc/blob/master/doc/wait-for-ready.md

external/urlfrontier/pom.xml
```diff
@@ -16,11 +16,53 @@
 	<url>https://github.com/DigitalPebble/storm-crawler/tree/master/external/urlfrontier</url>
 	<description>URL Frontier resources for StormCrawler</description>
 
+	<properties>
+		<urlfrontier.version>2.2</urlfrontier.version>
+	</properties>
+
+	<build>
```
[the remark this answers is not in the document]
FelixEngl: Can we keep it or do you want it to be removed?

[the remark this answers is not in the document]
jnioche: no big deal let's keep it

external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/Spout.java
jnioche: thanks, I would have expected this to be the case by default but good to have now

PR #1002, opened 2022-10-26
Create an AbstractFilter

Mikwiss:
Hello all, 

Our parseFilters.json configuration are more and more complex. We use same filter class many times to achieve specific behavior. Futhermore, we create a JsonParserBolt (to handle application/json response) and a FormatterBolt (used for example to format a price like "123 , 45" into "123.45"). So we have a lot of custom filter.

All those filters can log, a thousand times (our bots parse millions of url). So, having the filter name in log we'll help us to debug ! Currently it is only use on the setup : 

> 2022-10-26 13:23:16.080 c.d.s.u.ConfigurableHelper Thread-27-parser-executor[11, 11] [INFO] Setup MyNameFilter[com.mycompany.crawler.storm.bolt.parser.jsoup.filter.MyFilterClass]

In this PR, I also change the JsoupFilter abstract class to an interface. So we can use in the same way the new AbstractFilter :

`public class MyJsoupParseFilter extends AbstractFilter implements JSoupFilter { }`
`public class MyJsonParseFilter extends AbstractFilter implements JSonFilter { }`
`public class MyFormatterFilter extends AbstractFilter implements FormatterFilter { }`

I know the last point can affect a lot of project, so I'm open to feedback 👍 

Regards.

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

Thanks!

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/jsoup/LDJsonParseFilter.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/jsoup/LDJsonParseFilter.java
@@ -18,6 +18,7 @@ import com.digitalpebble.stormcrawler.Metadata;
 import com.digitalpebble.stormcrawler.parse.JSoupFilter;
 import com.digitalpebble.stormcrawler.parse.ParseData;
 import com.digitalpebble.stormcrawler.parse.ParseResult;
+import com.digitalpebble.stormcrawler.util.AbstractFilter;
 import com.fasterxml.jackson.core.JsonPointer;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -36,7 +37,7 @@ import org.slf4j.LoggerFactory;
  * Extracts data from JSON-LD representation (https://json-ld.org/). Illustrates how to use the
  * JSoupFilters
  */
-public class LDJsonParseFilter extends JSoupFilter {
+public class LDJsonParseFilter extends AbstractFilter implements JSoupFilter {
 
     public static final Logger LOG = LoggerFactory.getLogger(LDJsonParseFilter.class);
 
@@ -107,7 +108,7 @@ public class LDJsonParseFilter extends JSoupFilter {
             }
 
         } catch (Exception e) {
-            LOG.error("Exception caught when extracting json", e);
+            LOG.error("{} - Exception caught when extracting json", getName(), e);
         }
     }
 }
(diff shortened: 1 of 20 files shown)


jnioche:
Thanks for the PR @Mikwiss. 
This is a lot of files changed and affecting core classes. I am not convinced we should changed the existing abstract classes into interfaces. I agree it would be useful to have the class names in the logs but this in itself is not enough to justify such a change.
Why not adding the _getName()_ method directly into the abstract classes _JSoupFilter_ and _ParseFilter_ instead?

Mikwiss:
Hi @jnioche ! Thanks for feedback !

I fully understand ! I'll try to be more specific.

All the fetched url don't have the same behavior. For each url type (`url.type`, see below), we want to execute specific parser. For example, we create : 
- _BuilderOutlinkFilter_ : create a outlink from a base url and some parse data
- _SplitHtmlFilter_ : in order to split current htlml into more little pieces. Each one in a new `ParseData`.
- ...

So we created a useful `AbstractScopeFilter`. The filter evaluate the specified `scope` and execute the filter method only if the scope is correct.

```json
{
      "class": "com.mycompany.crawler.storm.bolt.parser.jsoup.filter.FilterA",
      "_comment": "Filter used to do a specific thing.",
      "name": "SpecificBehaviorFilterA",
      "params": {
        "scope": {
          "metadata" : [ "url.type=SEARCH_URL", "parse.previousParsedData!=(.+)" ],
          "operator" : "AND"
        },
        "other_parameter" : "..."
}
```

Now, let's look at the `JsoupFilters.configure()` (same as `ParserFilters.configure`) :

```java
 @Override
    public void configure(@NotNull Map<String, Object> stormConf, @NotNull JsonNode filtersConf) {
        List<JSoupFilter> list =
                Configurable.createConfiguredInstance(
                        this.getClass(), JSoupFilter.class, stormConf, filtersConf);
        filters = list.toArray(new JSoupFilter[0]);
    }
```
It's only instantiate (and configure) abstract `JsoupFilter`.  So we cannot use `AbstractScopeFilter` (or any other Abstract code) as `Abstract` but like a object inside each filters. It's seems greedy and not inheritance friendly. In my point of view it's better to use `JouspFilter` as interface. Please see a (beautiful) drawing of our proposal :

image

Regards.

jnioche:
> Please see a (beautiful) drawing of our proposal

a beautiful drawing indeed!

Thanks for the explanation, I can see the value of your proposal but am still trying to weigh the pros and cons of such a change.

Since _AbstractFilter_ doesn't do anything about filtering in itself, why not call it _AbstractConfigurable_ instead? 

I would also modify the _Configurable_ interface so that it exposes _getName()_. The reason for this is that every instantiation goes through _ConfigurableHelper_ anyway and it has access to the name (or generates it from the class name), so we might as well remove 

```
    default void configure(
            @NotNull Map<String, Object> stormConf, @NotNull JsonNode filterParams) {}
```

from Configurable and replace it with the variant using             `@NotNull String filterName`.

Now both _ParseFilter_ and _JSoupFilter_ could extend _AbstractConfigurable_ and would therefore gain _getName()_.

Could this be doable as an intermediate step? We can then discuss the merits of abstract classes vs interfaces in a separate PR, which would have fewer changes and would be easier to review.

jnioche:
@Mikwiss  see #1006

Issue #959, opened 2022-03-25
Setting "maxDepth": 0 in urlfilter.json prevents ES seed injection


orliac:
With StormCrawler 2.3-SNAPSHOT, setting "maxDepth": 0 in the urlfilters.json prevents the seed injection into the ES index.

Expected behavior would be that the seeds would be injected and crawled with no redirection.

jnioche:
thanks @orliac, 

Am on it. 

For an explanation I am copying my answer on https://stackoverflow.com/questions/71599054/stormcrawler-setting-maxdepth-0-prevents-es-seed-injection/71599845#71599845

> 
> See [MaxDepthFilter](https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/java/com/digitalpebble/stormcrawler/filtering/depth/MaxDepthFilter.java#L69), with a value of 0, everything gets filtered. Setting the filter to a value of 1 should do the trick, the seeds will be injected but their links won't be followed.
> 
> In MaxDepthFilter,
> 
> private String filter(final int depth, final int max, final String url) {
>     // deactivate the outlink no matter what the depth is
>     if (max == 0) {
>         return null;
>     }
>     if (depth >= max) {
>         LOG.debug("filtered out {} - depth {} >= {}", url, depth, maxDepth);
>         return null;
>     }
>     return url;
> }
> turns out that URLs need to have a depth of max-1 to be kept, so to put it differently, the actual maximum depth is max-1.
> 
> This feels not right and slightly confusing, I agree.
> 
> I think this is due to the sequence in which the outlinks get filtered. Often, this is done in the [StatusEmitterBolt](https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/java/com/digitalpebble/stormcrawler/bolt/StatusEmitterBolt.java#L96).
> 
> At the moment they first get filtered then inherit their metadata from the parent metadata. It is during that later step that their depth value gets incremented. I suspect this is why we are doing the max-1 trick.
> 
> There probably was a reason why the filtering was done first then the metadata inheritance but it has been a while and I can't remember any. I would be happy to change the order and get the metadata then filter and change the depth filtering so that it is more intuitive. 
>

jnioche:
thanks @orliac, should now be fixed. I simply removed the special rule for max depth 0

PR #1003, opened 2022-10-26
Can disable MaxDepthFilter

Mikwiss:
Hi all !

We spotted another border side. Our team recently upgrade StormCrawler version (2.2 to 2.5). The MaxDepthFilter was configured to `maxDepth : 0`, and no links was emitted (correct behavior).
After the upgrade, the first link (so with no `depth` in metadata) has emitted a hundred unwanted links. So in order to disable it, I propose to remove a condition in order to handle the `maxDepth : -1` parameter (equals to a disable). 

Let me know if it's ok for you.

Regards.

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

Thanks!

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/filtering/depth/MaxDepthFilter.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/filtering/depth/MaxDepthFilter.java
@@ -58,10 +58,9 @@ public class MaxDepthFilter implements URLFilter {
             return filter(depth, customMax, url);
         }
         // rely on the default max otherwise
-        else if (maxDepth >= 0) {
+        else {
             return filter(depth, maxDepth, url);
         }
-        return url;
     }
 
     private String filter(final int depth, final int max, final String url) {
--- a/core/src/test/java/com/digitalpebble/stormcrawler/filtering/MaxDepthFilterTest.java
+++ b/core/src/test/java/com/digitalpebble/stormcrawler/filtering/MaxDepthFilterTest.java
@@ -38,6 +38,15 @@ public class MaxDepthFilterTest {
     }
 
     @Test
+    public void testDepthNegative() throws MalformedURLException {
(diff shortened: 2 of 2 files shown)


jnioche:
Thanks for reporting this

> Our team recently upgrade StormCrawler version (2.2 to 2.5). The MaxDepthFilter was configured to maxDepth : 0, and no links was emitted (correct behavior).

This was changed in 2.4, see discussion at #959

> After the upgrade, the first link (so with no depth in metadata) has emitted a hundred unwanted links. So in order to disable it, I propose to remove a condition in order to handle the maxDepth : -1 parameter (equals to a disable).

-1 means "disable the filtering i.e let any outlink through", if you want to deactivate any outlink, set the value to 0. This should not change I think.

What should change though is how no existing depth is handled. Currently it is getting -1, like maxDepth but it should be 0 instead. I will commit a fix for it shortly, which will also make the code easier to read.

Feel free to reopen if you think that commit doesn't solve things. Thanks!

PS: haven't forgotten your other PR

jnioche:
Thanks for reporting this. 

The handling of 0 was changed in 2.4, see #959. I need to revisit that decision as it probably wasn't the best way of doing. It is simply that the filtering in that case was applied to the source URL whereas in the rest of the time it is applied to outlinks.

Either way, -1 means something different altogether. It indicates that the filtering should not be applied at all and as such all the outlinks should be produced. 

I will push a commit to simplify that part of the code, we can then think about how to handle #959.

jnioche:
Fixed in c7243e51
Setting the value to 0 should disallow outlinks. Let me know what you think @Mikwiss

Mikwiss:
Hey @jnioche, 

Our team has just tested the fix : https://github.com/DigitalPebble/storm-crawler/commit/c7243e510d65c96844cadbed5fb7477bed5d7df7 

It's perfect ! Much better as my fix, well done !

PR #1006, opened 2022-11-14
Added AbstractConfigurable + URLFilter becomes an abstract class. Nam…

jnioche:
See discussion in #1002

This creates an _AbstractConfigurable_ class and changes _URLFilter_ from being an interface to becoming an abstract class extending _AbstractConfigurable_. _Configurable_ now has a 

`    public String getName();
` 

which _AbstractConfigurable_ handles.

This allows extensions to access the name of the configurable e.g. for logging.

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/filtering/URLFilter.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/filtering/URLFilter.java
@@ -15,7 +15,7 @@
 package com.digitalpebble.stormcrawler.filtering;
 
 import com.digitalpebble.stormcrawler.Metadata;
-import com.digitalpebble.stormcrawler.util.Configurable;
+import com.digitalpebble.stormcrawler.util.AbstractConfigurable;
 import java.net.URL;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -26,7 +26,7 @@ import org.jetbrains.annotations.Nullable;
  *
  * @see URLFilters for more information.
  */
-public interface URLFilter extends Configurable {
+public abstract class URLFilter extends AbstractConfigurable {
 
     /**
      * Returns null if the URL is to be removed or a normalised representation which can correspond
@@ -39,7 +39,7 @@ public interface URLFilter extends Configurable {
      *     to the input URL
      */
     @Nullable
-    String filter(
+    public abstract String filter(
             @Nullable URL sourceUrl,
             @Nullable Metadata sourceMetadata,
             @NotNull String urlToFilter);
--- a/core/src/main/java/com/digitalpebble/stormcrawler/util/AbstractConfigurable.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/util/AbstractConfigurable.java
@@ -18,7 +18,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import java.util.Map;
 import org.jetbrains.annotations.NotNull;
 
-public class AbstractConfigurable implements Configurable {
+public abstract class AbstractConfigurable implements Configurable {
 
     private String name;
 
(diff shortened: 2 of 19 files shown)


core/src/main/java/com/digitalpebble/stormcrawler/filtering/URLFilter.java
```diff
@@ -26,7 +26,7 @@
  *
  * @see URLFilters for more information.
  */
-public interface URLFilter extends Configurable {
+public abstract class URLFilter extends AbstractConfigurable {
```
rzo1, line 29: I guess, that it is in the nature of this change, that we are breaking user code, right? So we should mention that in the release notes.

core/src/main/java/com/digitalpebble/stormcrawler/util/Configurable.java
```diff
@@ -42,7 +45,7 @@ default void configure(
      *     information.
      */
     @NotNull
-    static <T extends Configurable> List<@NotNull T> createConfiguredInstance(
+    static <T extends AbstractConfigurable> List<@NotNull T> createConfiguredInstance(
```
rzo1, line 48: Do we need these changes here? The `getName()` is `Configurable`, so is there a real need to change it here?

core/src/main/java/com/digitalpebble/stormcrawler/filtering/URLFilter.java
jnioche: yes, we will

core/src/main/java/com/digitalpebble/stormcrawler/util/Configurable.java
rzo1: Yes. I opened it in my IDE. Way better than using the web diff view, so yeah - you are right.

rzo1 (approved):

Mikwiss:
Hi @jnioche, 

We work on the same "feature" (https://github.com/Mikwiss/storm-crawler/tree/abstractconfigurable) :

- We made UrlFilter as an interface, but Abstract it's ok
- AbstractConfigurable is not Abstract (I commented)

In a second hand, we will submit a PR to discuss about making JouspFilter as interface (https://github.com/DigitalPebble/storm-crawler/pull/1002#issuecomment-1302198372). I close the https://github.com/DigitalPebble/storm-crawler/pull/1002.

core/src/main/java/com/digitalpebble/stormcrawler/util/AbstractConfigurable.java
```diff
@@ -0,0 +1,36 @@
...
+ * express or implied. See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.digitalpebble.stormcrawler.util;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import java.util.Map;
+import org.jetbrains.annotations.NotNull;
+
+public class AbstractConfigurable implements Configurable {
```
Mikwiss, line 21: public abstract class AbstractConfigurable implements Configurable {

Mikwiss (changes requested):

core/src/main/java/com/digitalpebble/stormcrawler/util/AbstractConfigurable.java
jnioche: yes indeed

jnioche:
> Thanks a lot for your job btw ;)

Pleasure. Thanks in turn for your suggestions and contributions, they are much appreciated.

core/src/main/java/com/digitalpebble/stormcrawler/util/AbstractConfigurable.java
jnioche: fixed in 8fbdeb80

Issue #1025, opened 2023-01-04
Need Code refactoring for better code reusability in StatusUpdaterBolt


msghasan:
We should update **store()** method inside **StatusUpdaterBolt**. I will explain why it can increase code re usability.

- If someone needs to extend this class for some business requirement to customize his requirement I think the person needs to copy paste all the things from **StatusUpdaterBolt** .
- Need to add a parameter in the **store()** method which accpets sha256 key as string without breaking the existing store method i.e the idea is to create a method with existing parameter which will call the latest **store()** method but will pass sha256 string as null.
- Thirdly we should refactor the content builder part in else part so that if it is in a separate method others can reuse it.

I hope I was able to explain the enhancement.

Below is how the new code will look for the first two points: -

NEW METHOD DEF: public void store( String url, Status status, Metadata metadata, Optional<Date> nextFetch, Tuple tuple, String sha256Hex) throws Exception {
if(StringUtils.isEmpty(sha256Hex)){
   sha256hex = org.apache.commons.codec.digest.DigestUtils.sha256Hex(url);
}
rest as is except the content builder in a new method
}

# Helper method so that existing code should not break
 public void store( String url, Status status, Metadata metadata, Optional<Date> nextFetch, Tuple tuple) throws Exception {
               store(  url, status, metadata,  nextFetch,  tuple, null);
}

msghasan:
I could not add the label as wish

msghasan:
@jnioche  Can I provide the fix and raise a PR

jnioche:
Sure, this will help us understand what you want to achieve. As far as the doc ID is concerned, I'd rather we had the option to take it from the metadata, a bit like what is suggested in #671 for the IndexerBolt. This would give you the flexibility to generate the Id any way you like.

msghasan:
81b85a159af6a7753d60b08d3dab1ff1cf7d0be4

msghasan:
@jnioche I thought I would be able to create a new PR but my new commits went into and existing pr I have mapped the particular commit here. 
Request you guys to verify my previous PR and also would be starting to work on Selenium Grid tomorrow. Please verify and approve the changes..
Regards

jnioche:
@msghasan it is not clear from the dangling commit you mentioned what you are trying to achieve and because it is not in a PR it makes it difficult to comment on specific changes.
In any case we will be addressing the doc id issue in #1028 so there is no point doing it in here.
[...]

Issue #671, opened 2018-12-05
Indexer bolts to use value from metadata as docID


jnioche:
The default value would still be the URL or canonical tag but if a value is found for a predefined key, its value would be used. This would help with deduplication as a ParseFIlter could generate a hash from the content and use it for _id.

PR #1028, opened 2023-01-05
Indexer ES document id

Mikwiss:
Hi @jnioche,

I saw the https://github.com/DigitalPebble/storm-crawler/issues/1025 issue and your answer related to https://github.com/DigitalPebble/storm-crawler/issues/671. That's was exactly the purpose of this PR !

In our case we use a random UUID to store the content (to keep all data). We add a new boolean parameter to handle this : `es.indexer.id.random`. The `getDocumentId()` can also be overrided to create any ID with metadata.

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

Thanks!

### DIFF
--- a/external/solr/src/main/java/com/digitalpebble/stormcrawler/solr/bolt/IndexerBolt.java
+++ b/external/solr/src/main/java/com/digitalpebble/stormcrawler/solr/bolt/IndexerBolt.java
@@ -104,7 +104,7 @@ public class IndexerBolt extends AbstractIndexerBolt {
                 // Distinguish the value used for indexing
                 // from the one used for the status
                 String normalisedurl = valueForURL(tuple);
-                doc.addField(fieldNameForURL(), normalisedurl);
+                doc.addField(fieldNameForURL(), getDocumentID(metadata, normalisedurl));
             }
 
             // select which metadata to index
@@ -132,5 +132,10 @@ public class IndexerBolt extends AbstractIndexerBolt {
             LOG.error("Send update request to SOLR failed due to {}", e);
             _collector.fail(tuple);
         }
+    }
+
+    @Override
+    protected String getDocumentID(Metadata metadata, String normalisedUrl) {
+        return normalisedUrl;
     }
 }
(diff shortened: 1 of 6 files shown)


rzo1:
I am wondering, if it would be enough to provide

 ```
 protected String getDocumentID(Metadata metadata, String normalisedUrl) {
 //Return current default
}
```

so user's can override as needed without introducing a new config option?

**Comment by Mikwiss** [comment]
It's enough indeed ! Maybe I am the only one to want keep data :)

**Comment by jnioche** [comment]
Yes that would work as well

**Review by rzo1 — APPROVED** [review]

**Comment by jnioche** [comment]
We'll need this for the other modules (SOLR, OpenSearch etc...) so I'd rather we did it for all of them in one go.
The new method could go in _AbstractIndexerBolt_ with the default implementation there, then call it in each implementation.

**Comment by rzo1** [comment]
> We'll need this for the other modules (SOLR, OpenSearch etc...) so I'd rather we did it for all of them in one go.
> The new method could go in _AbstractIndexerBolt_ with the default implementation there, then call it in each implementation.

+1

**Comment by Mikwiss** [comment]
Hi guys, 

I made the change, but I'm not sure about _solr_ and _sql_ (despite changes don't change the behavior).

File: external/solr/src/main/java/com/digitalpebble/stormcrawler/solr/bolt/IndexerBolt.java [hunk 1]
```diff
@@ -104,7 +104,7 @@ public void execute(Tuple tuple) {
                 // Distinguish the value used for indexing
                 // from the one used for the status
                 String normalisedurl = valueForURL(tuple);
-                doc.addField(fieldNameForURL(), normalisedurl);
+                doc.addField(fieldNameForURL(), getDocumentID(metadata, normalisedurl));
```
jnioche, line 107: This is to store the URL, not the doc ID. I don't think we should call getDocumentID here, even if it returns the normalised URL.

external/sql/src/main/java/com/digitalpebble/stormcrawler/sql/IndexerBolt.java
```diff
@@ -139,7 +139,7 @@ public void execute(Tuple tuple) {
 
             // send URL as field?
             if (fieldNameForURL() != null) {
-                preparedStmt.setString(1, normalisedurl);
+                preparedStmt.setString(1, getDocumentID(metadata, normalisedurl));
```
jnioche, line 142: same as above, this is about the URL, not the docID

jnioche (changes requested):

jnioche:
thanks @Mikwiss!

Issue #1038, opened 2023-02-10
Limit the amount of text to be returned by the text extraction


jnioche:
The TextExtraction can generates very long text on dodgy inputs and takes forever to do so.

This page http://robgagnon.net/GagnonCompositeRankingTopColleges2016-17.htm for instance can take 7secs just to extract the text.
This happens when ARTICLE is set for include param name, not sure why, maybe it is because the corresponding tags are never closed.

As a safeguard, we can add a new configuration element to limit the amount of text to be returned by the text extraction. When set to _10000000_ for the document above, the processing time drops to 378 msec.

PR #1039, opened 2023-02-10
Limit the amount of text to be returned by the text extraction, #1038

jnioche:
@rzo1 would you mind having a look? There could be further improvements to this code but at least for now it will prevent silly cases


core/src/main/java/com/digitalpebble/stormcrawler/parse/TextExtractor.java
```diff
@@ -138,14 +143,71 @@ public void tail(Node node, int depth) {
...
+                    continue; // don't tail removed
+                }
+            }
+
+            if (node.childNodeSize() > 0) { // descend
+                node = node.childNode(0);
+                depth++;
+            } else {
+                while (true) {
+                    assert node != null; // as depth > 0, will have parent
```
rzo1, line 195: don't know, if we get much from the assertion here as `-ea` isn't present at runtime?

core/src/main/java/com/digitalpebble/stormcrawler/parse/TextExtractor.java
```diff
@@ -138,14 +143,71 @@ public void tail(Node node, int depth) {
...
+                    if (!(node.nextSibling() == null && depth > 0)) break;
```
rzo1, line 196: IMHO it would be more readable to bring this breaking condition into the loop instead of using `true` there.

core/src/main/java/com/digitalpebble/stormcrawler/parse/TextExtractor.java
```diff
@@ -138,14 +143,71 @@ public void tail(Node node, int depth) {
...
-        String text = textNode.getWholeText();
+    /**
+     * Start a depth-first traverse of the root and all of its descendants.
+     *
+     * @param visitor Node visitor.
+     * @param root the root node point to traverse.
+     */
+    public static void traverse(
+            NodeVisitor visitor, Node root, int maxSize, StringBuilder builder) {
+        Validate.notNull(visitor);
```
rzo1, line 159: We could use `Validate.notNull(Object o, String msg)` to add some context to the `ValidationException` ?

rzo1 (approved):

core/src/main/java/com/digitalpebble/stormcrawler/parse/TextExtractor.java
jnioche: pretty harmless, could leave it just in case someone specifies -ea ;-)

jnioche:
thanks for reviewing @rzo1, have implemented your suggestions
one thing I really dislike with the current implementation is that we check the parents nodes in _preserveWhitespace_  over and over when in fact we are visiting the graph from the parents and could work this information just once and for all.

PR #1040, opened 2023-02-14
okhttp.httpprotocol : Replace User Agent if specified into Metadata

Mikwiss:
Hi !

We need to specify a User-Agent by request (set into a KafkaSpout). We tried to use `protocol.set-headers`, but the protocol don't replace the previous User Agent :

```json
"protocol%2E_request%2Eheaders_": 
[
"GET /index.html HTTP/2 User-Agent: defaultUA User-Agent: myUA"
]
```

So we create a new metadata `protocol.set-user-agent` in order to achieve it.

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

Thanks!

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
@@ -71,6 +71,10 @@ public class HttpProtocol extends AbstractHttpProtocol {
 
     private final MediaType JSON = MediaType.parse("application/json; charset=utf-8");
 
+    private static final String HEADER_USER_AGENT = "User-Agent";
+
+    protected static final String SET_USER_AGENT_BY_REQUEST = "set-user-agent";
+
     private OkHttpClient client;
 
     private int globalMaxContent;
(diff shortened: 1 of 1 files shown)


rzo1:
Wouldn't it be sufficient to only change from `private` to `protected` to allow overriding it in a custom impl of the protocol? Don't know, if setting the user agent on a per-request basis is a common use-case?

core/src/main/java/com/digitalpebble/stormcrawler/protocol/okhttp/HttpProtocol.java
```diff
@@ -267,6 +271,13 @@ private void addHeadersToRequest(Builder rb, Metadata md) {
                 rb.addHeader(h.getKey(), h.getValue());
             }
         }
+
+        final String agentName = md.getFirstValue(SET_USER_AGENT_BY_REQUEST, protocolMDprefix);
```
jnioche, line 275: what about having the logic in the _AbstractHTTPProtocol_ class and create a new method 
_String checkUAFromMD(Metadata md)_
?
If it returns non-null use it 

We can then have the same functionality in the other protocol implementations. I'd keep _"http.agent"_ as the key to look for in the metadata, put it in a static final String and use it in 

https://github.com/DigitalPebble/storm-crawler/blob/9b406a282cb4b1c8dff07074045e47918cecd280/core/src/main/java/com/digitalpebble/stormcrawler/protocol/AbstractHttpProtocol.java#L142

jnioche (changes requested):

jnioche:
> Hi @rzo1, that's my second option ! It's enough for us indeed +1

I'd be happy either way. We do override various things using the metadata, so it would not be shocking to do it for that as well, but then as @rzo1 pointed out, it is not something that ppl would routinely do

Issue #1043, opened 2023-02-26
urlfrontier.StatusUpdaterBolt fails after reconnecting the URL Frontier to the SC


michaeldinzinger:
Hello all,
the following issue popped up during some experimenting with the URL Frontier and a derivate of the SC: When the URL Frontier service is killed and restarted, the serving of new URLs to the StormCrawler via the urlfrontier.Spout continues to work nicely as if the Frontier service would never have been down.
However, the StatusUpdaterBolt seems to fail all tuples that it receives (after the restart, you can see no acks, but only fails for the StatusUpdaterBolt in the Storm UI). It seems that the cancelled connection between the Crawler and the frontier cannot be resumed after the restart of the URL Frontier.
As a consequence, after the Frontier service is restarted, also the SC topology has to be killed and resubmitted, which is not very convenient. Maybe a little code fix in the StatusUpdaterBolt would solve this issue.

jnioche:
Fixed in #1054

PR #1054, opened 2023-04-15
Issue #1043: Fixing problems after restart of Frontier service

michaeldinzinger:
Hello,

this PR should fix two problems which I faced when the crawler was communicating with an URLFrontier service, and this service was down and restarted. These problems do not appear in case the URLFrontier just keeps running without any complications. However, in case the URL Frontier goes down and has to be restarted, it means, as a consequence, that the StormCrawler has to be restarted because it is in a failed state.

The two code changes are therefore:
- Periodically cleaning up the waitAck cache in order to trigger the eviction of expired entires. Otherwise, it comes to a starvation as soon as the cache runs full and the waitAck cache has neither read nor write accesses, so that no eviction is triggered (but still the eviction of entries is necessary in order to release permits from the inFlightSemaphore)
The starvation looks like this:
```
13:41:24.635 [Time-limited test] WARN  c.d.s.u.StatusUpdaterBolt - Waiting more than 30000 ms for processing. There are 0 permits available for 0 waiting threads.
13:41:24.645 [Time-limited test] WARN  c.d.s.u.StatusUpdaterBolt - Waiting more than 30010 ms for processing. There are 0 permits available for 0 waiting threads.
13:41:24.656 [Time-limited test] WARN  c.d.s.u.StatusUpdaterBolt - Waiting more than 30020 ms for processing. There are 0 permits available for 0 waiting threads.
13:41:24.666 [Time-limited test] WARN  c.d.s.u.StatusUpdaterBolt - Waiting more than 30030 ms for processing. There are 0 permits available for 0 waiting threads.
13:41:24.677 [Time-limited test] WARN  c.d.s.u.StatusUpdaterBolt - Waiting more than 30040 ms for processing. There are 0 permits available for 0 waiting threads.
13:41:24.687 [Time-limited test] WARN  c.d.s.u.StatusUpdaterBolt - Waiting more than 30050 ms for processing. There are 0 permits available for 0 waiting threads.
13:41:24.698 [Time-limited test] WARN  c.d.s.u.StatusUpdaterBolt - Waiting more than 30060 ms for processing. There are 0 permits available for 0 waiting threads.
13:41:24.709 [Time-limited test] WARN  c.d.s.u.StatusUpdaterBolt - Waiting more than 30070 ms for processing. There are 0 permits available for 0 waiting threads.
13:41:24.719 [Time-limited test] WARN  c.d.s.u.StatusUpdaterBolt - Waiting more than 30080 ms for processing. There are 0 permits available for 0 waiting threads.
13:41:24.730 [Time-limited test] WARN  c.d.s.u.StatusUpdaterBolt - Waiting more than 30090 ms for processing. There are 0 permits available for 0 waiting threads.
13:41:24.740 [Time-limited test] WARN  c.d.s.u.StatusUpdaterBolt - Waiting more than 30100 ms for processing. There are 0 permits available for 0 waiting threads.
13:41:24.751 [Time-limited test] WARN  c.d.s.u.StatusUpdaterBolt - Waiting more than 30110 ms for processing. There are 0 permits available for 0 waiting threads.
13:41:24.761 [Time-limited test] WARN  c.d.s.u.StatusUpdaterBolt - Waiting more than 30120 ms for processing. There are 0 permits available for 0 waiting threads.
13:41:24.772 [Time-limited test] WARN  c.d.s.u.StatusUpdaterBolt - Waiting more than 30130 ms for processing. There are 0 permits available for 0 waiting threads.
13:41:24.782 [Time-limited test] WARN  c.d.s.u.StatusUpdaterBolt - Waiting more than 30140 ms for processing. There are 0 permits available for 0 waiting threads.
```
- Implementing an additional thread which periodically tries to initialize the connection to the URLFrontier service. This is necessary because the restarted Frontier would not receive a new PutURL call to start up the new communication
- Another code change was that I had to skip the code format validation. After a wearisome search for the mistake, I just couldn't find out the reason why the StatusUpdaterBoltTest.java class is not correctly formatted. So I'm happy about any comment

### DIFF
--- a/external/urlfrontier/pom.xml
+++ b/external/urlfrontier/pom.xml
@@ -40,13 +40,6 @@
 					</systemPropertyVariables>
 				</configuration>
 			</plugin>
-			<plugin>
-				<groupId>com.cosium.code</groupId>
-				<artifactId>git-code-format-maven-plugin</artifactId>
-				<configuration>
-					<skip>true</skip>
-				</configuration>
-			</plugin>
 		</plugins>
 	</build>
 
--- a/external/urlfrontier/pom.xml
+++ b/external/urlfrontier/pom.xml
@@ -40,6 +40,13 @@
 					</systemPropertyVariables>
 				</configuration>
 			</plugin>
+			<plugin>
+				<groupId>com.cosium.code</groupId>
+				<artifactId>git-code-format-maven-plugin</artifactId>
+				<configuration>
+					<skip>true</skip>
+				</configuration>
+			</plugin>
 		</plugins>
 	</build>
 
--- a/external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBolt.java
+++ b/external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBolt.java
@@ -117,7 +117,7 @@ public class StatusUpdaterBolt extends AbstractStatusUpdaterBolt
                 ConfUtils.getInt(
                         stormConf, URLFRONTIER_UPDATER_MAX_MESSAGES_KEY, maxMessagesInFlight);
 
-        LOG.info("Allowing up to {} message in flight", maxMessagesInFlight);
+        LOG.info("Allowing up to {} message(s) in flight", maxMessagesInFlight);
 
         // Fairness not necessary, we are not in a hurry, as long as we may be processed at some
         // point.
@@ -212,11 +212,8 @@ public class StatusUpdaterBolt extends AbstractStatusUpdaterBolt
             values = waitAck.getIfPresent(url);
             if (values != null) {
                 // Invalidate before releasing permits to protect from new entries for this URL
-                // until
-                // permits are handed out.
-                // Invalidate removes the key url from waitAck, therefore it is safe to use values
-                // without
-                // lock at this point.
+                // until permits are handed out. Invalidate removes the key url from waitAck,
+                // therefore it is safe to use values without lock at this point.
                 waitAck.invalidate(url);
             }
         } finally {
@@ -313,6 +310,12 @@ public class StatusUpdaterBolt extends AbstractStatusUpdaterBolt
                                 inFlightSemaphore.availablePermits(),
                                 inFlightSemaphore.getQueueLength());
                     }
+                    // To prevent a deadlock, it is necessary to periodically clean up the waitAck
+                    // cache. Otherwise, in case of a frontier-side or connection-wise error, all
+                    // incoming URLs will after some time be all caught up in this loop without
+                    // touching the cache, possibly leading to no eviction and thus leading to no
+                    // release of inFlightSemaphore permits.
+                    waitAck.cleanUp();
                 }
             } catch (InterruptedException e) {
                 LOG.warn(
--- a/external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBolt.java
+++ b/external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBolt.java
@@ -178,7 +178,8 @@ public class StatusUpdaterBolt extends AbstractStatusUpdaterBolt
         }
 
         channel = ManagedChannelUtil.createChannel(address);
-        channel.notifyWhenStateChanged(ConnectivityState.SHUTDOWN, () -> onChannelStateChange(ConnectivityState.SHUTDOWN));
+        channel.notifyWhenStateChanged(
+                ConnectivityState.SHUTDOWN, () -> onChannelStateChange(ConnectivityState.SHUTDOWN));
         URLFrontierStub frontier = URLFrontierGrpc.newStub(channel).withWaitForReady();
         requestObserver = frontier.putURLs(this);
     }
--- a/external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBolt.java
+++ b/external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBolt.java
@@ -36,6 +36,7 @@ import crawlercommons.urlfrontier.Urlfrontier.StringList;
 import crawlercommons.urlfrontier.Urlfrontier.StringList.Builder;
 import crawlercommons.urlfrontier.Urlfrontier.URLInfo;
 import crawlercommons.urlfrontier.Urlfrontier.URLItem;
+import io.grpc.ConnectivityState;
 import io.grpc.ManagedChannel;
 import io.grpc.stub.StreamObserver;
 import java.util.Collections;
@@ -45,8 +46,6 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantLock;
@@ -62,7 +61,6 @@ import org.slf4j.LoggerFactory;
 
 public class StatusUpdaterBolt extends AbstractStatusUpdaterBolt
         implements RemovalListener<String, List<Tuple>>,
-                Runnable,
                 StreamObserver<crawlercommons.urlfrontier.Urlfrontier.AckMessage> {
 
     private static final Logger LOG = LoggerFactory.getLogger(StatusUpdaterBolt.class);
@@ -87,8 +85,6 @@ public class StatusUpdaterBolt extends AbstractStatusUpdaterBolt
 
     /** Globally set crawlID * */
     private String globalCrawlID;
-
-    private String address;
 
     @Override
     public void prepare(
@@ -132,7 +128,7 @@ public class StatusUpdaterBolt extends AbstractStatusUpdaterBolt
         List<String> addresses = ConfUtils.loadListFromConf(URLFRONTIER_ADDRESS_KEY, stormConf);
 
         // Selected address
-        // String address;
+        String address;
         switch (addresses.size()) {
             case 0:
                 LOG.debug("{} has no addresses.", URLFRONTIER_ADDRESS_KEY);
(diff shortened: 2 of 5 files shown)


external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBolt.java
```diff
@@ -60,6 +62,7 @@
 
 public class StatusUpdaterBolt extends AbstractStatusUpdaterBolt
         implements RemovalListener<String, List<Tuple>>,
+                Runnable,
```
rzo1, line 65: Re-using a Storm bolt to be a `Runnable` sounds odd to me :-) personally, I would prefer a inner class or sth like that. Interested to hear @jnioche here.

external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBolt.java
```diff
@@ -170,16 +175,27 @@ public void prepare(
...
+                ConfUtils.getInt(stormConf, URLFRONTIER_CONNECTION_CHECKER_INTERVAL, 60);
+
+        // Create a periodic ExecutorService to check if the connection to the URLFrontier is
+        // still alive. If not, try to reconnect.
+        ScheduledExecutorService executorService = Executors.newScheduledThreadPool(1);
+        executorService.scheduleAtFixedRate(this, 1, connectionCheckerInterval, TimeUnit.SECONDS);
+    }
+
+    public void initFrontierConnection() {
+        channel = ManagedChannelUtil.createChannel(address);
```
rzo1, line 198: Do we really need to re-create the managed channel? 

I was under the impression, that `io.grpc.ManagedChannel` is capable of handling a re-connect after a connection loss or error? (as stated in https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md or discussed in https://groups.google.com/g/grpc-io/c/P1Qr6e0I1yA )

external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBolt.java
```diff
@@ -127,7 +132,7 @@ public void prepare(
         List<String> addresses = ConfUtils.loadListFromConf(URLFRONTIER_ADDRESS_KEY, stormConf);
 
         // Selected address
-        String address;
+        // String address;
```
rzo1, line 135: If it is not needed anymore, we should remove it.

external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBolt.java
jnioche: first, thanks @michaeldinzinger for the PR and @rzo1 for the review.
I agree that this feels wrong and given how Storm handles bolts in a multithreaded fashion, could there be side effects?

external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBolt.java
jnioche: Same, this was my assumption

external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBolt.java
jnioche: I  am probably misreading the code but isn't it used below?

jnioche:
Would there be a way of reproducing the issue in a JUnit test? The existing tests spin a Docker container, it would be possible to stop and restart it as part of the test.
I did try to reproduce the issue manually a while ago but I don't think I managed to.

external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBolt.java
rzo1: It became a class variable (so no local variable needed anymore).

jnioche: Thanks! Spotted it since. Need another coffee

jnioche:
Judging by https://github.com/grpc/grpc-java/issues/8177 it is not necessary to recreate the channel, only the streamObservers

external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBolt.java
michaeldinzinger: Thank you very much for your comments. I worked it over and omitted the re-creation of a new channel. Initially, I thought it would be necessary as it is hard to grasp when the automatic re-connection takes place and it leaded to some test scenarios, when for no obvious reason sometimes it worked and sometimes it didn't, depending on whether the channel has already been automatically re-connected or not

external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBolt.java
michaeldinzinger: Thanks, I'll avoid these:)

jnioche:
@michaeldinzinger I have made some changes to your PR, would it be easier if I pushed these in a separate PR and we take it as a continuation of the discussion or would you rather rework the one you have started here?

michaeldinzinger:
> I have made some changes to your PR, would it be easier if I pushed these in a separate PR and we take it as a continuation of the discussion or would you rather rework the one you have started here?

Hello Julien, sorry, I haven't seen your message as I was also just working on it and just pushed a change to the PR. You might want to have a look and, in case your and my code modifications diverge, you could modify it here in this PR

external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBolt.java
```diff
@@ -170,20 +171,28 @@ public void prepare(
...
+        channel = ManagedChannelUtil.createChannel(address);
+        channel.notifyWhenStateChanged(ConnectivityState.SHUTDOWN, () -> onChannelStateChange(ConnectivityState.SHUTDOWN));
         URLFrontierStub frontier = URLFrontierGrpc.newStub(channel).withWaitForReady();
         requestObserver = frontier.putURLs(this);
     }
 
+    private void onChannelStateChange(ConnectivityState state) {
+        ConnectivityState newState = channel.getState(true);
+        LOG.debug("Channel state changed from {} to {}", state, newState);
+        if (state == ConnectivityState.TRANSIENT_FAILURE) {
```
jnioche, line 189: equals() ?

external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBolt.java
```diff
@@ -170,20 +171,28 @@ public void prepare(
...
-        } else {
-            channel = ManagedChannelUtil.createChannel(address);
+            String host =
+                    ConfUtils.getString(stormConf, URLFRONTIER_HOST_KEY, URLFRONTIER_DEFAULT_HOST);
+            int port = ConfUtils.getInt(stormConf, URLFRONTIER_PORT_KEY, URLFRONTIER_DEFAULT_PORT);
+            address = host + ":" + port;
         }
 
+        channel = ManagedChannelUtil.createChannel(address);
+        channel.notifyWhenStateChanged(ConnectivityState.SHUTDOWN, () -> onChannelStateChange(ConnectivityState.SHUTDOWN));
```
jnioche, line 181: this is an elegant way of dealing with it. Have you tried this on a crawl?

external/urlfrontier/src/test/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBoltTest.java
```diff
@@ -87,6 +88,9 @@ public void before() {
                 "scheduler.class", "com.digitalpebble.stormcrawler.persistence.DefaultScheduler");
         config.put("status.updater.cache.spec", "maximumSize=10000,expireAfterAccess=1h");
         config.put("metadata.persist", persistedKey);
+        config.put("urlfrontier.connection.checker.interval", 2);
```
jnioche, line 91: this config has been removed

jnioche (changes requested):
a few comments for now, please look at the existing ones as well. Will have a closer look on Monday. Thanks!

external/urlfrontier/src/test/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBoltTest.java
```diff
@@ -99,37 +104,109 @@ public void after() {
...
+        }
+
+        url = "http://example.com/?test=3";
+        store(url, Status.DISCOVERED, new Metadata());
+        Assert.assertEquals(true, isAcked(url, 1));
     }
 
     @Test
-    public void canAckASimpleTuple()
+    public void exceedingMaxMessagesInFlight()
```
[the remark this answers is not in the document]
michaeldinzinger: Thank you again very much for your comments
Yep, with the last commit I made these two tests to one

external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBolt.java
michaeldinzinger: == seems to be equally good ([StackOverflow](https://stackoverflow.com/questions/1750435/comparing-java-enum-members-or-equals)). Nevertheless, I can change it if wished

external/urlfrontier/src/main/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBolt.java
michaeldinzinger: I did one test crawl, in which I was in particular focusing on whether the crawler is able to continue as usual, after stopping the frontier for some time and restarting it then again. This work for me. Within the real crawling setup, the `notifyWhenStateChanged` callback is necessary and the restarting wouldn't have worked without using the callback. Remarkably, not so in the simpler test environment.

external/urlfrontier/src/test/java/com/digitalpebble/stormcrawler/urlfrontier/StatusUpdaterBoltTest.java
michaeldinzinger: I will remove this line

external/urlfrontier/pom.xml
```diff
@@ -40,6 +40,13 @@
 					</systemPropertyVariables>
 				</configuration>
 			</plugin>
+			<plugin>
+				<groupId>com.cosium.code</groupId>
```
jnioche, line 44: why do you need to add this?

jnioche (changes requested):
looks like we are nearly there but could you please look at the remaining few comments? Thanks @michaeldinzinger

michaeldinzinger:
I hope I didn't miss any comment. In case there are further remarks to work on, please let me know

jnioche (approved):
That's great thanks @michaeldinzinger

Issue #1042, opened 2023-02-20
Adapting rules for parsing robots.txt file


michaeldinzinger:
Hello all,
while crawling, we ran into a politeness issue and we suppose that its cause is that there was apparently a Connection Timeout when trying to fetch the `robots.txt`. We suppose that as a consequence the other webpages for this host were crawled without any restriction, just as if there would have been a 404 on the `robots.txt`.

As far as I see, is the logic on parsing the robots.txt file implemented as follows:
```
            if (code == 200) // found rules: parse them
            {
                String ct = response.getMetadata().getFirstValue(HttpHeaders.CONTENT_TYPE);
                robotRules = parseRules(url.toString(), response.getContent(), ct, agentNames);
            } else if ((code == 403) && (!allowForbidden)) {
                robotRules = FORBID_ALL_RULES; // use forbid all
            } else if (code >= 500) {
                cacheRule = false;
                robotRules = EMPTY_RULES;
            } else robotRules = EMPTY_RULES; // use default rules
```
In HttpRobotRulesParser.java; line 168-177

More suitable would be a logic as it is described here: https://support.google.com/webmasters/answer/9679690#robots_details
To differentiate between the cases:
- 403/404/410: For the crawling of this website is no robots.txt necessary
- 429/5xx: Connection Error results in no crawling of this website

Please tell me your thoughts on this

rzo1:
Sounds valid to apply `FORBID_ALL_RULES`, if we encounter "too many requests" or a http 5xx. 

However, I can also think of usecases in which you still want to apply `EMPTY_RULES` in such a case (or just stop being polite at all) ;-)

Maybe we can adjust the default here and add a configuration option to just ignore it (similar to `http.robots.403.allow`) ?

michaeldinzinger:
> Maybe we can adjust the default here and add a configuration option to just ignore it (similar to `http.robots.403.allow`) ?

Thank you, sounds good:)
Maybe something like `http.robots.connectionerror.skip` or `http.robots.5xx.allow` which defaults to false.
The code could then look like this
```
            robotRules = FORBID_ALL_RULES; // use default rules
            if (code == 200) // found rules: parse them
            {
                String ct = response.getMetadata().getFirstValue(HttpHeaders.CONTENT_TYPE);
                robotRules = parseRules(url.toString(), response.getContent(), ct, agentNames);
            } else if (code == 403 && allowForbidden) {
                robotRules = EMPTY_RULES; // allow all
[...]

PR #1055, opened 2023-04-15
Issue #1042: Adapt parsing of robots.txt files

michaeldinzinger:
Modified implemention of Robots.txt parsing as described in Issue #1042
Motivation was the following: In case the web server answers with an server error, according to the IETF RFC 9309 the website should not be crawled (except you encounter these Server errors over a long time such that you can assume that there is no Robots.txt file). For now, it means that forbidding all rules becomes the default and by setting a new configuration parameter `http.robots.5xx.allow` it is possible to bypass the default and assume the crawling allowance in case of server errors (the current SC behaviour)


sebastian-nagel:
+1 looks good!

> For now, it means that forbidding all rules becomes the default

Yes, it's an improvement for now - until there is a way to throttle or block new "fetch items" arriving (see discussion in #1042). Because the forbid-all rule is kept in the error cache, fetches for the given host are blocked for some reasonable time (error cache default config: expireAfterWrite = 1h).

core/src/main/java/com/digitalpebble/stormcrawler/protocol/HttpRobotRulesParser.java
```diff
@@ -165,16 +168,20 @@ public BaseRobotRules getRobotRulesSet(Protocol http, URL url) {
                             response.getContent() != null ? response.getContent().length : 0);
                 }
             }
-            if (code == 200) // found rules: parse them
-            {
+
+            // Parsing found rules; by default, all robots are forbidden (RFC 9309)
+            robotRules = FORBID_ALL_RULES;
```
sebastian-nagel, line 173: The default for all HTTP response codes not handled explicitly is now forbid-all instead of allow-all. Are there codes except 200, 403, 5xx, redirects which need a special treatment? Maybe not.

sebastian-nagel (approved):

core/src/main/java/com/digitalpebble/stormcrawler/protocol/HttpRobotRulesParser.java
sebastian-nagel: > Maybe not.

Sorry, I was wrong. In case of a HTTP 404 (robots.txt not found), crawling is allowed.

In doubt, maybe do not swap the defaults? "In the wild" the crawler may see any HTTP status code.

sebastian-nagel (changes requested):

core/src/main/java/com/digitalpebble/stormcrawler/protocol/HttpRobotRulesParser.java
michaeldinzinger: Oh yes, how could I forget about status code 404?! After 200 and redirects, it is probably what the crawler encounters most often when fetching the robots.txt.
In the RFC 9309, it is written that for all status codes in the range of 400-499 the robots file may be considered as "unavailable" and therefore allow-all. This is also true for 300-399 in case the crawler followed at least five consecutive redirects.

This implementation might be better:

```
// TODO: Follow up to five redirections (at the moment, it is only one)

if (200 <= code && code <= 299) {
	String ct = response.getMetadata().getFirstValue(HttpHeaders.CONTENT_TYPE);
	robotRules = parseRules(url.toString(), response.getContent(), ct, agentNames);
} else if (300 <= code && code <= 499) {
	robotRules = EMPTY_RULES; // allow all
	if (code == 403 && !allowForbidden) {
		robotRules = FORBID_ALL_RULES; // forbid all
	}
	// E.g. Google handles Too many requests similar to a server error
	// https://support.google.com/webmasters/answer/9679690#robots_details
	if (code == 429) {
		robotRules = FORBID_ALL_RULES; // forbid all
	}
} else if (500 <= code) {
	cacheRule = false;
	robotRules = FORBID_ALL_RULES; // forbid all
	if (allow5xx) {
		robotRules = EMPTY_RULES; // allow all
	}
} else {
	robotRules = EMPTY_RULES; // allow all
}
```
I'm not sure whether e.g. `if (200 <= code && code <= 299)` or `if (code == 200)` is more correct?
In case of `if (300 <= code && code <= 499)` and `if (code == 301 || code == 302 || code == 403 || code == 404 || code == 410)` it is not important because both options would result in allow-all

sebastian-nagel: > After 200 and redirects, it is probably what the crawler encounters most often when fetching the robots.txt

Yes. In the long tail the crawler gets almost every possible HTTP status code, see the attached [robotstxt-cc-main-2023-14.txt](https://github.com/DigitalPebble/storm-crawler/files/11247290/robotstxt-cc-main-2023-14.txt).

That's why I'd be careful about `if (200 <= code && code <= 299)` or `else if (500 <= code)`.

> In the RFC 9309, it is written that for all status codes in the range of 400-499 the robots file may be considered as "unavailable" and therefore allow-all.

Yes. The 403 (and 401) handling is defined in the "original" [norobots-RFC](https://www.robotstxt.org/norobots-rfc.txt). So, there are some differences - you mentioned the 429 handled by Google's crawler.

> // TODO: Follow up to five redirections (at the moment, it is only one)

Good catch!

Maybe keep this in a separate issue / PR to limit the amount of changes for an easy review. A unit test would be also useful to ensure that no change accidentally breaks something.

michaeldinzinger: > Yes. In the long tail the crawler gets almost every possible HTTP status code, see the attached [robotstxt-cc-main-2023-14.txt](https://github.com/DigitalPebble/storm-crawler/files/11247290/robotstxt-cc-main-2023-14.txt).

Wow, impressing numbers. Thanks for sharing. While crawling, you'll always have to be prepared to the unexpected, this was not clear to me in this extent.
As a consequence, the condition `else if (500 <= code)` is definitely wrong and has to be substituted with `else if (500 <= code && code <= 599)` in order to implement more precisely what is written in the RFC. Otherwise, the crawler would restrict itself more than necessary.

Besides that, there are probably two approaches: First, stick with the old way to go and parse the Robots.txt file only for Status code 200. In case the Status code is weird and unexpected, such as 204, 711 or 1, the crawler assumes the default ALLOW_ALL.
Or second, parse the Robots.txt file for any Status code except 300-599. This is probably a bit closer to the concrete wording in the RFC, in which no status code range is mentioned, as it only speaks of a "successful download".
```
2.3.1.1. Successful Access
If the crawler successfully downloads the robots.txt file, the
crawler MUST follow the parseable rules.
```
I think, in this case, the crawler wouldn't restrict itself too much either, as the parseContent() method by crawlercommons.robots.SimpleRobotRulesParser, which is furtheron called, also assumes ALLOW_ALL as default for an empty or unparsable Robots.txt file etc.

jnioche:
@michaeldinzinger I see that you've pushed a commit since. is the PR ready to be reviewed or does it need more work?
if so @sebastian-nagel since you started looking at this, any chance you could review this PR again?
thanks to you both

jnioche:
> With the last commit, I have implemented the small code change that I mentioned two weeks ago. I think it should be done now

Fab thanks. @sebastian-nagel would you be able to review this PR? You've looked at it more than I have.
I'll have a quick look now

core/src/main/java/com/digitalpebble/stormcrawler/protocol/HttpRobotRulesParser.java
```diff
@@ -165,16 +168,31 @@ public BaseRobotRules getRobotRulesSet(Protocol http, URL url) {
...
+                if (code == 403 && !allowForbidden) {
+                    robotRules = FORBID_ALL_RULES; // forbid all
+                }
+                // E.g. Google handles Too many requests similar to a server error
+                // https://support.google.com/webmasters/answer/9679690#robots_details
+                if (code == 429) {
+                    cacheRule = false;
+                    robotRules = FORBID_ALL_RULES; // forbid all
+                }
+            } else if (500 <= code && code <= 599) {
```
jnioche, line 187: same as above, the reversed logic is a bit confusing

jnioche (changes requested):

core/src/main/java/com/digitalpebble/stormcrawler/protocol/HttpRobotRulesParser.java
```diff
@@ -165,16 +168,31 @@ public BaseRobotRules getRobotRulesSet(Protocol http, URL url) {
...
-            {
+
+            // Parsing found rules according to RFC 9309
+            if (code == 200) {
                 String ct = response.getMetadata().getFirstValue(HttpHeaders.CONTENT_TYPE);
                 robotRules = parseRules(url.toString(), response.getContent(), ct, agentNames);
-            } else if ((code == 403) && (!allowForbidden)) {
-                robotRules = FORBID_ALL_RULES; // use forbid all
-            } else if (code >= 500) {
+            } else if (300 <= code && code <= 499) {
```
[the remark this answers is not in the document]
michaeldinzinger: For me, it was more intuitive to have all `<=` to point in one direction, but not a big deal, I'll change it

sebastian-nagel (approved):
Looks good! - need to resolve conflicts with changes to increase the number of followed redirects.

sebastian-nagel:
@michaeldinzinger - to resolve the merge conflicts, could you try to rebase the branch to the current master?

jnioche:
thanks @michaeldinzinger and @sebastian-nagel for the review

PR #1056, opened 2023-04-17
Change HttpProtocol to defer to configured values for retryOnConnectionFailure and followRedirects

ndtreviv:
In this version we defer to the configuration for the okhttp client settings on:
```
.retryOnConnectionFailure
.followRedirects
```

The rationale behind this is that a lot of sites are increasingly introducing DDos protection strategies and bot checks. For example,  [DDos Guard](https://ddos-guard.net/en/terminology/attack_type/http-challenge) is a popular one. One such check is the "HTTP Challenge" which issues a simple 302 response to the first request. If the requestor is a bot engaged in denial of service activities it is expected to just ignore it and end another request.
Simply by following redirects we can overcome this.

Disclaimer: I haven't squashed my commits because you can do that on merge these days, which is much easier.


rzo1 (approved):

jnioche:
Thanks @ndtreviv for the PR and @rzo1 for the review

PR #1057, opened 2023-04-18
Cache redirected robots.txt for target host only if path is /robots.txt and query is empty

sebastian-nagel:
Get/put redirected robots.txt from/to cache of target host only if the URL path including the query part is `/robots.txt`. The current code does not take the URL query part into consideration. In some (likely rare) cases, this may cause that the wrong robots.txt, e.g. `/robots.txt?from=source.example.com` is cached.

Note: the host port (and protocol) are part of the cache key. If the redirect leads to a different port on the target host, the robots.txt received from a request on this port is cached separately.


jnioche:
thanks @sebastian-nagel

Issue #1059, opened 2023-05-01
BasicURLNormalizer .unmangleQueryString() returns invalid results if "&" symbol in a parents path 


syefimov:
```
    private URLFilter createNormolizeFilter() {
        ObjectNode filterParams = new ObjectNode(JsonNodeFactory.instance);
        filterParams.put("unmangleQueryString", Boolean.valueOf(true));
        return createFilter(filterParams);
    }

    private void testNormolizeFilter()  throws MalformedURLException {
         URLFilter urlFilter = createNormolizeFilter();
         URL url = new URL("http://eadiv.state.wy.us/s&utax/NEWSRLSE_fy18.pdf");
         String filterResult = urlFilter.filter(url, metadata, url.toExternalForm());
         Assert.assertEquals("http://eadiv.state.wy.us/s&utax/NEWSRLSE_fy18.pdf", filterResult);
   }
```

jnioche:
Ideally the & should have been URL encoded btu we should make the normalizers as robust as possible.
@syefimov thanks for reporting it, do you want to suggest a fix for it?

syefimov:
Yes, analyze only string after last "/". If it is OK I can create pull request.

PR #1062, opened 2023-05-02
BasicURLNormalizer .unmangleQueryString() returns invalid results if "&" symbol in a parents path #1059

syefimov:
BasicURLNormalizer .unmangleQueryString() returns invalid results if "&" symbol in a parents path #1059

Analyze only last item in a URL path. Test added.

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/filtering/basic/BasicURLNormalizer.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/filtering/basic/BasicURLNormalizer.java
@@ -303,11 +303,14 @@ public class BasicURLNormalizer extends URLFilter {
      * @return corrected url
      */
     private String unmangleQueryString(String urlToFilter) {
-        int firstAmp = urlToFilter.indexOf('&');
+        String[] pathElements = urlToFilter.split("/");
+        int firstAmp = pathElements[pathElements.length - 1].indexOf('&');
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/digitalpebble/stormcrawler/filtering/basic/BasicURLNormalizer.java
```diff
@@ -303,11 +303,14 @@ public int compare(NameValuePair p1, NameValuePair p2) {
      * @return corrected url
      */
     private String unmangleQueryString(String urlToFilter) {
-        int firstAmp = urlToFilter.indexOf('&');
+        String[] pathElements = urlToFilter.split("/");
+        int firstAmp = pathElements[pathElements.length - 1].indexOf('&');
```
rzo1, line 307: I don't recall the complete normalizer code, so the following might not happen (as other methods guard against it...)

If `pathElements` is empty, we will access index `-1`. While this case might not happen in production (assuming every encountered URL starts with a protocol), we should be robust against everything we encounter in the wild. Can we guard it with a `pathElements.length > 0` check?

rzo1:
Nevermind, need more coffee :-)

jnioche (changes requested):

jnioche (approved):

jnioche:
thanks @syefimov

Issue #1058, opened 2023-04-20
Increasing the number of redirects for Robots.txt fetching


michaeldinzinger:
Hello all, as suggested in PR #1055 this issue should address the following:
The [RFC 9309](https://datatracker.ietf.org/doc/pdf/rfc9309) concerning the Robots Exclusion Protocol states that a crawler should follow at least 5 redirects. This is the corresponding section:
```
2.3.1.2. Redirects
It's possible that a server responds to a robots.txt fetch request
with a redirect, such as HTTP 301 or HTTP 302 in the case of HTTP.
The crawlers SHOULD follow at least five consecutive redirects, even
across authorities (for example, hosts in the case of HTTP).
If a robots.txt file is reached within five consecutive redirects,
the robots.txt file MUST be fetched, parsed, and its rules followed
in the context of the initial authority.
If there are more than five consecutive redirects, crawlers MAY
assume that the robots.txt file is unavailable.
```
At the moment, the StormCrawler follows only one level of redirects. The corresponding implementation is in HttpRobotRulesParser.java, beginning from line 128.

PR #1074, opened 2023-05-19
Increase the number of redirects to 5 for Robots.txt fetching

michaeldinzinger:
Described in Issue #1058
In IETF RFC9309 (Robots Exclusion Protocol), it is stated that crawlers should follow up to 5 consecutive redirects in their attempt to fetch a Robots.txt file. Up to now, the SC only followed one level of redirects. So this code change might slightly improve the politeness of the crawler

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/protocol/HttpRobotRulesParser.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/protocol/HttpRobotRulesParser.java
@@ -37,6 +37,8 @@ public class HttpRobotRulesParser extends RobotRulesParser {
 
     protected Metadata fetchRobotsMd;
 
+    private static final int MAX_NUM_REDIRECTS = 5;
+
     HttpRobotRulesParser() {}
 
     public HttpRobotRulesParser(Config conf) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/digitalpebble/stormcrawler/protocol/HttpRobotRulesParser.java
```diff
@@ -122,13 +122,18 @@ public BaseRobotRules getRobotRulesSet(Protocol http, URL url) {
                     http.getProtocolOutput(new URL(url, "/robots.txt").toString(), fetchRobotsMd);
             int code = response.getStatusCode();
             bytesFetched.add(response.getContent() != null ? response.getContent().length : 0);
-            // try one level of redirection ?
-            if (code == 301 || code == 302 || code == 307 || code == 308) {
+
+            // According to RFC9309, the crawler should follow at least 5 consecutive redirects
+            // to get the robots.txt file.
+            int max_num_redirects = 5;
```
jnioche, line 128: make it a static final field instead?

jnioche (approved):
Thanks for adding the tests. Just a minor suggestion

core/src/main/java/com/digitalpebble/stormcrawler/protocol/HttpRobotRulesParser.java
rzo1: > make it a static final field instead?

+1 

(otherwise, I would argue for `maxNumRedirects` :-) )

jnioche:
thanks @michaeldinzinger and @rzo1

Issue #1101, opened 2023-09-29
Refactor configuration of timeouts in Selenium protocol


jnioche:
The configuration of the timeouts is quite error prone.  

[The default values are set to 0
](https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/resources/crawler-default.yaml#L174)
```
  selenium.implicitlyWait: 0
  selenium.pageLoadTimeout: 0
  selenium.scriptTimeout: 0
```
If not set explicitly in the configuration, the default value of 0 is used which leads to a script timeout. 

Instead we should it at -1 in the default config so that it is visible for illustration purposes and modify the code so that it is only set if the value is >=0. This would in effect mean: rely on Selenium's default unless specified otherwise, which feels a lot cleaner. It also adds an additional safety as negative values are not accepted.

We will also change the config to use a map, feels a bit cleaner conceptually

```
  # rely on selenium's default values
  # set to a value >= 0 to override
  selenium.timeouts:
    script: -1
    pageLoad: -1
    implicit: -1
```

PR #1102, opened 2023-09-29
refactoring timeouts Selenium

jnioche:
see #1101

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/protocol/selenium/RemoteDriverProtocol.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/protocol/selenium/RemoteDriverProtocol.java
@@ -68,6 +68,15 @@ public class RemoteDriverProtocol extends SeleniumProtocol {
 
         final boolean tracing = ConfUtils.getBoolean(conf, "selenium.tracing", false);
 
+        // TEMPORARY: draw attention to config change
+        for (String p : new String[] {"implicitlyWait", "pageLoadTimeout", "scriptTimeout"}) {
+            if (conf.containsKey("selenium." + p)) {
+                String message = "selenium." + p + " is deprecated. Please use selenium.timeouts!";
+                LOG.error(message);
+                throw new RuntimeException(message);
+            }
+        }
+
         for (String cdaddress : addresses) {
             try {
                 RemoteWebDriver driver =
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/digitalpebble/stormcrawler/protocol/selenium/RemoteDriverProtocol.java
```diff
@@ -72,13 +72,20 @@ public void configure(Config conf) {
             try {
                 RemoteWebDriver driver =
                         new RemoteWebDriver(new URL(cdaddress), capabilities, tracing);
+                // setting timouts
+                // see https://www.browserstack.com/guide/understanding-selenium-timeouts
                 Timeouts touts = driver.manage().timeouts();
-                int implicitWait = ConfUtils.getInt(conf, "selenium.implicitlyWait", 0);
```
rzo1, line 76: Shall we add a (temporary) logging for these old settings for a smooth transision?

Something like

```
selenium.implicitlyWait is deprecated. Please use selenium.timeouts!
```

and fail hard?

core/src/main/java/com/digitalpebble/stormcrawler/protocol/selenium/RemoteDriverProtocol.java
```diff
@@ -72,13 +72,20 @@ public void configure(Config conf) {
...
-                int scriptTimeout = ConfUtils.getInt(conf, "selenium.scriptTimeout", 0);
-                touts.implicitlyWait(Duration.ofMillis(implicitWait));
-                touts.pageLoadTimeout(Duration.ofMillis(pageLoadTimeout));
-                touts.scriptTimeout(Duration.ofMillis(scriptTimeout));
+                Map<String, Number> timeouts = (Map<String, Number>) conf.get("selenium.timeouts");
+                if (timeouts != null) {
+                    long implicitTimeout = timeouts.getOrDefault("implicit", -1).longValue();
+                    long pageLoadTimeout = timeouts.getOrDefault("pageLoad", -1).longValue();
+                    long scriptTimeout = timeouts.getOrDefault("script", -1).longValue();
+                    if (implicitTimeout != -1)
```
rzo1, line 83: I would add the `{ }` :-)

rzo1: Wonder if we want a (debug) log entry, that the value is unset and the default is used?

rzo1 (approved):

core/src/main/java/com/digitalpebble/stormcrawler/protocol/selenium/RemoteDriverProtocol.java
jnioche: > Wonder if we want a (debug) log entry, that the value is unset and the default is used?

we can't tell whether the value has been explicitly set to -1 or if the default set it this way. Let's not bother

PR #1103, opened 2023-10-02
Improvements and fixes to HttpRobotRulesParser when following redirects

sebastian-nagel:
- remove unsafe check for absolute URLs on redirects (HTTP Location header): a string starting with `http` is not necessarily a valid and resolvable URL. For safety, always resolve the redirect location using [java.net.URL(URL context, String spec)](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/net/URL.html#%3Cinit%3E(java.net.URL,java.lang.String))
- resolve relative redirects using the redirect source as base URL. Resolving with the original URL as base/context may lead to a wrong redirect target in a chain of redirects with different hosts/authorities.
- cache robot rules for all `/robots.txt` (if on default location) in a chain of redirects
- add unit test for the three points above

Note: this PR is a result of implementing the RFC 9309 redirect rules to Nutch, see [NUTCH-2990](https://issues.apache.org/jira/browse/NUTCH-2990) and apache/nutch#779. I deliberately took the implementation in StormCrawler (#1058/#1074) as a starting point.

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/protocol/HttpRobotRulesParser.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/protocol/HttpRobotRulesParser.java
@@ -20,9 +20,11 @@ import com.github.benmanes.caffeine.cache.Cache;
 import com.google.common.primitives.Ints;
 import crawlercommons.robots.BaseRobotRules;
 import java.net.URL;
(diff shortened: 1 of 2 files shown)


jnioche:
Thanks @sebastian-nagel

Issue #728, opened 2019-05-21
Use regular expressions for list of metadata to persist and transfer


jnioche:
This would allow defining generic patterns e.g. _xxx_s_ or _yyy_p_ so that any matching key gets store or persisted without having to list them in the config explicitly.

jnioche:
Overlaps with #728 which is probably sufficient as an approach

PR #1117, opened 2023-11-04
Issue #728: Adding asterisk for metadata transfer

michaeldinzinger:
Hello all

Goal:
When having a lot of metadata fields that all start with the same prefix, e.g. `parse.[...]`, one can simply write
```
metadata.persist:
- parse.*
```
instead of listing them all.

Besides that, making mdToTransfer, mdToPersistOnly, trackPath and trackDepth protected instead of private. This makes it easier to create a Custom Metadata Transfer class.

### DIFF
--- a/core/src/main/java/com/digitalpebble/stormcrawler/util/MetadataTransfer.java
+++ b/core/src/main/java/com/digitalpebble/stormcrawler/util/MetadataTransfer.java
@@ -65,13 +65,13 @@ public class MetadataTransfer {
     /** Metadata key name for tracking a non-default max depth */
     public static final String maxDepthKeyName = "max.depth";
 
-    private final Set<String> mdToTransfer = new HashSet<>();
+    protected final Set<String> mdToTransfer = new HashSet<>();
 
-    private final Set<String> mdToPersistOnly = new HashSet<>();
+    protected final Set<String> mdToPersistOnly = new HashSet<>();
 
-    private boolean trackPath = true;
+    protected boolean trackPath = true;
 
-    private boolean trackDepth = true;
+    protected boolean trackDepth = true;
 
     public static MetadataTransfer getInstance(Map<String, Object> conf) {
         String className = ConfUtils.getString(conf, metadataTransferClassParamName);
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/digitalpebble/stormcrawler/util/MetadataTransfer.java
```diff
@@ -156,12 +156,34 @@ public Metadata filter(Metadata metadata) {
...
+                    if (mdKey.startsWith(prefix)) {
+                        filterKeys.add(mdKey);
+                    }
+                }
+            } else {
+                filterKeys.add(key);
+            }
+
+            for (String filterKey : filterKeys) {
+                String[] values = metadata.getValues(filterKey);
```
jnioche, line 180: Could add a method _copy(Metadata target, String key, boolean allowNull)_ to Metadata. I am sure a similar operation is done elsewhere in the code and we might as well make it easier.

Instead of storing the keys in a set then iterate on it, could simply 


  ```
if (mdKey.startsWith(prefix)) {
                        metadata.copy(filtered_md, mdKey, false);
                    }
                }
            } else {
                        metadata.copy(filtered_md, key, false);
            }
```

File: core/src/main/java/com/digitalpebble/stormcrawler/util/MetadataTransfer.java [hunk 2]
```diff
@@ -156,12 +156,34 @@ public Metadata filter(Metadata metadata) {
...
     private Metadata _filter(Metadata metadata, Set<String> filter) {
         Metadata filtered_md = new Metadata();
+
+        Set<String> filterKeys = new HashSet<>();
         for (String key : filter) {
-            String[] vals = metadata.getValues(key);
-            if (vals != null) filtered_md.setValues(key, vals);
+            if (key.endsWith("*")) {
+                String prefix = key.substring(0, key.length() - 1);
+                for (String mdKey : metadata.keySet()) {
```
jnioche, line 170: Could add a new method to Metadata

_ Set<String> getKeysWithPrefix(String prefix)_

as this is certainly used elsewhere

and iterate on it e.g.

  for (String mdKey : metadata.getKeysWithPrefix(prefix)) {

jnioche (commented):
Thanks! this would be very helpful. See comments on possible improvements

core/src/main/java/com/digitalpebble/stormcrawler/util/MetadataTransfer.java
michaeldinzinger: Thank you for the hint. I named the new method `keySet(String prefix)` for keeping the same naming schema as `metadata.keySet()`. It is indeed also used in FetcherBolt and SimpleFetcherBolt.

core/src/main/java/com/digitalpebble/stormcrawler/util/MetadataTransfer.java
michaeldinzinger: In the new commit, I added `copy(targetMetadata, key)`. The `allowNull` flag is actually not necessary, I found out, because in `metadata.SetValues(...)` null values are ignored anyways. This copy method takes only a single key. Also having a copy method with a set of input keys would be possible.  
I didn't find any other similar operation elsewhere in the code.

jnioche:
That's fab, thanks @michaeldinzinger!

I think the change would affect both the metadata to transfer and the metadata to persist, in which case ideally we would like to have a test demonstrating the impact on the latter. We'd also need to make it explicit in the [default config](https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/resources/crawler-default.yaml#L54) as well as the configs generated by the archetypes.

michaeldinzinger:
Thank you again for your quick reply.

> I think the change would affect both the metadata to transfer and the metadata to persist, in which case ideally we would like to have a test demonstrating the impact on the latter.

How could such a demonstration test look like? For now, I extended the method `testTransfer` to cover the metadata.transfer scenario, whereas the method `testFilter` tests the metadata.persist scenario.

> We'd also need to make it explicit in the [default config](https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/resources/crawler-default.yaml#L54) as well as the configs generated by the archetypes.

I added a comment line in each of the `crawler-default.yaml` files.

PR #1126, opened 2023-11-22
Fix wrong most recent date was set

chhsiao90:
With the wrong potentialNewDate was assigned, it was always 1970-xx-xx.


rzo1 (approved):

jnioche:
thanks @chhsiao90

PR #1130, opened 2023-12-02
Glob field mapping for indexer.md.mapping

jnioche:
Make it easier to map entire lists of k/v sharing a prefix


core/src/main/java/com/digitalpebble/stormcrawler/indexing/AbstractIndexerBolt.java
```diff
@@ -175,26 +202,43 @@ protected Map<String, String[]> filterMetadata(Metadata meta) {
...
-            // check whether we want a specific value or all of them?
-            int index = key.index;
-            // want a value index that it outside the range given
-            if (index >= values.length) {
-                continue;
-            }
-            // store all values available
-            if (index == -1) {
-                fieldVals.put(entry.getValue(), values);
+        // for (Entry<Key, String> entry : metadata2field.entrySet()) {
```
rzo1, line 205: That line can be removed

Issue #1137, opened 2023-12-13
Dependency updates


jnioche:
mvn versions:display-dependency-updates 

[INFO]   org.opensearch.client:opensearch-rest-high-level-client ... 2.11.0 -> 2.11.1
[INFO]   com.amazonaws:aws-java-sdk-cloudsearch .......... 1.12.585 -> 1.12.616
[INFO]   com.amazonaws:aws-java-sdk-s3 ................... 1.12.585 -> 1.12.617
[INFO]   com.fasterxml.jackson.core:jackson-annotations ...... 2.15.2 -> 2.16.0
[INFO]   com.fasterxml.jackson.core:jackson-core ............. 2.15.2 -> 2.16.0
[INFO]   com.fasterxml.jackson.core:jackson-databind ......... 2.15.2 -> 2.16.0
[INFO]   com.ibm.icu:icu4j ....................................... 74.1 -> 74.2
~~[INFO]   com.squareup.okhttp3:okhttp ................. 4.12.0 -> 5.0.0-alpha.11~~
~~[INFO]   com.squareup.okhttp3:okhttp-brotli .......... 4.12.0 -> 5.0.0-alpha.11~~
[INFO]   javax.xml.bind:jaxb-api .................. 2.3.1 -> 2.4.0-b180830.0359
[INFO]   org.apache.logging.log4j:log4j-api ............. 2.20.0 -> 2.22.0
[INFO]   org.apache.logging.log4j:log4j-core ............ 2.20.0 -> 2.22.0
[INFO]   org.apache.logging.log4j:log4j-slf4j-impl ...... 2.20.0 -> 2.22.0
~~[INFO]   org.apache.tika:tika-core ........................ 2.9.1 -> 3.0.0-BETA~~
~~[INFO]   org.apache.tika:tika-parsers-standard-package .... 2.9.1 -> 3.0.0-BETA~~
[INFO]   org.jetbrains:annotations ........................... 24.0.1 -> 24.1.0
[[INFO]   org.jsoup:jsoup ..................................... 1.16.2 -> 1.17.1
](https://jsoup.org/news/release-1.17.1)[INFO]   org.mockito:mockito-core .............................. 5.7.0 -> 5.8.0
[INFO]   org.netpreserve:jwarc ............................... 0.28.3 -> 0.28.5
[INFO]   org.seleniumhq.selenium:selenium-chrome-driver ...... 4.15.0 -> 4.16.1
[...]

PR #1139, opened 2023-12-19
dependency upgrades

jnioche:
see #1137

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,11 +79,11 @@
 		<!-- dependency versions -->
 		<junit.version>4.13.2</junit.version>
 		<storm-client.version>2.6.0</storm-client.version>
-		<jackson.version>2.16.0</jackson.version>
+		<jackson.version>2.15.2</jackson.version>
 		<tika.version>2.9.1</tika.version>
 		<mockito.version>5.8.0</mockito.version>
 		<jetbrains.annotations.version>24.1.0</jetbrains.annotations.version>
-		<log4j2.version>2.22.0</log4j2.version>
+		<log4j2.version>2.21.1</log4j2.version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -79,13 +79,13 @@
 		<!-- dependency versions -->
 		<junit.version>4.13.2</junit.version>
 		<storm-client.version>2.6.0</storm-client.version>
-		<jackson.version>2.15.2</jackson.version>
```
jnioche, line 82: Given that Storm is on 2.15.2 - probably best to stay on that as well

pom.xml
```diff
@@ -79,13 +79,13 @@
...
+		<jackson.version>2.16.0</jackson.version>
 		<tika.version>2.9.1</tika.version>
-		<mockito.version>5.7.0</mockito.version>
-		<jetbrains.annotations.version>24.0.1</jetbrains.annotations.version>
-		<log4j2.version>2.20.0</log4j2.version>
+		<mockito.version>5.8.0</mockito.version>
+		<jetbrains.annotations.version>24.1.0</jetbrains.annotations.version>
+		<log4j2.version>2.22.0</log4j2.version>
```
jnioche: Storm is on 2.21.1 
https://github.com/apache/storm/blob/ae3a96e762095553311d9e335f7505c0b351d810/pom.xml#L109
stay on that

pom.xml
rzo1: 2.16.x should be compatible.

jnioche: I have had too many issues with incompatibilities in Jackson over the years e.g class not found etc...
Will just use whatever Storm uses

PR #1146, opened 2024-02-21
Generate THIRD-PARTY.txt file, fixes #1145

### LINKED ISSUE #1145: Generate THIRD-PARTY.txt file
`mvn license:aggregate-add-third-party -Dlicense.outputDirectory=.`

generates a THIRD-PARTY.txt file with all the direct and transitive dependencies. It would be good to have this done automatically as part of the build.

jnioche:
See #1145


rzo1:
Tests currently fail due to `/bin/bash: line 1: /dev/tcp/localhost/9200: Cannot assign requested address`

jnioche:
> Tests currently fail due to `/bin/bash: line 1: /dev/tcp/localhost/9200: Cannot assign requested address`

This is the OpenSearch module. There has been a new release of OpenSearch and since we rely on opensearch:latest, it might behave differently from what it used to do.

I will have a look and change it so that it relies on a specific version

jnioche:
Merged as the error is unrelated

Issue #1156, opened 2024-03-27
can we rename the 'master' branch as 'main'?


pjfanning:
Just feels a bit better name wise. 'main' is now the default in GitHub.

jnioche:
On my other repos I can see that this is done via _Settings -> General -> Default branches_ but now that it is at Apache, I don't have access to the Settings section.
Do we do it manually instead e.g. following https://www.git-tower.com/learn/git/faq/git-rename-master-to-main ?

pjfanning:
We might need to raise an INFRA issue. They have full access. If we get agreement to proceed, I can raise the JIRA.

rzo1:
Needs to be done bei INFRA ;-) (done that for a few ASF projects)

pjfanning:
I raised https://issues.apache.org/jira/browse/INFRA-25661

jnioche:
It has been done by infra.

PR #1161, opened 2024-03-27
Create .asf.yaml

pjfanning:
see https://cwiki.apache.org/confluence/display/INFRA/Git+-+.asf.yaml+features

ideally, we would do #1156 first - so we could change this to use main instead of master

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,0 +1,44 @@
+# https://cwiki.apache.org/confluence/display/INFRA/Git+-+.asf.yaml+features
+
+github:
+  description: "A scalable, mature and versatile web crawler based on Apache Storm"
+  homepage: https://github.com/apache/incubator-stormcrawler
+  labels:
+    - java
+    - crawler
+    - web-crawler
+    - distributed
+    - apache-storm
+    - stormcrawler
+
+  protected_tags:
+    - "v*.*.*"
+
+  dependabot_alerts:  true
+  dependabot_updates: false
+
+  features:
+    # Enable wiki for documentation
+    wiki: true
+    # Enable issue management
+    issues: true
+    # Enable projects for project management boards
+    projects: true
+
+  enabled_merge_buttons:
+    squash:  true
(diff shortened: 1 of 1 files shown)


.asf.yaml
```diff
@@ -0,0 +1,44 @@
...
+  protected_branches:
+    master:
+      required_pull_request_reviews:
+        dismiss_stale_reviews: false
+        require_code_owner_reviews: false
+        required_approving_review_count: 1
+
+notifications:
+  commits:              commits@stormcrawler.apache.org
+  issues:               dev@pekko.apache.org
```
jnioche, line 42: should be stormcrawler instead of pekko

jnioche (changes requested):

.asf.yaml
pjfanning: thanks - fixed

In Pekko, we have a separate notifications mailing list because these emails can overwhelm everything else on the dev list.

dave2wave: It is better to have issues and discussions remain on the dev@ list as long as possible.

jnioche:
@pjfanning is it in this file that the values for milestones are defined? The next one will probably be 3.0 (see discussion on dev@), how do we add it?

pjfanning:
@jnioche now that the main branch rename has happened, we should be ok to merge this - if it is ok

.asf.yaml
```diff
@@ -0,0 +1,44 @@
...
+    # Enable projects for project management boards
+    projects: true
+
+  enabled_merge_buttons:
+    squash:  true
+    merge:   false
+    rebase:  true
+
+  protected_branches:
+    master:
```
jnioche, line 34: shouldn't it be main now?

jnioche (changes requested):

.asf.yaml
pjfanning: fixed that now - thanks

jnioche:
thanks @pjfanning

PR #1165, opened 2024-03-28
Changed package names to org.apache 

### LINKED ISSUE #1157: Change package names from com.digitalpebble to org.apache
Now that SC is in incubation at the ASF, we need the package names to reflect the changes.

jnioche:
+ fixed references to DigitalPebble where possible

Fix #1157 

Getting a failing test but unrelated to these changes I think

[ERROR]   HttpRobotRulesParserTest.testRobotRulesParsing:113->parseRobotRules:78


rzo1 (approved):
We might need to remove other references to DigitalPebble later, i.e. ansilbe-storm, etc. - but it is ok for now, I guess.

PR #1182, opened 2024-04-02
Fix #1167 - Remove Elasticsearch module 

### LINKED ISSUE #1167: Deal with non-open source version of Elasticsearch
As mentioned in the incubation [proposal](https://cwiki.apache.org/confluence/display/INCUBATOR/StormCrawler+Proposal)

`Most dependencies are ASFv2 licensed. There are possible concerns around the Elasticsearch dependencies (Elastic License 2.0 - not open source). We will address these either by downgrading the version of Elasticsearch to 7.10.2, which is under ASFv2, remove the Elastic module altogether or make the dependency optional and instruct the users on how to activate it themselves to avoid having this non ASLv2 compliant licence in the distributions.`

rzo1:
Remove elasticsearch and update README to notify users, that a migration to opensearch is possible.


jnioche (approved):
Perfect, thanks!

PR #1186, opened 2024-04-03
Fix #1168 - Add a modified version of CONTRIBUTING.md

### LINKED ISSUE #1168: change CONTRIBUTING.md
The current file is out of date for an Apache podling.

I'm not sure what an updated file would say but it should probably mention that CLAs may be required for big changes.

https://www.apache.org/licenses/contributor-agreements.html

rzo1:
as the title says

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,66 +1,59 @@
-Contributing to StormCrawler
+Contributing to Apache StormCrawler
 ============================
+
+Thank you for your intention to contribute to Apache StormCrawler. As an open-source community, we highly appreciate contributions to our project.
+
+To make the process smooth for the project *committers* (those who review and accept changes) and *contributors* (those who propose new changes via pull requests), there are a few rules to follow.
+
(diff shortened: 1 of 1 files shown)


jnioche (approved):

CONTRIBUTING.md
```diff
@@ -1,66 +1,59 @@
-Contributing to StormCrawler
+Contributing to Apache StormCrawler
 ============================
 
+Thank you for your intention to contribute to Apache StormCrawler. As an open-source community, we highly appreciate external contributions to our project.
```
ayushtkn, line 4: do we need 'external' here, who is considered internal then?

ayushtkn (approved):
LGTM

CONTRIBUTING.md
rzo1: Let's drop external :-)

rzo1: Fixed @ayushtkn - thx for review!

PR #1203, opened 2024-04-30
Fix #1202 - Add release documentation and comply with source package naming requirements

### LINKED ISSUE #1202: Add Release Documentation
as the title says

rzo1:
Hi all,

can you do a careful check, if the RELEASE docs comply with the incubator requirements and are technically correct? 

I don't really now the specifics for the incubator, but should cover the content from https://cwiki.apache.org/confluence/display/INCUBATOR/Incubator+Release+Checklist 

Gruß
Richard

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,315 @@
+# Guide to release Apache StormCrawler (Incubating)
+
+## Release Preparation
+
+- Elect a release manager. This can also be done via lazy consensus on the mailing list.
+- Create an issue for a new release in https://github.com/apache/incubator-stormcrawler/issues  If you do not have permission to do so just ask to be given permissions on the mailing list.
+- Review all of the [issues](https://github.com/apache/incubator-stormcrawler/issues) associated with the release. All issues should be resolved and closed.
+- Any issues assigned to the release that are not complete should be assigned to the next release. Any critical or blocker issues should be resolved on the mailing list. Discuss any issues that you are unsure of on the mailing list.
+
+## Steps for the Release Manager
+
+The following steps need only to be performed once.
+
+- Make sure you have your PGP fingerprint added into https://id.apache.org/
+- Make sure you have your PGP keys password.
+- Add your PGP key to the [KEYS](https://dist.apache.org/repos/dist/release/incubator/stormcrawler/KEYS) file. 
+
+Examples of adding your key to this file:
+
+```
+pgp -kxa <your name> and append it to this file.
+(pgpk -ll <your name> && pgpk -xa <your name>) >> this file.
+(gpg --list-sigs <your name>
+&& gpg --armor --export <your name>) >> this file.
+```
+
+- In a local temp folder, svn checkout the StormCrawler artifacts and update the KEYS file
+- Note: This can only be done by an IPMC member. If you are a committer acting as a release manager, ask an IPMC member to add your key.
+
+```
+svn co https://dist.apache.org/repos/dist/release/incubator/stormcrawler
+svn commit -m "Added Key for <name>" KEYS
+```
+
+- Create a maven `settings.xml` to publish to `repository.apache.org`. An example configuration:
+
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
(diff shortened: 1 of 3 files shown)


RELEASING.md
```diff
@@ -0,0 +1,315 @@
+# Guide to release Apache StormCrawler (Incubating)
+
+## Release Preparation
+
+- Elect a release manager. This can also be done via lazy consensus on the mailing list.
```
dave2wave, line 5: I think it would be better to write `Select a release manager on the dev mailing list.` Generally we do not want to vote on everything. We should prefer to have multiple volunteers to be release manager.

RELEASING.md
```diff
@@ -0,0 +1,315 @@
...
+- Create an issue for a new release in https://github.com/apache/incubator-stormcrawler/issues  If you do not have permission to do so just ask to be given permissions on the mailing list.
```
dave2wave, line 6: The release manager should be at least a committer. If they do not have permissions to create issues on GitHub then they have not mapped their apache id to their GitHub id. This mapping also requires 2FA be enabled in the user's GitHub account.

RELEASING.md
```diff
@@ -0,0 +1,315 @@
...
+                <repository>
+                    <id>apache.dist.https</id>
+                    <url>https://dist.apache.org/repos/dist</url>
+                </repository>
+            </repositories>
+        </profile>
+    </profiles>
+</settings>
+```
+
```
dave2wave, line 75: It would be good to add a Note that reminds the release manager to not checkin this settings file.  I have never pushed a release to repository.apache.org and I wonder if there is an ASFInfra request needed to be enabled.

RELEASING.md
```diff
@@ -0,0 +1,315 @@
...
+svn mv https://dist.apache.org/repos/dist/dev/incubator/stormcrawler/stormcrawler-x.y.z https://dist.apache.org/repos/dist/release/incubator/stormcrawler/stormcrawler-x.y.z -m "Release StormCrawler x.y.z"
+```
+
+This will make the release artifacts available on dist.apache.org and the artifacts will start replicating.
+
+### Delete Old Release(s)
+
+To reduce the load on the ASF mirrors, projects are required to delete old releases (see https://www.apache.org/legal/release-policy.html#when-to-archive).
+
+Remove the old releases from SVN under https://dist.apache.org/repos/dist/release/incubator/stormcrawler/.
```
dave2wave, line 267: Thank you for including this step

RELEASING.md
```diff
@@ -0,0 +1,315 @@
...
+### Update the Website
+
+- Merge the release branch to `main` to start the website deployment. 
+- Check, that the website is deployed successfully.
+- Instruction on how to do so can be found on https://github.com/apache/incubator-stormcrawler-site
+
+### Post-Release Steps
+
+- Close the present release ticket
+- Send announcement email to announce@apache.org, dev@stormcrawler.apache.org, general@incubator.apache.org. This needs to be done from your @apache.org email address or the email will bounce from the announce list.
```
dave2wave, line 278: The email should be plain text too.

dave2wave (approved):
Please consider my comments, but generally I approve.

RELEASING.md
rzo1: The settings.xml is located in ~/.m2 , so it cannot be pushed, but I agree, that we need to make it clear, where to put it.

Nexus access is enabled by default (at least for all projects I am involved with). If the deploy fails, we can open a Jira later.

Might be good to also include the steps needed to do on the Maven Staging repo such as "close" before open a vote.

RELEASING.md
rzo1: Fixed. Thx!

RELEASING.md
rzo1: I think, that this is an issue from copy pasting. I rewrote this section. Everybody with a GitHub account can create an issue.

RELEASING.md
rzo1: I updated it with the location.

rzo1:
Thanks @dave2wave for your feedback. I have addressed your comments and suggestions.

dave2wave (approved):

jnioche:
thanks @rzo1 and @dave2wave for the review

PR #1229, opened 2024-05-17
Implements a unit test for Playwright PR

rzo1:
- Implements a unit test to actually test that playwright is rendering HTML/JS content correctly.
- Adds an option to specify wait until state to ensure JS is executed properly
- Removes PDF test, which was a "false friend" because it was returned by the mock server with an HTML mimetype and the actual returned PDF object was corrupted.


rzo1:
Added a flag to skip those tests in GH actions.

jnioche (approved):

PR #1231, opened 2024-05-20
Fix #1230: Set sitemap key before outlink processing

### LINKED ISSUE #1230: Redirected sitemaps in SiteMapParserBolt / SiteMapFilter
Hi!

When the SiteMapParserBolt processes a redirected sitemap that contains links (see for example [http://sample-sitemap.xml](http://www.contexity.ch/sitemap_index.xml)), the [isSitemapKey is added](https://github.com/apache/incubator-stormcrawler/blob/0aad46b5c8a4ebcf455385272d742f8383c64e46/core/src/main/java/org/apache/stormcrawler/bolt/SiteMapParserBolt.java#L148-L150) only after the outlinks have been processed. If the SiteMapFilter is also enabled, then all the links are filtered out since the [SiteMapParserBolt.isSitemapKey is false](https://github.com/apache/incubator-stormcrawler/blob/0aad46b5c8a4ebcf455385272d742f8383c64e46/core/src/main/java/org/apache/stormcrawler/filtering/sitemap/SitemapFilter.java#L53-L54). This is fixed when the `metadata.setValue(isSitemapKey, "true")` is moved before the outlinks processing.

Also, the [sniff](https://github.com/apache/incubator-stormcrawler/blob/0aad46b5c8a4ebcf455385272d742f8383c64e46/core/src/main/java/org/apache/stormcrawler/bolt/SiteMapParserBolt.java#L377) method returns false for some sitemaps, because of the small `maxOffsetGuess`. For example, for this [sitemap](https://www.contexity.ch/post-sitemap.xml). This can be fixed by increasing `sitemap.offset.guess`, however, this might impact performance.

Should I create a PR with the fix for SiteMapParserBolt?
Should I take a shot at a different (efficient) way of "sniffing" a possible sitemap format?

Thanks,
Markos

mvolikas:
Fixes #1230

### DIFF
--- a/core/src/main/java/org/apache/stormcrawler/bolt/SiteMapParserBolt.java
+++ b/core/src/main/java/org/apache/stormcrawler/bolt/SiteMapParserBolt.java
@@ -128,6 +128,10 @@ public class SiteMapParserBolt extends StatusEmitterBolt {
             return;
         }
 
+        // mark the current doc as a sitemap
+        // as it won't have the k/v if it is a redirected sitemap
+        metadata.setValue(isSitemapKey, "true");
+
         List<Outlink> outlinks;
         try {
             outlinks = parseSiteMap(url, content, ct, metadata);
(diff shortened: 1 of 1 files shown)


jnioche (approved):

rzo1 (approved):

jnioche:
thanks @mvolikas

Issue #626, opened 2018-10-03
SOLR Status Updater - configure byDomain or byIP


jnioche:
only by host is currently implemented

jnioche:
@mvolikas is that of interest to you?

mvolikas:
I want to give it a try! I'm not 100% sure about the extra functionality we are aiming for though. Is this related to #620? I guess we could start by adding the partition key in the metadata like we do in OpenSearch?

jnioche:
It is exactly that. What this is about it to have the field name to use for the key (and whether it should be in the metadata) configurable, just like in [OpenSearch](https://github.com/apache/incubator-stormcrawler/blob/main/external/opensearch/src/main/java/org/apache/stormcrawler/opensearch/persistence/StatusUpdaterBolt.java#L154) and also add the logic of host/domain/IP .  The logic of how to query based on the shards will be added to the spouts later on in #620

Issue #620, opened 2018-10-01
Add support for shards in SOLR


jnioche:
Just like it's done in ES, we could route the documents in the statusupdaterbolt based on the host / name or IP and in the spouts check that the number of instances is equal to the # of shards and filter the queries per shard accordingly.

At the moment, we can have only one instance of a spout.

mvolikas:
This one is really interesting. Will it be up to the user to correctly create the status core with the same number of shards as the `parallelism: ***` used in the crawler.flux?

jnioche:
> This one is really interesting. Will it be up to the user to correctly create the status core with the same number of shards as the `parallelism: ***` used in the crawler.flux?

yes. we should set it to a reasonable default value (10?) but then it is up to the user to manage it

mvolikas:
@jnioche can we merge branch #851 into main at this point? This way we can update / add tests along with the new functionality.

jnioche:
Hasn't that been done in https://github.com/apache/incubator-stormcrawler/pull/1240?

mvolikas:
> Hasn't that been done in #1240? Is there more in that branch that hasn't been merged? If so, would you mind creating a PR from that branch? Thanks!

The changes from #1240 were merged into branch apache:851 but not into main. Should I open an additional PR from apache:851 to main?

jnioche:
Hi @mvolikas - sorry about the delayed response, I have just returned from holidays 

>  Should I open an additional PR from apache:851 to main?

yes please, that would be great

PR #1242, opened 2024-06-15
#626: Add routing field in metadata - Solr StatusUpdaterBolt

mvolikas:
#626 Now one can specify a routing key, which is added to the status metadata.

### DIFF
--- a/external/solr/cores/status/conf/schema.xml
+++ b/external/solr/cores/status/conf/schema.xml
@@ -25,5 +25,6 @@ under the License.
 	<field name="status" type="string" indexed="true" stored="true" required="true"/>
 	<field name="nextFetchDate" type="pdate" stored="true" indexed="true"/>
 	<dynamicField name="metadata*" type="string" indexed="true" stored="true" multiValued="true"/>
+	<field name="key" type="string" stored="true" indexed="true"/>
 	<uniqueKey>url</uniqueKey>
 </schema>
--- a/external/solr/cores/status/conf/schema.xml
+++ b/external/solr/cores/status/conf/schema.xml
@@ -26,5 +26,6 @@ under the License.
 	<field name="nextFetchDate" type="pdate" stored="true" indexed="true"/>
 	<dynamicField name="metadata*" type="string" indexed="true" stored="true" multiValued="true"/>
 	<field name="key" type="string" stored="true" indexed="true"/>
+	<dynamicField name="*" type="string" indexed="true" stored="true" multiValued="false"/>
 	<uniqueKey>url</uniqueKey>
 </schema>
--- a/external/solr/cores/status/conf/schema.xml
+++ b/external/solr/cores/status/conf/schema.xml
@@ -25,7 +25,7 @@ under the License.
 	<field name="status" type="string" indexed="true" stored="true" required="true"/>
 	<field name="nextFetchDate" type="pdate" stored="true" indexed="true"/>
 	<dynamicField name="metadata*" type="string" indexed="true" stored="true" multiValued="true"/>
-	<field name="key" type="string" stored="true" indexed="true"/>
+	<field name="key" type="string" stored="true" indexed="true" multiValued="false"/>
 	<dynamicField name="*" type="string" indexed="true" stored="true" multiValued="false"/>
 	<uniqueKey>url</uniqueKey>
 </schema>
--- a/external/solr/src/main/java/org/apache/stormcrawler/solr/persistence/StatusUpdaterBolt.java
+++ b/external/solr/src/main/java/org/apache/stormcrawler/solr/persistence/StatusUpdaterBolt.java
@@ -46,7 +46,6 @@ public class StatusUpdaterBolt extends AbstractStatusUpdaterBolt {
 
     private static final String SolrMetadataPrefix = "solr.status.metadata.prefix";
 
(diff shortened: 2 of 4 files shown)


external/solr/cores/status/conf/schema.xml
```diff
@@ -25,5 +25,7 @@ under the License.
 	<field name="status" type="string" indexed="true" stored="true" required="true"/>
 	<field name="nextFetchDate" type="pdate" stored="true" indexed="true"/>
 	<dynamicField name="metadata*" type="string" indexed="true" stored="true" multiValued="true"/>
+	<field name="key" type="string" stored="true" indexed="true"/>
```
jnioche, line 28: @mvolikas why not set the key field to multiValued = false?
I assume the dynamic field is to allow users to set a different name than 'key'

external/solr/src/main/java/org/apache/stormcrawler/solr/persistence/StatusUpdaterBolt.java
```diff
@@ -83,6 +126,23 @@ public void store(
             doc.setField(String.format(Locale.ROOT, "%s.%s", mdPrefix, key), values);
         }
 
+        String partitionKey = partitioner.getPartition(url, metadata);
+        if (partitionKey == null) {
+            partitionKey = "_DEFAULT_";
+        }
+
+        // store routing key in metadata?
+        if (StringUtils.isNotBlank(fieldNameForRoutingKey) && routingFieldNameInMetadata) {
```
jnioche, line 135: given that this will not change from one document to the other, why not determine the field name to use once and for all instead of doing it for every document?

external/solr/cores/status/conf/schema.xml
mvolikas: Yes, that's the idea behind the dynamic field, and you are right, "key" should have `multiValued = false` too. thanks for pointing it out!

external/solr/src/main/java/org/apache/stormcrawler/solr/persistence/StatusUpdaterBolt.java
mvolikas: Yes, i will change this. I followed the logic from opensearch StatusUpdaterBolt that does this check inside `store()` but I guess this improvement can also be applied there.

rzo1 (approved):

jnioche:
do we need _doRouting_ at all? Not sure how the routing is done for SOLR, is it via the collection definition or the indexing API as in OpenSearch?

rzo1:
> do we need _doRouting_ at all? Not sure how the routing is done for SOLR, is it via the collection definition or the indexing API as in OpenSearch?

At least the variable is never used, so might be dropped.

mvolikas:
> do we need _doRouting_ at all? 

I'm aware that `doRouting` is not currently used. I left the variable for use in #620 but we can remove it for now.

> Not sure how the routing is done for SOLR, is it via the collection definition or the indexing API as in OpenSearch?

If I'm not mistaken it depends on the `router.name` specified [when creating the collection](https://solr.apache.org/guide/solr/latest/deployment-guide/collection-management.html#create-parameters). I was thinking to use the `compositeId` option so that we do not manually specify the destination shard during indexing.

jnioche:
> > do we need _doRouting_ at all?
> 
> I'm aware that `doRouting` is not currently used. I left the variable for use in #620 but we can remove it for now.

yes please, let's add it back later  with the rest of the code that relies on it

> 
> > Not sure how the routing is done for SOLR, is it via the collection definition or the indexing API as in OpenSearch?
> 
> If I'm not mistaken it depends on the `router.name` specified [when creating the collection](https://solr.apache.org/guide/solr/latest/deployment-guide/collection-management.html#create-parameters). I was thinking to use the `compositeId` option so that we do not manually specify the destination shard during indexing.

I suppose what we need is to use 

`router.field (v1), router (v2)`

and point it to the _key_ field, otherwise it will take the unique document key. Is that correct?

mvolikas:
> and point it to the key field, otherwise it will take the unique document key. Is that correct?

Yes, you are right - thanks for noting!

jnioche:
> > and point it to the key field, otherwise it will take the unique document key. Is that correct?
> 
> Yes, you are right - thanks for noting!

where should `router.field=key` be specified? in solrconfig.xml?

mvolikas:
Thanks for the patience - I just found some time to look again into this :-)
> where should router.field=key be specified? in solrconfig.xml?

I don't think this will work. The `router.field` is configured during the collection creation with the [Collections API](https://solr.apache.org/guide/solr/latest/deployment-guide/collection-management.html#create). (collection specific)

So, for #620 we could do the following:

- Add a new property `solr.status.routing.shards: 10` to solr-conf.yaml
- Add a script that creates the "status" collection with the numShards equal to the property above, and the `router.field` equal to the `solr.status.routing.fieldname: "key"`.
- The user should make sure that the SolrSpout parallelism equals the number of shards specified.

jnioche:
Thanks @mvolikas 
Ok so the logic will be done in #620 so for this one we just need to remove SolrStatusRoutingParamName, correct?

mvolikas:
Yes, this variable is #620 related and was unused. I removed it for this PR.

jnioche:
Thanks @mvolikas

PR #1250, opened 2024-07-08
#1247 Fix race condition in HttpProtocol while having multiple proxy settings

### LINKED ISSUE #1247: HttpProtocol (both okhttp and apache) race condition while having different proxies in different threads
What kind of issue is this?

 - [ ] Question. This issue tracker is not the best place for questions. If you want to ask how to do
       something, or to understand why something isn't working the way you expect it to, use StackOverflow
       instead with the label 'stormcrawler': https://stackoverflow.com/questions/tagged/stormcrawler 

 - [x] Bug report. If you’ve found a bug, please include a test if you can, it makes it a lot easier to fix things. Use the label 'bug' on the issue.
 
 - [ ] Feature request. Please use the label 'wish' on the issue.

### Reproduce steps

To reproduce it, we can run the HttpProtocol main function with many urls with MultiProxyFactory

the crawler.conf
```
config:
  http.agent.name: test
  http.proxy.manager: org.apache.stormcrawler.proxy.MultiProxyManager
  http.proxy.file: proxies
  http.robots.file.skip: true
```

the proxies file
```
http://first:password@proxy1:8888
http://second:password@proxy2:8888
```

### Root cause

The HttpProtocol (both okhttp and apache) is not thread-safe
- the same instance which was initiated by ProxyFactory may be used in different bolts (different workers) at same time
- the shared request/client builder was manipulated by different bolt/thread at same time

Example 1 (wrong proxy auth)
- (Thread 2) builder.setProxy(secondProxy)
- (Thread 1) builder.setProxy(firstProxy)
- (Thread 1) builder.setAuth(firstAuth)
- (Thread 2) builder.setAuth(secondAuth)
- (Thread 1) builder.build()
- We'll have firstProxy + secondAuth

Example 2 (wrong proxy used)
- (Thread 1) builder.setProxy(firstProxy)
- (Thread 1) builder.setAuth(firstAuth)
- (Thread 2) builder.setProxy(secondProxy)
- (Thread 2) builder.setAuth(secondAuth)
- (Thread 1) builder.build()
- Now both requests use the second proxy

chhsiao90:
In HttpProtocol implementation, the client builder was singleton and may be accessed and modified by different threads at same time. The result is that a wrong proxy will be used or a wrong proxy auth will be configured.

To fix it, create a local builder insteand of having a field-level builder.

Fixes #1247

### For all changes:

### For code changes:

- [x] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)?
- [x] If applicable, have you updated the LICENSE file, including the main LICENSE file?
- [x] If applicable, have you updated the NOTICE file, including the main NOTICE file?

### Note:

### DIFF
--- a/core/src/main/java/org/apache/stormcrawler/protocol/okhttp/HttpProtocol.java
+++ b/core/src/main/java/org/apache/stormcrawler/protocol/okhttp/HttpProtocol.java
@@ -32,9 +32,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
-import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLContext;
-import javax.net.ssl.SSLSession;
 import javax.net.ssl.SSLSocketFactory;
 import javax.net.ssl.TrustManager;
 import javax.net.ssl.X509TrustManager;
@@ -43,8 +41,6 @@ import okhttp3.Call;
 import okhttp3.Connection;
 import okhttp3.ConnectionPool;
 import okhttp3.Credentials;
-import okhttp3.EventListener;
-import okhttp3.EventListener.Factory;
 import okhttp3.Handshake;
 import okhttp3.Headers;
 import okhttp3.Interceptor;
@@ -87,6 +83,18 @@ public class HttpProtocol extends AbstractHttpProtocol {
 
     private int completionTimeout = -1;
 
+    private int httpTimeout;
+
+    private boolean retryOnConnectionFailure;
+
+    private boolean followRedirects;
+
+    private boolean insecure;
+
+    private final List<okhttp3.Protocol> protocols = new ArrayList<>();
+
+    private ConnectionPool connectionPool = null;
+
     /** Accept partially fetched content as trimmed content */
     private boolean partialContentAsTrimmed = false;
 
@@ -94,8 +102,6 @@ public class HttpProtocol extends AbstractHttpProtocol {
 
     // track the time spent for each URL in DNS resolution
     private final Map<String, Long> DNStimes = new HashMap<>();
-
-    private OkHttpClient.Builder builder;
 
     private static final TrustManager[] trustAllCerts =
             new TrustManager[] {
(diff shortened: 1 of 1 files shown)


jnioche:
This is a substantial change that could have an impact on how things work in SC. It needs careful reviewing, annoyingly I don't have much time to do that at the moment but I hope I will be able to soon. Reviews from other committers would be welcome

pjfanning:
My gut reaction is that adding even more shared state to the classes is not a good way to go.

How about just using a ReentrantLock to wrap the code where there is a race condition?

pjfanning:
The solutions applied in okhttp and httpclient classes seem very different. It would greatly simply review to have 2 different PRs that can be reviewed independently.

chhsiao90:
> The solutions applied in okhttp and httpclient classes seem very different. It would greatly simply review to have 2 different PRs that can be reviewed independently.

Sure, I'll separate the httpclient in a separated PR.

> My gut reaction is that adding even more shared state to the classes is not a good way to go.

Actually, those newly added fields are not shared state. They are just the http options retrieved from the conf. 

The easiest and most readable way is to just have a local field conf, and we always build the http client from the conf. But that would introduce unnecessary map.get operations as the conf should keep unchanged when it was loaded.

So I decided to have these option fields to be loaded during the `void configure(conf)` method call. And I think that's how most other storm crawler classes do, ex: JSoupParserBolt.

What do you think?

> The more I look at this the more I think there is a major design flaw. The HttpProtocol classes should not be reused for HTTP calls that have different configs. There should be separate HttpProtocol class instances created for each different config. In #1247, there are 2 proxies described, HTTP calls for those 2 proxies should not be made with the same HTTPProtocol instance.

I think it's might be a common case to round-robin a large proxy pool (may have thousands or more ips). So it makes sense to me to have a single HttpProtocol, and select a proxy from ProxyManager whenever it was called.

If we'd like to have multiple HttpProtocol instances for each proxy, it means that we may round-robin the HttpProtocol in ProtocolFactory. And I'm not sure if it's easy to implement it.

chhsiao90:
Pushed the update to remove the change of httpclient HttpProtocol.

rzo1:
Actually, I am not using proxies for my setup, so I am unable to test it. Would it be possible to add unit tests (leveraging test containers) to spin up a proxy-based setup to demonstrate the correctness of this change?

jnioche:
> My gut reaction is that adding even more shared state to the classes is not a good way to go.
> 
> The more I look at this the more I think there is a major design flaw. The HttpProtocol classes should not be reused for HTTP calls that have different configs. There should be separate HttpProtocol class instances created for each different config. In #1247, there are 2 proxies described, HTTP calls for those 2 proxies should not be made with the same HTTPProtocol instance.

I like this approach. We already have the ability to have more than 1 instance of each protocol, ideally we would be able to configure them differently

jnioche:
Closing due to a lack of activity. Feel free to reopen if necessary

Issue #1162, opened 2024-03-27
Drop coveralls code-coverage flow?


rzo1:
Can we drop the coveralls github action? 

I did a quick search at ASF level and I think, that only datasketches-cpp is using it. So might be good to just drop it and rely on other test coverage report metrics, which are more common at ASF level.

lewismc:
SonarCloud.io is very good.

sigee:
Hi,
For checking and requiring a minimum test coverage you could use jacoco as it is used in [apache commons project](https://github.com/apache/commons-parent/blob/master/pom.xml#L218).

There is no generated badge for readme, but the pipeline could work perfectly by require a minimum coverage for different metrics (line coverage, branch coverage, class coverage, etc.).
Jacoco is already part of this project, so only the configuration is needed here.

What do you think?

rzo1:
I am using Jacoco in most of my projects ;-) - I have no objections in replacing it.

PR #1255, opened 2024-08-28
#1162 - Replace Coveralls with JaCoCo

sigee:
### For all changes:
- [X] Is there a issue associated with this PR? Is it referenced in the commit message?

- [X] Does your PR title start with `#XXXX` where `XXXX` is the issue number you are trying to resolve? 

- [X] Has your PR been rebased against the latest commit within the target branch (typically main)?

- [X] Is your initial contribution a single, squashed commit?

- [X] Is the code properly formatted with `mvn git-code-format:format-code -Dgcf.globPattern="**/*" -Dskip.format.code=false`?

### For code changes:

- [X] Have you ensured that the full suite of tests is executed via `mvn clean verify`?
- [X] If applicable, have you updated the README file, including the main README file?

### Note:
I chceked the current coverage in every module and set the minimum threshold for the curent values.
From now the build will be failed if the coverage drops, but the threshold should be updated time to time if the coverage is increased.


rzo1 (approved):
I am ok with it, but wouldn't actually hard-code jacoco to fail on certain metrics. But that is only my personal preference. WDYT @jnioche ?

jnioche:
> I am ok with it, but wouldn't actually hard-code jacoco to fail on certain metrics. But that is only my personal preference. WDYT @jnioche ?

If the thresholds are set to the current values, this would mean that any future contribution would need to have perfect test coverage which is drastic but maybe not a bad thing. We could give it a try, review later and decide to scrap it if it turns out to be a pain

jnioche (approved):

jnioche:
Thanks @sigee

PR #1325, opened 2024-09-29
#1323 Update archetype Storm version from 2.6.4

### LINKED ISSUE #1323: Upgrade dependency Storm 2.6.4
The [maven archetype pom](https://github.com/apache/incubator-stormcrawler/blob/main/archetype/src/main/resources/archetype-resources/pom.xml#L35) has `storm.version 2.6.2`. When using this with `stormcrawler-solr` dependency a jetty version incompatibility issue arises. In particular, when submitting the shaded jar to Storm I get:
```
java.lang.IncompatibleClassChangeError:
class org.eclipse.jetty.http.HttpFields$Mutable can not implement org.eclipse.jetty.http.HttpFields,
because it is not an interface (org.eclipse.jetty.http.HttpFields is in unnamed module of loader 'app')
```
This is because Solr 9.6.1 uses Jetty 10.* while Storm 2.6.2 Jetty 9.*. Am I missing something?
When switching to storm 2.6.4 this gets resolved. Given that in storm-crawler 3.1.0 Storm 2.6.4 is used in the [main pom](https://github.com/apache/incubator-stormcrawler/blob/main/pom.xml#L68), maybe we could update the archetype pom to the latest version as well?

mvolikas:
Fixes #1323

### DIFF
--- a/archetype/src/main/resources/archetype-resources/pom.xml
+++ b/archetype/src/main/resources/archetype-resources/pom.xml
@@ -32,7 +32,7 @@ under the License.
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<stormcrawler.version>${project.version}</stormcrawler.version>
-		<storm.version>2.6.2</storm.version>
+		<storm.version>2.6.4</storm.version>
 	</properties>
 
 	<build>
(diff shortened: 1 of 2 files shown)


jnioche (approved):

jnioche:
Thanks @mvolikas

PR #1341, opened 2024-10-04
Store metadata as WARC Metadata records

jnioche:
The WARC format defines [Metadata records](https://iipc.github.io/warc-specifications/specifications/warc-format/warc-1.1/#example-of-metadata-record). We want to be able to store key values from the Metadata in there, separately from the _response_ records.

Writing these records will be triggered by the presence in the configuration of _warc.metadata.keys_.

### DIFF
--- a/external/warc/src/main/java/org/apache/stormcrawler/warc/MetadataRecordFormat.java
+++ b/external/warc/src/main/java/org/apache/stormcrawler/warc/MetadataRecordFormat.java
@@ -1,0 +1,127 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. The ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * <p>http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * <p>Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.stormcrawler.warc;
+
+import java.net.URI;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import java.util.UUID;
+import org.apache.commons.lang.StringUtils;
+import org.apache.storm.tuple.Tuple;
+import org.apache.stormcrawler.Metadata;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/*
+
+Example if metadata record
+https://iipc.github.io/warc-specifications/specifications/warc-format/warc-1.1/#example-of-metadata-record
+
+WARC/1.1
+WARC-Type: metadata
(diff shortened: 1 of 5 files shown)


external/warc/src/main/java/org/apache/stormcrawler/warc/MetadataRecordFormat.java
```diff
@@ -0,0 +1,127 @@
...
+        buffer.append("Content-Length: ").append(Integer.toString(contentLength)).append(CRLF);
+
+        String blockDigest = getDigestSha1(metadata_representation);
+
+        String captureTime = getCaptureTime(metadata);
+        buffer.append("WARC-Date: ").append(captureTime).append(CRLF);
+
+        // must be a valid URI
+        try {
+            String normalised = url.replaceAll(" ", "%20");
```
rzo1, line 103: Do we need to escape other URI stuff as well?

Side note: If we don't need a regex, it might be faster to just go for a `String.replace(...)`.

external/warc/src/main/java/org/apache/stormcrawler/warc/MetadataRecordFormat.java
```diff
@@ -0,0 +1,127 @@
...
+            String targetURI = URI.create(normalised).toASCIIString();
+            buffer.append("WARC-Target-URI: ").append(targetURI).append(CRLF);
+        } catch (Exception e) {
+            LOG.warn("Incorrect URI: {}", url);
+            return new byte[] {};
```
rzo1, line 108: new byte[0]; ?

external/warc/src/main/java/org/apache/stormcrawler/warc/MetadataRecordFormat.java
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        buffer.append("Content-Type: application/warc-fields").append(CRLF);
+        buffer.append("WARC-Block-Digest: ").append(blockDigest).append(CRLF);
+
+        byte[] buffAsBytes = buffer.toString().getBytes(StandardCharsets.UTF_8);
+
+        int capacity = 6 + buffAsBytes.length + metadata_representation.length;
```
rzo1, line 116: Why 6 ? (magic number)

external/warc/src/test/java/org/apache/stormcrawler/warc/WARCRecordFormatTest.java
```diff
@@ -16,18 +16,22 @@
  */
 package org.apache.stormcrawler.warc;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.*;
```
rzo1, line 19: Can we avoid star imports?

rzo1 (approved):
Left some minor stuff, but overall lgtm.

external/warc/src/main/java/org/apache/stormcrawler/warc/MetadataRecordFormat.java
jnioche: Not that I am aware of, I reproduced the behaviour of the existing classes.
I will put the logic in the parent class so that if it needs fixing, we can do it in a single place

external/warc/src/test/java/org/apache/stormcrawler/warc/WARCRecordFormatTest.java
jnioche: naughty IDE...

Issue #621, opened 2018-10-02
asynchronous queries and updates in SOLR?


jnioche:
Maybe https://github.com/inoio/solrs would be useful?

jnioche:
see https://solr.apache.org/guide/solr/latest/deployment-guide/solrj.html

[ConcurrentUpdateHttp2SolrClient](https://solr.apache.org/docs/9_0_0/solrj/org/apache/solr/client/solrj/impl/ConcurrentUpdateHttp2SolrClient.html) - just like ConcurrentUpdateSolrClient but using Http2SolrClient instead. This class is experimental therefore its API’s might change or be removed in minor versions of SolrJ.

Not sure whether CloudSolrClient is asynchronous

[CloudSolrClient] (https://solr.apache.org/docs/9_0_0/solrj/org/apache/solr/client/solrj/impl/CloudSolrClient.html) - geared towards communicating with SolrCloud deployments. Uses already-recorded ZooKeeper state to discover and route requests to healthy Solr nodes.

mvolikas:
My understanding is the following:

- [ConcurrentUpdateHttp2SolrClient](https://solr.apache.org/docs/9_7_0/solrj/org/apache/solr/client/solrj/impl/ConcurrentUpdateHttp2SolrClient.html) is not experimental in version 9.7.0 and seems promising for the `StatusUpdater` and `Inderer` bolts where it can batch updates without changing the way we submit documents.

- Regarding querying in the spouts we could use the [requestAsync](https://solr.apache.org/docs/9_7_0/solrj/org/apache/solr/client/solrj/impl/Http2SolrClient.html#requestAsync(org.apache.solr.client.solrj.SolrRequest,java.lang.String)) of the `Http2SolrClient` with futures.

- I'm not sure about the benefits of the [CloudSolrClient](https://solr.apache.org/docs/9_7_0/solrj/org/apache/solr/client/solrj/impl/CloudSolrClient.html), to be honest. The API seems to be synchronous from the caller's perspective. Additionally, since we currently have a single node (with multiple shards for the status collection) we will not use any of the automatic routing (e.g. load balancing requests across nodes) and failover (e.g. a node goes down) features.

Any thoughts?

mvolikas:
[...]

PR #1343, opened 2024-10-05
#620 Add support for shards - SolrSpout

mvolikas:
This PR (work in progress) employs the following strategy for supporting shards in `status`:

1. Add a script to start Solr in cloud mode and create the collections using configsets. This script configures the `status` collection to use the number of shards defined in `solr-conf.yaml` (`solr.status.routing.shards`) and sets the sharding field to be the same as `solr.status.routing.fieldname`.
2. Each SolrSpout instance fetches documents from the corresponding shard.
3. Unlike OpenSearch, the StatusUpdaterBolt does not need further changes to route to the correct shard. This is done implicitly by Solr because of (1).

Pending tasks:

- [x] Tests should be updated to use Solr cloud and verify the shards are populated correctly.
- [x] Add test (spout) for the case of multiple (2) Solr shards.
- [x] Test with Storm 2.7.0


mvolikas (commented):
Unlike OpenSearch, in SolrSpout we do not call `markQueryReceivedNow()`. Should we add this?

external/solr/setup-solr.sh
mvolikas, line 1: I guess we should create an `archetype` module under `external/solr` and add the scripts in the `archetype-resourcses` [like in OpenSearch](https://github.com/apache/incubator-stormcrawler/blob/main/external/opensearch/archetype/src/main/resources/archetype-resources/OS_IndexInit.sh) right?

jnioche: Could do but it's another thing to maintain and keep in sync.

jnioche:
> Unlike OpenSearch, in SolrSpout we do not call `markQueryReceivedNow()`. Should we add this?

yes and also set  ` isInQuery.set(true);`

external/solr/setup-solr.sh
```diff
@@ -0,0 +1,59 @@
...
+$SOLR_HOME/bin/solr start -cloud -p $SOLR_PORT
+
+$SOLR_HOME/bin/solr zk upconfig -n "docs" -d configsets/docs -z localhost:9983
+$SOLR_HOME/bin/solr zk upconfig -n "status" -d configsets/status -z localhost:9983
+$SOLR_HOME/bin/solr zk upconfig -n "metrics" -d configsets/metrics -z localhost:9983
+
+echo "Creating 'docs' collection ..."
+curl -X POST "http://localhost:$SOLR_PORT/api/collections" -H "Content-type:application/json" -d '
+  {
+    "name": "docs",
```
jnioche, line 33: does this replace the core.properties files?

external/solr/src/main/java/org/apache/stormcrawler/solr/persistence/SolrSpout.java
```diff
@@ -135,7 +157,8 @@ else if (resetFetchDateAfterNSecs != -1) {
 
         query.setQuery("*:*")
                 .addFilterQuery("nextFetchDate:[* TO " + lastNextFetchDate + "]")
-                .setSort("nextFetchDate", ORDER.asc);
+                .setSort("nextFetchDate", ORDER.asc)
+                .setParam("shards", "shard" + shardID);
```
jnioche, line 161: IIRC the way it works with the other backends is that if the number of shards in the conf is either unspecified or 1, we let a single instance take care of everything and so this line
```
  .setParam("shards", "shard" + shardID);
```
would not be executed

external/solr/setup-solr.sh
mvolikas: Yes. Since we are switching to SolrCloud, the recommended way to create a collection is through the [Collections API](https://solr.apache.org/guide/solr/latest/deployment-guide/collection-management.html#create), where `name` is a required parameter. Additionally, in SolrCloud the configuration (configsets) and data are [managed by ZooKeeper](https://solr.apache.org/guide/solr/latest/configuration-guide/config-sets.html#configsets-in-solrcloud-clusters), so my understanding is that `core.properties` is not relevant anymore.

external/solr/setup-solr.sh
mvolikas: Hmm, you are right. But the alternative would be to have the user copy the scripts and the `solr-conf.yaml` from the source code into their project? I had difficulty setting things up the first time and it would have been much less friction if one could create a `stormcrawler-solr-archetype` directly. But even if we do not create the Solr archetype, where should the scripts be located? Directly under `.../external/solr/`?

external/solr/src/main/java/org/apache/stormcrawler/solr/persistence/SolrSpout.java
mvolikas: The idea was to have the `shardID` default to 1 when the respective property is not set.

- [Definition](https://github.com/apache/incubator-stormcrawler/pull/1343/files#diff-a499db920515864e856bb6616ff489daf6a58af1159b2f6057ea7da2273e774eR64)
- [Getting the number of shards from the conf](https://github.com/apache/incubator-stormcrawler/pull/1343/files#diff-a499db920515864e856bb6616ff489daf6a58af1159b2f6057ea7da2273e774eR88-R94)

The request to Solr will then work since the collection will have just one shard: `shard1`. I will add a test for this to make sure I'm not missing anything.

mvolikas: I added some defaults to the scripts as well. [d22c325](https://github.com/apache/incubator-stormcrawler/pull/1343/commits/d22c3259df1c9ec07b4b3d2c386cf2385ee0bc23)

external/solr/src/main/java/org/apache/stormcrawler/solr/persistence/SolrSpout.java
```diff
@@ -156,10 +176,15 @@ else if (resetFetchDateAfterNSecs != -1) {
...
         try {
+            LOG.trace("isInQuery set to true");
+            isInQuery.set(true);
+
             long startQuery = System.currentTimeMillis();
             QueryResponse response = connection.getClient().query(query);
             long endQuery = System.currentTimeMillis();
 
+            markQueryReceivedNow();
+
```
mvolikas, line 187: I guess the `isInQuery.set(true)` and `markQueryReceivedNow()` are only relevant when the calls to Solr are asynchronous right? If so, we will have to review this for #621.

external/solr/src/test/java/org/apache/stormcrawler/solr/persistence/SpoutTest.java
```diff
@@ -0,0 +1,169 @@
...
+        componentToTasks.put("solrSpout", Arrays.asList(0, 1));
+
+        return new TopologyContext(
+                null,
+                Map.of("storm.cluster.mode", "local"),
+                taskToComponent,
+                componentToTasks,
+                new HashMap<>(),
+                null,
+                null,
```
mvolikas, line 113: When writing the test with the 2 spouts, I manually created this Storm `TopologyContext` object with most of the parameters set to `null`. Is this ok? Should we set anything else?

external/solr/setup-solr.sh
jnioche: or in a sub directory?
I am not against having an archetype for SOLR but so far there has been very little interest from the community, which is why it is lagging behind in terms of features compared to OpenSearch. Having an archetype would definitely help getting people on board though. Is that something you would like to contribute?

external/solr/src/main/java/org/apache/stormcrawler/solr/persistence/SolrSpout.java
jnioche: I would simply not set the param constraint in the query if the config for the number of shards is set or defaults to 1, this way it wouldn't matter how the collection is configured - you would get all the results. Doesn't matter too much though and the current behaviour is also fine.

external/solr/src/main/java/org/apache/stormcrawler/solr/persistence/SolrSpout.java
jnioche: It is used in the super class and has an impact on when _populateBuffer_ is called
see [AbstractQueryingSpout](https://github.com/apache/incubator-stormcrawler/blob/6ba510c91848dd3d8bce09dc65ac97194eea829c/core/src/main/java/org/apache/stormcrawler/persistence/AbstractQueryingSpout.java#L182)

external/solr/src/test/java/org/apache/stormcrawler/solr/persistence/SpoutTest.java
jnioche: no idea to be honest but it feels more complicated than what we've had to do for the other tests.
What about reusing ` TestUtil.getMockedTopologyContext() `?

external/solr/setup-solr.sh
rzo1: Guess it would be good to have an archetype. We can still remove it, if it gets dormant. At least it is a good possibility to promote that module, no?

mvolikas: In our team, we currently have an ES deployment for StormCrawler. The main reason for not using Solr in the first place was that we could not have many spouts running in parallel and were not sure how well it would scale. I think this will no longer be the case after issues #620 and #621 get resolved and we would prefer Solr for many reasons - Apache ecosystem, more experience using it, and easier deployment (at least in our use case) to name a few.

rzo1:
@mvolikas Do you aim to include this PR in **3.1.1** - if so, do you think you can work on the open comments or would it be ok to move it to **3.1.2**?

mvolikas:
> @mvolikas Do you aim to include this PR in **3.1.1** - if so, do you think you can work on the open comments or would it be ok to move it to **3.1.2**?

From my side, a safe estimate for having this ready and tested would be the first week of November. If the 3.1.1 release is planned earlier please move this to 3.1.2. Thanks!

jnioche:
> > @mvolikas Do you aim to include this PR in **3.1.1** - if so, do you think you can work on the open comments or would it be ok to move it to **3.1.2**?
> 
> From my side, a safe estimate for having this ready and tested would be the first week of November. If the 3.1.1 release is planned earlier please move this to 3.1.2. Thanks!

I am sure we can wait. This will be a great addition to the next release

external/solr/archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -0,0 +1,154 @@
...
+  <artifactId>${artifactId}</artifactId>
+  <version>${version}</version>
+  <packaging>jar</packaging>
+
+  <name>${artifactId}</name>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <stormcrawler.version>${StormCrawlerVersion}</stormcrawler.version>
+    <storm.version>2.6.4</storm.version>
```
jnioche, line 37: we're now on 2.7.0

external/solr/archetype/src/main/resources/archetype-resources/README.md
```diff
@@ -0,0 +1,144 @@
+This has been generated by the StormCrawler Maven Archetype as a starting point for building your own crawler with [Solr](https://solr.apache.org/) as a backend.
+Have a look at the code and resources and modify them to your heart's content.
+
+You need to have Apache Storm installed, as well as a running instance of OpenSearch.
```
jnioche, line 4: ... of Apache SOLR

jnioche (changes requested):
Currently doesn't pass the tests because of missing license headers

Running the archetype generation with 

mvn archetype:generate -DarchetypeGroupId=org.apache.stormcrawler -DarchetypeArtifactId=stormcrawler-solr-archetype -DarchetypeVersion=3.1.1-SNAPSHOT 

fails

```
Caused by: org.apache.maven.plugin.MojoFailureException: java.io.IOException: No such file or directory
    at org.apache.maven.archetype.mojos.CreateProjectFromArchetypeMojo.execute (CreateProjectFromArchetypeMojo.java:216)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:126)
```

external/solr/README.md
```diff
@@ -1,117 +1,30 @@
...
+You'll be asked to enter a groupId (e.g. com.mycompany.crawler), an artefactId (e.g. stormcrawler), a version, a package name and details about the user agent to use.
 
-* `solr.metrics.ttl`: [Date expression](https://cwiki.apache.org/confluence/display/solr/Working+with+Dates) to specify when the document should expire.
-* `solr.metrics.ttl.field`: Field to be used to specify the [date expression](https://cwiki.apache.org/confluence/display/solr/Working+with+Dates) that defines when the document should expire.
+This will not only create a fully formed project containing a POM with the dependency above but also a set of resources, configuration files and sample topology classes. Enter the directory you just created (should be the same as the artefactId you specified earlier) and follow the instructions on the README file.
 
-*Note: The date expression specified in the `solr.metrics.ttl` parameter is not validated. To use this feature some changes in the Solr configuration must be done.*
+You will of course need to have both Storm and Solr installed.
```
jnioche, line 15: call them by their full names and maybe indicate which version of Apache Solr is needed?
9.7.0 currently

mvolikas:
> Running the archetype generation with
> 
> mvn archetype:generate -DarchetypeGroupId=org.apache.stormcrawler -DarchetypeArtifactId=stormcrawler-solr-archetype -DarchetypeVersion=3.1.1-SNAPSHOT
> 
> fails

@jnioche This is strange; I cannot reproduce it locally. In my case, it runs as expected by first running:
`mvn clean install` for `incubator-stormcrawler project`
Then running the `mvn archetype` command you wrote.
```
mvn --version
Apache Maven 3.8.7
Maven home: /usr/share/maven
Java version: 17.0.12, vendor: Ubuntu, runtime: /usr/lib/jvm/java-17-openjdk-amd64
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "6.8.0-47-generic", arch: "amd64", family: "unix"
```

Could it be a directory permissions issue?
Can you provide the full exception trace?

external/solr/src/test/java/org/apache/stormcrawler/solr/persistence/SpoutTest.java
mvolikas: This didn't work, since what I wanted to test involved the SolrSpout calling [context.getComponentTasks()](https://github.com/apache/incubator-stormcrawler/blob/main/external/solr/src/main/java/org/apache/stormcrawler/solr/persistence/SolrSpout.java#L78) which in turn reads the [componentToTasks](https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/task/GeneralTopologyContext.java#L105) for example. I started from the [FileSpoutTopologyContextMock](https://github.com/apache/incubator-stormcrawler/blob/main/core/src/test/java/org/apache/stormcrawler/spout/mocks/FileSpoutTopologyContextMock.java#L28-L48) and we could in principle have something similar if we want to have more such SolrSpout tests in the future.

jnioche:
archetype generated successfully, no idea why it  had failed

jnioche:
@mvolikas compiling the project generated from the archetype fails with 

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project crawlzob: Compilation failure
[ERROR] /tmp/crawlzob/src/main/java/com/dipe/CrawlTopology.java:[34,8] class SolrCrawlTopology is public, should be declared in a file named SolrCrawlTopology.java

```

jnioche:
@mvolikas The Java based topologies could actually go. We don't have them in the OpenSearch module and I think the huge majority of people just rely on the Flux files.
@rzo1 any thoughts?

mvolikas:
> @mvolikas compiling the project generated from the archetype fails with
> 
> ```
> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project crawlzob: Compilation failure
> [ERROR] /tmp/crawl/src/main/java/com/dipe/CrawlTopology.java:[34,8] class SolrCrawlTopology is public, should be declared in a file named SolrCrawlTopology.java
> ```

Yes I have not yet tested with the java topologies. One thing to note is that after generating the default StormCrawler archetype with
```
mvn archetype:generate -DarchetypeGroupId=org.apache.stormcrawler -DarchetypeArtifactId=stormcrawler-archetype -DarchetypeVersion=3.1.0
```
and then running `mvn clean package` I also get an error:
```
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/markos/apache/test/src/main/java/test/CrawlTopology.java:[35,36] cannot find symbol
  symbol: class ConfigurableTopology
[INFO] 1 error
[INFO] -------------------------------------------------------------
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  15.617 s
[INFO] Finished at: 2024-11-03T15:46:48+02:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project test: Compilation failure
[ERROR] /home/markos/apache/test/src/main/java/test/CrawlTopology.java:[35,36] cannot find symbol
[ERROR]   symbol: class ConfigurableTopology
[ERROR] 
[ERROR] -> [Help 1]
```

Can you confirm that?

external/solr/archetype/src/main/resources/archetype-resources/src/main/java/CrawlTopology.java
rzo1, line 1: This class file needs to be renamed to `SolrCrawlTopology` **or** the class name must change to `CrawlTopology`.

rzo1:
> 
> ```
> [ERROR] COMPILATION ERROR : 
> [INFO] -------------------------------------------------------------
> [ERROR] /home/markos/apache/test/src/main/java/test/CrawlTopology.java:[35,36] cannot find symbol
>   symbol: class ConfigurableTopology
> [INFO] 1 error
> [INFO] -------------------------------------------------------------
> [INFO] ------------------------------------------------------------------------
> [INFO] BUILD FAILURE
> [INFO] ------------------------------------------------------------------------
> [INFO] Total time:  15.617 s
> [INFO] Finished at: 2024-11-03T15:46:48+02:00
> [INFO] ------------------------------------------------------------------------
> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project test: Compilation failure
> [ERROR] /home/markos/apache/test/src/main/java/test/CrawlTopology.java:[35,36] cannot find symbol
> [ERROR]   symbol: class ConfigurableTopology
> [ERROR] 
> [ERROR] -> [Help 1]
> ```
> 
> Can you confirm that?

```bash
[INFO] -------------------------------------------------------------
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  33.148 s
[INFO] Finished at: 2024-11-03T15:16:00+01:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project bla: Compilation failure
[ERROR] /home/rz/tmp/bla/src/main/java/bla/CrawlTopology.java:[35,36] cannot find symbol
[ERROR]   symbol: class ConfigurableTopology
[ERROR] 
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException

```

The actually issue is, that the template misses an import for `import org.apache.stormcrawler.ConfigurableTopology;` -> https://github.com/apache/incubator-stormcrawler/issues/1389

external/solr/archetype/src/main/resources/archetype-resources/src/main/java/CrawlTopology.java
rzo1: It also requires the missing import from https://github.com/apache/incubator-stormcrawler/issues/1389

jnioche:
> The actually issue is, that the template misses an import for `import org.apache.stormcrawler.ConfigurableTopology;` -> #1389

The fact that it has been broken for ever and no one reported it kind of suggests that hardly anyone uses the core archetype. 
I would be in favour of getting rid of duplicate functionality and remove the Java topologies and haev only the Flux files. 
We need one for the injection @mvolikas

jnioche:
> Yes I have not yet tested with the java topologies

The compilation fails whether you use the Java topologies or not...

rzo1:
Bascially, the Java topologies are only good for testing in local mode (IMHO) and are actually only usable, if the IDE is configured to include the `provided` scoped dependencies in the run configuration (if started from within an IDE). For this reason, I am fine with dropping ;-) (but from ASF perspective, this needs to be discussed on the dev@ list)

external/solr/archetype/src/main/resources/archetype-resources/src/main/java/CrawlTopology.java
mvolikas: Thanks, it's clear now. I have removed the java topologies for now. Should I add them back in and fix this at least for the solr archetype case?

mvolikas:
> We need one for the injection @mvolikas

Ok, so I guess I will add this back.

jnioche:
> > We need one for the injection @mvolikas
> 
> Ok, so I guess I will add this back.

Sorry if I wasn't clear - we need a Flux for the injection, not the Java topology

mvolikas:
An update from my side:

- I have now tested in local mode with 1 and 4 shards.
- I have updated the `SolrSpout` code so that the query param for shards is not added for just one shard.
- I tested that the compilation with the Java topologies included succeeds.
- Improved the scripts to handle the case of commented-out properties.
- After the latest commit, the following workflow worked locally without any issues for me:

  - `mvn archetype:generate -DarchetypeGroupId=org.apache.stormcrawler -DarchetypeArtifactId=stormcrawler-solr-archetype -DarchetypeVersion=3.1.1-SNAPSHOT`
  - `cd test && mvn clean compile`
  - `/opt/apache-storm-2.6.4/bin/storm local target/test-1.0-SNAPSHOT.jar  org.apache.storm.flux.Flux crawler.flux --local-ttl 3600` - the provided flux topology starts from the StormCrawler webpage URL and indexes documents in Solr.
  
Still to do/decide:

- Should we keep any of the Java topologies? (probably just the `SeedInjector.java` one)
- Testing with Storm 2.7.0. (I have tested with Storm 2.6.4 and Solr 9.7.0)
- Review the changes and READMEs one more time after the previous 2 are done.

mvolikas:
> > > We need one for the injection @mvolikas
> > 
> > 
> > Ok, so I guess I will add this back.
> 
> Sorry if I wasn't clear - we need a Flux for the injection, not the Java topology

No worries; I will add this too.

jnioche:
@mvolikas, latest comments

* need to bring in change from #1390


` storm jar target/crawler-1.0-SNAPSHOT.jar  org.apache.storm.flux.Flux injection.flux --local-ttl 3600
`

needs changing to 
`
storm local target/crawler-1.0-SNAPSHOT.jar  org.apache.storm.flux.Flux injection.flux --local-ttl 3600
`

* delete the java topology classes -> there is a discussion on whether they should be removed in the core topology but here I would advocate that we shouldn't add them in the first place. The opensearch module doesn't have them.
* remove the MemorySpout from the crawl.flux, rely on the separate injection and make sure the README reflects this, like we do in the OpenSearch module

Thanks!

mvolikas:
Hi there!

@jnioche I think I have made the changes; also pushed some comments and minor fixes to the readme files.

I ran some more tests with a greater number of shards (e.g. 10), and everything seems ok.
From my side we could merge the changes. What do you think?

jnioche (approved):
Checked with a crawl in local mode, works fine
Thanks a lot @mvolikas, this is a great contribution

PR #1373, opened 2024-10-25
#1353 Fix for URLFrontier spout not taking into account the crawl ID

### LINKED ISSUE #1353: URLFrontier spout doesn't take into account crawl Id
The URLFrontier Spout
( https://github.com/apache/incubator-stormcrawler/blob/main/external/urlfrontier/src/main/java/org/apache/stormcrawler/urlfrontier/Spout.java ) doesn't take into account the crawl Id that can be specified in the configuration parameters (URLFRONTIER_CRAWL_ID_KEY = "urlfrontier.crawlid" defined in  org.apache.stormcrawler.urlfrontier.Constants)

This results in a mix of URLs coming from distinct frontiers in URLFrontier.

klockla:
Fixes #1353 

The URLFrontier spout will pass the crawl id to GetURLs according to the value of urlfrontier.crawlid

If not set, or equal to "ANYCRAWL"', the crawl ID will not be taken into account in the URLFrontier GetURLs request
(the request will be populated with an AnyCrawlID object which in turn results in a null crawl id on the URLFrontier side).

### For all changes:
- [X] Is there a issue associated with this PR? Is it referenced in the commit message?

- [X] Does your PR title start with `#XXXX` where `XXXX` is the issue number you are trying to resolve? 

- [X] Has your PR been rebased against the latest commit within the target branch (typically main)?

- [X] Is your initial contribution a single, squashed commit?

- [X] Is the code properly formatted with `mvn git-code-format:format-code -Dgcf.globPattern="**/*" -Dskip.format.code=false`?

### For code changes:

### Note:

### DIFF
--- a/external/urlfrontier/src/main/java/org/apache/stormcrawler/urlfrontier/Spout.java
+++ b/external/urlfrontier/src/main/java/org/apache/stormcrawler/urlfrontier/Spout.java
@@ -17,7 +17,6 @@
 package org.apache.stormcrawler.urlfrontier;
 
 import static org.apache.stormcrawler.urlfrontier.Constants.URLFRONTIER_ADDRESS_KEY;
-import static org.apache.stormcrawler.urlfrontier.Constants.URLFRONTIER_ANY_CRAWL_ID;
 import static org.apache.stormcrawler.urlfrontier.Constants.URLFRONTIER_CRAWL_ID_KEY;
 import static org.apache.stormcrawler.urlfrontier.Constants.URLFRONTIER_DEFAULT_HOST;
 import static org.apache.stormcrawler.urlfrontier.Constants.URLFRONTIER_DEFAULT_PORT;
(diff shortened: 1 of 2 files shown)


external/urlfrontier/src/main/java/org/apache/stormcrawler/urlfrontier/Spout.java
```diff
@@ -131,12 +144,19 @@ protected void populateBuffer() {
...
+        GetParams.Builder builder =
                 GetParams.newBuilder()
                         .setMaxUrlsPerQueue(maxURLsPerBucket)
-                        .setMaxQueues(maxBucketNum)
                         .setDelayRequestable(delayRequestable)
-                        .build();
+                        .setMaxQueues(maxBucketNum);
+
+        if (URLFRONTIER_ANY_CRAWL_ID.equals(globalCrawlID)) {
+            builder.setAnyCrawlID(anyCrawlID);
```
jnioche, line 154: we can set it explicitly but given it is what it does by default we can probably get away with just adding a comment.
We could then remove the constant  `URLFRONTIER_ANY_CRAWL_ID` as well as `anyCrawlID` and simplify the logic?
Said differently, if a value is set use it, otherwise rely on the default behaviour?

klockla: Hi @jnioche , ok I removed references to anyCrawlId and populate the crawlId only if found in the configuration

jnioche (approved):

jnioche:
thanks @klockla

PR #1417, opened 2024-11-22
Fixes Issues mentioned in IPMC Vote

rzo1:
See details: https://lists.apache.org/thread/xljxn3901mffhyjhdx7orwy938kf2vjv


core/src/test/java/org/apache/stormcrawler/bolt/FeedParserBoltTest.java
```diff
@@ -51,7 +51,10 @@ void testFeedParsing() throws IOException {
         Metadata metadata = new Metadata();
         // specify that it is a Feed file
         metadata.setValue(FeedParserBolt.isFeedKey, "true");
-        parse("http://www.guardian.com/Feed.xml", "guardian.rss", metadata);
+        parse(
+                "https://de.wikipedia.org/w/index.php?title=Apache_Software_Foundation&action=history&feed=rss",
```
ayushtkn, line 55: can we use some page listed under apache scope instead? not sure if using wikipedia might upset some as well

rzo1: Do you have an example for an RSS feed on an apache site ?

rzo1: Alternative is I will just create something random myself.

rzo1: Let me create one for stormcrawler.apache.org (just to be save)

ayushtkn: maybe comdev has some
https://community.apache.org/blog/

the latest says
https://community.apache.org/blog/rss-feeds-activated.html

rzo1:
Replaced the RSS Feeds as well as the compressed Tripadvisor-based Sitemap tests with original stormcrawler.apache.org content.

core/src/test/java/org/apache/stormcrawler/bolt/FeedParserBoltTest.java
rzo1: Thanks. Think we are good now. I replaced it with handmade rss based upon the sitemap of our actual stormcrawler website.

ayushtkn (approved):
LGTM

external/tika/pom.xml
```diff
@@ -60,6 +60,11 @@ under the License.
 					<groupId>org.slf4j</groupId>
 					<artifactId>slf4j-log4j12</artifactId>
 				</exclusion>
+				<!-- This comes in through the Tika Image Parser module and has an incompatible BSD license -->
```
tballison, line 63: Apache Tika cleared the use of `jai-imageio-core` here: https://issues.apache.org/jira/browse/LEGAL-304.

I'm ok excluding it to make licensing cleaner for stormcrawler, and I don't think there will be large consequences for stormcrawler.

tballison (approved):

external/tika/pom.xml
rzo1: Great to have the link!

PR #1444, opened 2024-12-19
Remove log4j override

rzo1:
As discussed previously we can simply remove the log4j override since we depend on the version provided by Storm.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -116,22 +116,6 @@ under the License.
 			<groupId>org.apache.storm</groupId>
 			<artifactId>storm-client</artifactId>
 		</dependency>
-		<!-- dependency overrides for vulnerable log4j versions -->
-		<dependency>
-			<groupId>org.apache.logging.log4j</groupId>
-			<artifactId>log4j-core</artifactId>
-			<version>${log4j2.version}</version>
-		</dependency>
-		<dependency>
-			<groupId>org.apache.logging.log4j</groupId>
-			<artifactId>log4j-api</artifactId>
-			<version>${log4j2.version}</version>
-		</dependency>
-		<dependency>
-			<groupId>org.apache.logging.log4j</groupId>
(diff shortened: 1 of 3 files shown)


core/pom.xml
```diff
@@ -116,22 +116,6 @@ under the License.
 			<groupId>org.apache.storm</groupId>
 			<artifactId>storm-client</artifactId>
 		</dependency>
-		<!-- dependency overrides for vulnerable log4j versions -->
```
tballison, line 119: Some of these dependencies become "provided" instead of "compile". Are we ok with that? If yes, great, let's go forward. If not, we can at least inherit the version from storm?

For example, in stormcrawler-opensearch, these disappear: 
+- org.apache.logging.log4j:log4j-core:jar:2.24.3:compile
+- org.apache.logging.log4j:log4j-api:jar:2.24.3:compile
+- org.apache.logging.log4j:log4j-slf4j-impl:jar:2.24.3:compile

and they're now picked up via storm-client, but two deps are now "provided":
[INFO] +- org.apache.storm:storm-client:jar:2.7.1:provided
[INFO] |  +- org.apache.logging.log4j:log4j-api:jar:2.24.1:compile
[INFO] |  +- org.apache.logging.log4j:log4j-core:jar:2.24.1:provided
[INFO] |  +- org.apache.logging.log4j:log4j-slf4j-impl:jar:2.24.1:provided

For example, `org.apache.logging.log4j:log4j-core:jar:2.24.3:compile` shows up 11 times currently, but it only shows up as provided in the patched dependency tree.

rzo1: We don't have a compile dependency towards log4j, so provided should be fine afaik. A SC topology is submitted as shade to Storm and in Storm, we have log4j available on the (different) classpaths.

tballison: Thank you!

tballison (approved):

Issue #1474, opened 2025-02-21
Race condition in creating OpenSearch indices


tballison:
We should catch "index already exists" exceptions and log them. At startup, if there's more than one worker, there can be a "check then write" race condition on creating a new index. This causes the exception just mentioned.

tballison:
```
at org.opensearch.client.IndicesClient.create(IndicesClient.java:159) ~[stormjar.jar:?]
at org.apache.stormcrawler.opensearch.IndexCreation.createIndex(IndexCreation.java:101) ~[stormjar.jar:?]
at org.apache.stormcrawler.opensearch.IndexCreation.checkOrCreateIndex(IndexCreation.java:43) ~[stormjar.jar:?]
at org.apache.stormcrawler.opensearch.persistence.StatusUpdaterBolt.prepare(StatusUpdaterBolt.java:194) ~[stormjar.jar:?]
at org.apache.storm.executor.bolt.BoltExecutor.init(BoltExecutor.java:128) ~[storm-client-2.7.0.jar:2.7.0]
at org.apache.storm.executor.bolt.BoltExecutor.call(BoltExecutor.java:138) ~[storm-client-2.7.0.jar:2.7.0]
at org.apache.storm.executor.bolt.BoltExecutor.call(BoltExecutor.java:54) ~[storm-client-2.7.0.jar:2.7.0]
```

tballison:
Merged. Thank you @jnioche !

PR #1476, opened 2025-02-21
#1474 -- catch OpenSearchExceptions to prevent percolation of Runtime…

tballison:
…Exceptions, including check-then-update race condition in creating an index.

My guess is that the original intention was to catch everything from the client. However, the client can throw an OpenSearchStatusException which extends OpenSearchException which extends RuntimeException.

This catches and logs problems creating indices and templates.

The initial bug to fix was a "check-the-update" race condition if there are multiple status bolts, for example. This solution may be too broad.

I didn't add a unit test, because I _think_ this is a pretty small fix, and it didn't warrant multithreading the creation of several bolts.

Let me know what you think.

### For all changes:
- [ ] Is there a issue associated with this PR? Is it referenced in the commit message?

- [ ] Does your PR title start with `#XXXX` where `XXXX` is the issue number you are trying to resolve? 

- [ ] Has your PR been rebased against the latest commit within the target branch (typically main)?

- [ ] Is your initial contribution a single, squashed commit?

- [ ] Is the code properly formatted with `mvn git-code-format:format-code -Dgcf.globPattern="**/*" -Dskip.format.code=false`?

### For code changes:

### Note:

### DIFF
--- a/external/opensearch/src/main/java/org/apache/stormcrawler/opensearch/IndexCreation.java
+++ b/external/opensearch/src/main/java/org/apache/stormcrawler/opensearch/IndexCreation.java
@@ -85,7 +85,7 @@ public class IndexCreation {
                     client.indices().putTemplate(createIndexRequest, RequestOptions.DEFAULT);
             return createIndexResponse.isAcknowledged();
         } catch (IOException | OpenSearchException e) {
-            log.warn("template not created", e);
+            log.warn("template '{}' not created", templateName, e);
             return false;
         }
     }
@@ -108,7 +108,7 @@ public class IndexCreation {
                     client.indices().create(createIndexRequest, RequestOptions.DEFAULT);
             return createIndexResponse.isAcknowledged();
         } catch (IOException | OpenSearchException e) {
-            log.warn("index not created", e);
+            log.warn("index '{}' not created", indexName, e);
             return false;
         }
     }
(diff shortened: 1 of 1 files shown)


rzo1 (approved):

external/opensearch/src/main/java/org/apache/stormcrawler/opensearch/IndexCreation.java
```diff
@@ -78,13 +84,14 @@ private static boolean createTemplate(
             final AcknowledgedResponse createIndexResponse =
                     client.indices().putTemplate(createIndexRequest, RequestOptions.DEFAULT);
             return createIndexResponse.isAcknowledged();
-        } catch (IOException e) {
+        } catch (IOException | OpenSearchException e) {
+            log.warn("template not created", e);
```
jnioche, line 88: maybe log the name of the template at the same time?

external/opensearch/src/main/java/org/apache/stormcrawler/opensearch/IndexCreation.java
```diff
@@ -100,7 +107,8 @@ private static boolean createIndex(
             final CreateIndexResponse createIndexResponse =
                     client.indices().create(createIndexRequest, RequestOptions.DEFAULT);
             return createIndexResponse.isAcknowledged();
-        } catch (IOException e) {
+        } catch (IOException | OpenSearchException e) {
+            log.warn("index not created", e);
```
jnioche, line 111: maybe log the name of the index at the same time?

Issue #1475, opened 2025-02-21
Regular crawling should work when autodiscovery of sitemaps is turned off


tballison:
When I turned autodiscovery of sitemaps off (via `sitemap.discovery: false`), I noticed that sites with sitemaps were not getting crawled.

Hat tip and many thanks to @jnioche for confirming this as a bug and pointing to the source of the problem in the code.

tballison:
Merged. Thank you @jnioche !

PR #1477, opened 2025-02-21
#1475 -- Regular crawling should work when autodiscovery of sitemaps is turned off

tballison:
I tested this offline with https://www.cdc.gov and https://www.fda.gov.

I confirmed that when `sitemap.discovery=false`, I could set one in the seed file to `true`, and the behavior was as expected.

I also tested the opposite, where the default was `true`, but the seed for one of them was `false`, and the behavior was as expected.

I'm not sure this is the best solution. I don't like tightly coupling logic for the SitemapFilter in the FetcherBolts, but so it goes.

And, as usual, unit tests are, well, hard.

Let me know what you think.

### For all changes:
- [ ] Is there a issue associated with this PR? Is it referenced in the commit message?

- [ ] Does your PR title start with `#XXXX` where `XXXX` is the issue number you are trying to resolve? 

- [ ] Has your PR been rebased against the latest commit within the target branch (typically main)?

- [ ] Is your initial contribution a single, squashed commit?

- [ ] Is the code properly formatted with `mvn git-code-format:format-code -Dgcf.globPattern="**/*" -Dskip.format.code=false`?

### For code changes:

### Note:

### DIFF
--- a/core/src/main/java/org/apache/stormcrawler/bolt/FetcherBolt.java
+++ b/core/src/main/java/org/apache/stormcrawler/bolt/FetcherBolt.java
@@ -551,9 +551,6 @@ public class FetcherBolt extends StatusEmitterBolt {
                         smautodisco = false;
                     } else {
                         smautodisco = sitemapsAutoDiscovery;
-                        // SitemapFilter needs this
-                        metadata.setValue(
-                                SITEMAP_DISCOVERY_PARAM_KEY, Boolean.toString(smautodisco));
                     }
 
                     if (!fromCache && smautodisco) {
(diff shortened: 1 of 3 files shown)


jnioche (changes requested):
Instead of passing a configuration value through the metadata, I think we should instead make use of the [configure method ](https://github.com/apache/incubator-stormcrawler/blob/7cbacdaa36cc64a1b2123f902163c9544c33b8d8/core/src/main/java/org/apache/stormcrawler/util/AbstractConfigurable.java#L31) which all filters inherit.
We can get the value of the conf from there and if the sitemap detection is off, have a simple check at the beginning of the filter method and exit.

tballison:
Thank you @jnioche . To confirm, though, we'll still need to check the parameter on the metadata for cases where a user overrides the default behavior for a given URL.

I'll push updates shortly that puts all of the logic in the {{SitemapFilter}}.

jnioche (approved):

jnioche:
thanks @tballison

PR #1488, opened 2025-03-02
#621 Async queries in Solr

mvolikas:
#621 Support for async queries and batched updates.

Some notes:

- For the Spout, a similar approach to OpenSearch was taken: A separate method handleSuccess processes the response when the async future completes.
- http2 clients were used in all cases.
- For Solr Cloud, using the async client is convoluted and probably against the architecture of CloudSolrClient, but it was the best I could think of. Adding a test seems complicated, so I'll resort to manual testing for now.
- Performance wise, I did not notice major improvements, since the ConcurrentUpdateClient was already in use before and the gains of requestAsync were not major when running locally with limited resources.

Pending tasks:

- [x] Test async method for CloudSolrClient

### DIFF
--- a/external/solr/src/main/java/org/apache/stormcrawler/solr/SolrConnection.java
+++ b/external/solr/src/main/java/org/apache/stormcrawler/solr/SolrConnection.java
@@ -22,6 +22,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import org.apache.solr.client.solrj.SolrClient;
 import org.apache.solr.client.solrj.SolrServerException;
@@ -106,7 +107,7 @@ public class SolrConnection {
             cloud = true;
 
             CloudHttp2SolrClient.Builder builder =
-                    new CloudHttp2SolrClient.Builder(Collections.singletonList(zkHost));
+                    new CloudHttp2SolrClient.Builder(Collections.singletonList(zkHost), Optional.empty());
 
             if (StringUtils.isNotBlank(collection)) {
                 builder.withDefaultCollection(collection);
--- a/external/solr/src/main/java/org/apache/stormcrawler/solr/SolrConnection.java
+++ b/external/solr/src/main/java/org/apache/stormcrawler/solr/SolrConnection.java
@@ -17,52 +17,127 @@
 package org.apache.stormcrawler.solr;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import org.apache.solr.client.solrj.SolrClient;
 import org.apache.solr.client.solrj.SolrServerException;
+import org.apache.solr.client.solrj.impl.CloudHttp2SolrClient;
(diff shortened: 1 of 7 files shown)


external/solr/src/main/java/org/apache/stormcrawler/solr/SolrConnection.java
```diff
@@ -17,88 +17,127 @@
...
+                    }
+                }
+            }
+
+            // Shuffle the endpoints for basic load balancing
+            Collections.shuffle(endpoints);
+
+            // Get the async client
+            LBHttp2SolrClient lbHttp2SolrClient = cloudHttp2SolrClient.getLbClient();
+            LBSolrClient.Req req = new LBSolrClient.Req(request, endpoints);
```
mvolikas, line 86: From [LBHttp2SolrClient.html#requestAsync](https://solr.apache.org/docs/9_7_0/solrj/org/apache/solr/client/solrj/impl/LBHttp2SolrClient.html#requestAsync(org.apache.solr.client.solrj.impl.LBSolrClient.Req)):

> Execute an asynchronous request against one or more hosts for a given collection. The passed-in Req object includes a List of Endpoints. This method always begins with the first Endpoint in the list and if unsuccessful tries each in turn until the request is successful. Consequently, this method does not actually Load Balance. It is up to the caller to shuffle the List of Endpoints if Load Balancing is desired.

I just randomly shuffle the endpoints for now.

external/solr/src/main/java/org/apache/stormcrawler/solr/SolrConnection.java
```diff
@@ -17,88 +17,127 @@
...
-                                .withQueueSize(queueSize)
-                                .build();
-            }
-        } else {
-            throw new RuntimeException("SolrClient should have zk or solr URL set up");
-        }
 
-        return client;
-    }
+            CloudHttp2SolrClient cloudHttp2SolrClient = builder.build();
```
mvolikas, line 115: CloudHttp2SolrClient does not support requestAsync directly like Http2SolrClient, that's why I used the wrapped LBHttp2SolrClient instead.
Additionally, batched updates are not supported like in the ConcurrentUpdateSolrClient. The only workaround I can think of is implementing the batching manually.
I will submit a question to the Solr users list on whether any of those features will come with future Solr versions.

external/solr/src/main/java/org/apache/stormcrawler/solr/SolrConnection.java
```diff
@@ -17,88 +17,127 @@
...
+
+            // Shuffle the endpoints for basic load balancing
+            Collections.shuffle(endpoints);
+
+            // Get the async client
+            LBHttp2SolrClient lbHttp2SolrClient = cloudHttp2SolrClient.getLbClient();
+            LBSolrClient.Req req = new LBSolrClient.Req(request, endpoints);
+
+            return lbHttp2SolrClient
+                    .requestAsync(req)
```
mvolikas, line 89: An alternative would be to use the synchronous request method wrapped in a CompletableFuture.supplyAsync so that it does not block. This, however, will block another thread instead.

mvolikas:
I have now tested the cloud version, and after some fixes, it looks good. If there are no objections, I would like to merge this and follow up with a PR on cloud-related improvements (after some feedback from the Solr users list 🧐, of course)

jnioche (approved):

jnioche:
Great, thanks @mvolikas

PR #1544, opened 2025-05-31
#621 add batching for cloud updates, fix cloud requests

mvolikas:
#621 Batching for Solr Cloud updates and other Cloud improvements

- I have tested this in Solr Cloud with multiple shards for the status collection, and 2 replicas.
- I tried to make the minimum set of changes to have "async" support for the Cloud client while keeping most of the advantages offered by solrj out of the box. Of course, this approach is not as robust as the solrj CloudSolrClient, but I think it is enough in the context of crawling.

### DIFF
--- a/external/solr/src/main/java/org/apache/stormcrawler/solr/SolrConnection.java
+++ b/external/solr/src/main/java/org/apache/stormcrawler/solr/SolrConnection.java
@@ -113,7 +113,14 @@ public class SolrConnection {
                 if (waitingUpdates.isEmpty()) continue;
 
                 Slice slice = col.getSlice(entry.getKey());
-                flushUpdates(slice, waitingUpdates, cloudHttp2SolrClient);
+                Replica leader = slice.getLeader();
+
+                if (leader == null) {
+                    LOG.error("Could not find the leader for slice {}", slice.getName());
+                    return;
+                }
+
+                flushUpdates(leader, waitingUpdates, cloudHttp2SolrClient);
             }
         }
     }
@@ -145,8 +152,15 @@ public class SolrConnection {
                     updateQueues.computeIfAbsent(slice.getName(), k -> new ArrayList<>());
             waitingUpdates.add(update);
 
-            if (waitingUpdates.size() == updateQueueSize) {
-                flushUpdates(slice, waitingUpdates, cloudHttp2SolrClient);
+            if (waitingUpdates.size() >= updateQueueSize) {
+                Replica leader = slice.getLeader();
+
+                if (leader == null) {
+                    LOG.error("Could not find the leader for slice {}", slice.getName());
+                    return;
+                }
+
+                flushUpdates(leader, waitingUpdates, cloudHttp2SolrClient);
             }
         }
     }
@@ -156,12 +170,10 @@ public class SolrConnection {
      * leader goes down before handling it.
      */
     private void flushUpdates(
-            Slice slice, List<Update> waitingUpdates, CloudHttp2SolrClient cloudHttp2SolrClient) {
-        Replica leader = slice.getLeader();
-        if (leader == null) {
-            LOG.error("Could not find the leader for slice {}", slice.getName());
-            return;
-        }
+            Replica leader,
+            List<Update> waitingUpdates,
+            CloudHttp2SolrClient cloudHttp2SolrClient) {
+
         List<LBSolrClient.Endpoint> endpoints = new ArrayList<>();
         endpoints.add(new LBSolrClient.Endpoint(leader.getBaseUrl(), leader.getCoreName()));
 
@@ -180,6 +192,8 @@ public class SolrConnection {
         UpdateRequest updateRequest = new UpdateRequest();
         updateRequest.add(docs);
         updateRequest.deleteById(deletionIds);
+
+        List<Update> batch = new ArrayList<>(waitingUpdates);
         waitingUpdates.clear();
 
         // Get the async client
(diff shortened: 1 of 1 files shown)


external/solr/src/main/java/org/apache/stormcrawler/solr/SolrConnection.java
```diff
@@ -32,55 +35,213 @@
...
+            if (waitingUpdates.size() == updateQueueSize) {
+                flushUpdates(slice, waitingUpdates, cloudHttp2SolrClient);
+            }
+        }
+    }
+
+    /**
+     * Flush all waiting updates for this slice to the slice leader The request will fail, if the
+     * leader goes down before handling it
+     */
```
mvolikas, line 152: We could add a retry policy to avoid this, but it will complicate the code even more. Additionally, missed Solr updates should not cause long-term issues in the crawl, since pages get refetched after some time, or am I missing something here?

external/solr/src/main/java/org/apache/stormcrawler/solr/SolrConnection.java
```diff
@@ -32,55 +35,213 @@
...
+    private final String collection;
+
+    private long lastUpdate;
+    private final Map<String, List<Update>> updateQueues;
+    private final Object lock = new Object();
+
+    private final int updateQueueSize = 10;
+
+    // Maximum time (ms) without updates before we flush all the queues
+    private final long noUpdateThreshold = 20_000;
```
mvolikas, line 65: I guess these should be external properties. Any name recommendations?

rzo1:
I am wondering, if it would be possible, to add some (simple) Testcontainer based unit test? I guess, it would need zookeeper for cloud mode? Maybe there is a way to do it, because it would be really good to have some coverage :)

external/solr/src/main/java/org/apache/stormcrawler/solr/SolrConnection.java
rzo1: Maybe something like: `flushAfterNoUpdatesMillis`

external/solr/src/main/java/org/apache/stormcrawler/solr/SolrConnection.java
rzo1: They only get re-fetched, if the user has enabled re-fetch on error or didn't disable re-fetching at all :)

external/solr/src/main/java/org/apache/stormcrawler/solr/SolrConnection.java
```diff
@@ -32,55 +35,213 @@
...
-    private boolean cloud;
-    private String collection;
+    private final boolean cloud;
+    private final String collection;
+
+    private long lastUpdate;
+    private final Map<String, List<Update>> updateQueues;
+    private final Object lock = new Object();
+
+    private final int updateQueueSize = 10;
```
rzo1, line 62: Should that be a parameter too?

mvolikas: Yes, `batchUpdateSize`?

external/solr/src/main/java/org/apache/stormcrawler/solr/SolrConnection.java
mvolikas: Hmm, then I will add a simple retry policy to be on the safe side.

mvolikas:
> I am wondering, if it would be possible, to add some (simple) Testcontainer based unit test? I guess, it would need zookeeper for cloud mode? Maybe there is a way to do it, because it would be really good to have some coverage :)

In principle, yes, but I had difficulties running everything in containers because the Zookeeper container advertises something like "zookeeper:2181" to the Solr containers inside the Docker network, and from the host (where StormCrawler runs), you cannot resolve "zookeeper". I will give it another try.

rzo1:
> > I am wondering, if it would be possible, to add some (simple) Testcontainer based unit test? I guess, it would need zookeeper for cloud mode? Maybe there is a way to do it, because it would be really good to have some coverage :)
> 
> In principle, yes, but I had difficulties running everything in containers because the Zookeeper container advertises something like "zookeeper:2181" to the Solr containers inside the Docker network, and from the host (where StormCrawler runs), you cannot resolve "zookeeper". That being said, I agree that we should have some tests for Solr Cloud, and I will give it another try.

The guys from Apache Camel had a similar issue: https://github.com/apache/camel-quarkus/tree/e5e768acd680b0d78122fb7eee30b0a70947f3f9/integration-tests/solr/src/test - looks like their setup worked (somehow) - maybe we can just port it for our needs?

They bascially rely on docker compose (in two variants) to spin it up. Maybe worth a look?

rzo1:
We can add tests later for it, so no direct blocker from my side

external/solr/src/main/java/org/apache/stormcrawler/solr/SolrConnection.java
mvolikas: I have added logic that re-queues update batches that failed (e.g. the shard leader we sent the update to went down before handling the request). The batch will be dealt with the next time a flush is triggered.

external/solr/src/main/java/org/apache/stormcrawler/solr/SolrConnection.java
```diff
@@ -32,55 +35,237 @@
...
+
+            if (statusCollection) {
+                // We have multiple shards, find the correct one
+                slice = getSlice(((DocUpdate) update).doc, col);
+            } else {
+                slice = col.getActiveSlices().stream().findFirst().orElse(null);
+            }
+
+            if (slice == null) {
+                LOG.error("Could not find an active slice for update {}", update);
```
mvolikas, line 146: This should not happen unless a shard is being split, which, in turn, should not happen in the context of StormCrawler while the crawl is running.

rzo1 (approved):

PR #1565, opened 2025-06-16
Enable GH discussions

rzo1:
as the title says. They were enabled before incubation and we don't have a user list, so might be good to have it for user equestions.


pjfanning (approved):
lgtm

rzo1:
Thanks PJ!

PR #1586, opened 2025-07-01
Bump dependencies version (Commons CLI 1.9.0, OpenSearch 2.19.2)

sigee:
Updates to Jackson-annotations 2.19.1
Updates to Jackson-core 2.19.1
Updates to jackson-databind 2.19.1

Updates to Apache Commons CLI 1.9.0

Updates to opensearch-rest-high-level-client 2.19.2
Updates to opensearch-rest-client-sniffer 2.19.2

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -47,7 +47,7 @@ under the License.
 		<wiremock.version>3.13.1</wiremock.version>
 		<rometools.version>2.1.0</rometools.version>
 		<selenium.version>4.33.0</selenium.version>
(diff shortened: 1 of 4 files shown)


rzo1 (commented):
> As for OpenSearch, there are various places where the version should be changed: README, archetypes etc... maybe grep for the current version number to see where it is used

Think we fixed it earlier. Did a quick search and couldn't find anything.

We should update AbstractOpenSearchTest#OPENSEARCH_VERSION with the new version though.

Rest lgtm.

PR #1609, opened 2025-07-21
Fix java 5 language level issues

sigee:


mvolikas (commented):
There are still some places where we use raw types. One example is [here](https://github.com/sigee/stormcrawler/blob/fix-java5-issues/external/opensearch/src/main/java/org/apache/stormcrawler/opensearch/metrics/MetricsConsumer.java#L72).

rzo1 (approved):

Issue #1604, opened 2025-07-16
Decouple Selenium and OkHttp from Core Module


rzo1:
Hi everyone,

Would it make sense to move the Selenium and OkHttp implementations into separate modules (similar to playwright)? This way, those who need them can still include them via Maven, while helping to reduce the overall footprint of the core framework.

What do you think?

jnioche:
+1

jnioche:
I would not add okhttp into a separate module: it is far more robust and efficient than httpclient, which is why it is used in the archetypes

https://github.com/apache/stormcrawler/blob/main/archetype/src/main/resources/archetype-resources/crawler-conf.yaml#L72

It is not the default value in [crawler-default](https://github.com/apache/stormcrawler/blob/f392ddd1c0eb9db0b72dd4e9e03e190537bc5b86/core/src/main/resources/crawler-default.yaml#L170) - I thought this was the case

rzo1:
So we should change the default? ;-)

jnioche:
> So we should change the default? ;-)

yes, probably

PR #1646, opened 2025-09-02
#1604 - Externalise Selenium

jnioche:
Moves the Selenium code to `external/selenium`.

The README for the new module should be improved (contributions welcome)


external/selenium/pom.xml
```diff
@@ -0,0 +1,108 @@
...
+	<name>stormcrawler-selenium</name>
+	<url>https://github.com/apache/stormcrawler/tree/master/external/selenium</url>
+	<description>selenium-based protocol for StormCrawler</description>
+
+	<properties>
+		<selenium.version>4.35.0</selenium.version>
+		<awaitility.version>4.3.0</awaitility.version>
+		<wiremock.version>3.13.1</wiremock.version>
+		<jacoco.haltOnFailure>true</jacoco.haltOnFailure>
+		<jacoco.classRatio>0.00</jacoco.classRatio>
```
sigee, line 44: I know it is annoying when the coverage drops and this verification drops, but you can set the jacoco ratio limits a bit higher.

classRatio: 1.00
There is 1.00 class ratio, because all of the 4 classes are affected. (tbh, I'm not sure if it is right or not, because there is only one single test case, but anyway. This is the actual class ratio.
instructionRatio: 0.69
methodRatio: 0.80
branchRatio: 0.62
lineRatio: 0.66
complexityRatio: 0.55

Or if you want to disable this for some reasons, you only need to set haltOnFailure to false and don't need to override every single ratio.

external/selenium/src/main/java/org/apache/stormcrawler/protocol/selenium/RemoteDriverProtocol.java
sigee, line 1: I'm not sure why this file changed back to an older version.
The for loop is cleaner than the while with the iterator and easier to read.
The raw use of List is recommended. You should use the parameterised type declaration and you can use the diamond for value assignment.

jnioche: Strange, the branch has been merged with main though. But yes, good points.
Thanks @sigee

sigee (approved):
LGTM

external/selenium/README.md
```diff
@@ -0,0 +1,6 @@
+# Selenium
```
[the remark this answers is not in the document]
rzo1: GitHub gets crapped if using multiple escape lines in a code suggestion ...

Issue #711, opened 2019-04-26
MetadataFilter to filter based on multiple key values


jnioche:
see #710 to filter docs if _isSitemap=false_ and _foundSitemap=true_

sigee:
What would be the expected solution for this?
At the moment MetadataFilter works as if any of the parameters match, then filters.
Do we want to change this isSitemap=false **OR** foundSitemap=true to an **AND**?
Or somehow we want to add options to select operation **OR**/**AND**?

jnioche:
thanks @sigee 
it has been a while but IIRC what I wanted to do was to be able to specify the operator

PR #1647, opened 2025-09-02
Improve MetadataFilter

sigee:
This pull request suppose to address the #711 issue.

I created a complex filtering logic where any operation can be assembled.
E.g.: I created test cases where these are the filters
- key=>val OR (key2=>val2 AND key3=>val3)
- key=>val AND (key2=>val2 OR key3=>val3)

What do you think?


rzo1:
> > how to specify in a config
> 
> The config did not change. I mean from json config there is no option to specify the operation. I left that part as it was for backward compatibily. So from json it can be only 1 level OR operation. If you want to create an AND operation or a complex filter you have to assembe it from code. At least for now, but if it is neccessary I can figure out a json structure for set the operation too, but probably it won't be backward compatible.

I think it would be useful to make it available in JSON, since that’s how most people will define URL filters.
Breaking backward compatibility would be acceptable in my opinion (or alternatively, a fallback could be provided in the implementation).

PR #1651, opened 2025-09-07
#1650 Replace Tika's auto-detector with magic marker detection

### LINKED ISSUE #1650: Tika's detector breaks mime type guessing
### Version

main branch

### Describe what's wrong

Tika's detector (used by [JsoupParserBolt](https://github.com/apache/stormcrawler/blob/main/core/src/main/java/org/apache/stormcrawler/bolt/JSoupParserBolt.java#L500)) seems to treat every file as potentially being an archive file and then fails because it actually isn't.
Related to https://issues.apache.org/jira/browse/TIKA-4469.
Comment by @rzo1 in dev list:

> In the end, we should not use Tika's Detector but a TikaInputStream instead like that:
> try (TikaInputStream tis = TikaInputStream.get(data)) { final Metadata metadata = new Metadata(); metadata.add(TikaCoreProperties.RESOURCE_NAME_KEY, file.getFileName()); final MediaType mediaType = MimeTypes.getDefaultMimeTypes().detect(tis, metadata); 

### Error message and/or stacktrace

> Exception while guessing mimetype on https://apache.org/: org.apache.commons.compress.archivers.ArchiveException: No Archiver found for the stream signature

### How to reproduce

Run a crawl with the single seed URL "https://apache.org/".

### Additional context

_No response_

mvolikas:
Fixes #1650.

### DIFF
--- a/core/src/main/java/org/apache/stormcrawler/bolt/JSoupParserBolt.java
+++ b/core/src/main/java/org/apache/stormcrawler/bolt/JSoupParserBolt.java
@@ -18,9 +18,7 @@ package org.apache.stormcrawler.bolt;
 
 import static org.apache.stormcrawler.Constants.StatusStreamName;
 
-import java.io.ByteArrayInputStream;
 import java.io.IOException;
-import java.io.InputStream;
 import java.lang.reflect.InvocationTargetException;
 import java.net.MalformedURLException;
 import java.net.URL;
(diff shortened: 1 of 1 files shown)


rzo1:
Maybe we can try to find out if we can update commons compress first (or check what is in our classpath) as suggested by @tballison ? :)

rzo1:
see https://github.com/apache/stormcrawler/pull/1652

Issue #1698, opened 2025-10-20
Make waitAck Cache Configurable in StatusUpdaterBolt


rzo1:
Allow the waitAck cache in [StatusUpdaterBolt](https://github.com/apache/stormcrawler/blob/main/external/opensearch/src/main/java/org/apache/stormcrawler/opensearch/persistence/StatusUpdaterBolt.java#L163)  to be configured via a cache specification string (similar to how AbstractStatusUpdaterBolt handles its internal cache), instead of being hard-coded.

In the current implementation, waitAck is initialized as follows:

```
waitAck = Caffeine.newBuilder()
    .expireAfterWrite(60, TimeUnit.SECONDS)
    .removalListener(this)
    .build();
```

This configuration is static: both the expiration time and cache size are fixed. Users cannot adjust it through configuration. Under heavy indexing loads or high latency in OpenSearch responses, this can lead to:

- Premature eviction of pending tuples.
- “Purged from waitAck” errors in logs.
- Storm replaying tuples and OpenSearch returning 409 CONFLICT or 429 TOO MANY REQUESTS responses.

**Idea**:

Add a configuration parameter (e.g. `opensearch.status.updater.waitack.cache.spec`) that allows the cache to be built from a Caffeine specification string, just like the one used in `AbstractStatusUpdaterBolt`.

rzo1:
Feel free.

PR #1700, opened 2025-10-20
#1698 OpenSearch: Make waitAck cache configurable via cache specifica…

HyemIin:

### For all changes

### For code changes

### Note

### DIFF
--- a/external/opensearch/src/test/java/org/apache/stormcrawler/opensearch/bolt/StatusBoltTest.java
+++ b/external/opensearch/src/test/java/org/apache/stormcrawler/opensearch/bolt/StatusBoltTest.java
@@ -16,7 +16,7 @@
  */
 package org.apache.stormcrawler.opensearch.bolt;
 
-import static org.junit.Assert.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;


external/opensearch/src/main/java/org/apache/stormcrawler/opensearch/persistence/StatusUpdaterBolt.java
```diff
@@ -160,11 +159,13 @@ public void prepare(
...
-        waitAck =
-                Caffeine.newBuilder()
-                        .expireAfterWrite(60, TimeUnit.SECONDS)
-                        .removalListener(this)
-                        .build();
+        String waitAckSpec =
+                ConfUtils.getString(
+                        stormConf,
+                        "opensearch.status.waitack.cache.spec",
+                        "maximumSize=10000,expireAfterWrite=60s");
```
rzo1: see comment above, might be inlined in `Caffeine.from(...)` as well.

external/opensearch/src/test/java/org/apache/stormcrawler/opensearch/bolt/StatusBoltTest.java
```diff
@@ -128,23 +124,27 @@ private Future<Integer> store(String url, Status status, Metadata metadata) {
     }
 
     @Test
-    @Timeout(value = 2, unit = TimeUnit.MINUTES)
```
rzo1: Why is this test removed? If it fails, we should adjust it (or provide a reasoning why it was dropped)

external/opensearch/src/test/java/org/apache/stormcrawler/opensearch/bolt/StatusBoltTest.java
```diff
@@ -128,23 +124,27 @@ private Future<Integer> store(String url, Status status, Metadata metadata) {
...
+
+        StatusUpdaterBolt bolt = new StatusUpdaterBolt();
+        try {
+            bolt.prepare(conf, mockContext, mockCollector);
+        } catch (RuntimeException e) {
+            // 연결 실패 시 예외 분기 발생 → Jacoco branch coverage 확보
+            assertTrue(e.getMessage().contains("Can't connect"));
+        }
+
+        Field field = StatusUpdaterBolt.class.getDeclaredField("waitAck");
```
rzo1: What are we trying to achieve here? Why do we check for the classname? If the class can't be initalized, the bolt is merely useless (so we should fail early), so no need for such a test imho.

external/opensearch/src/test/java/org/apache/stormcrawler/opensearch/bolt/StatusBoltTest.java
HyemIin: It was my bad. 
I was trying to fix the Jacoco coverage issue and thought the test case was missing.
I’ll restore the original test case and include them in this commit.

rzo1:
Yes. You have typically a cascade of YAML files when submitting a topology via flux for example, so it doesn't need to be defined in the OS specific config.

jnioche (changes requested):

external/opensearch/src/main/java/org/apache/stormcrawler/opensearch/persistence/StatusUpdaterBolt.java
```diff
@@ -160,11 +159,16 @@ public void prepare(
             fieldNameForRoutingKey = fieldNameForRoutingKey.replaceAll("\\.", "%2E");
         }
 
-        waitAck =
-                Caffeine.newBuilder()
-                        .expireAfterWrite(60, TimeUnit.SECONDS)
-                        .removalListener(this)
-                        .build();
+        String defaultSpec = String.format("maximumSize=10000,expireAfterWrite=%ds",
```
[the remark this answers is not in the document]
HyemIin: I was considering a scenario where waitAck entries might not be properly removed due to an unknown issue in the ACK handling logic, which could cause the cache to grow indefinitely and potentially lead to an OOM (OutOfMemory) error. In that sense, maximumSize could serve as a last line of defense.

However, since such a situation is highly unlikely in practice, I agree that it’s unnecessary and removing it, as you suggested, is the right approach.

rzo1 (approved):

Issue #1662, opened 2025-09-19
[Improvement] Work towards Maven 4 Compatibility


rzo1:
### What would you like to be improved?

We should work on Maven 4 Compatbility

### How should we improve?

see https://github.com/gnodet/maven4-testing/issues/7376

mvolikas:
Some useful links:

- [What's new](https://maven.apache.org/whatsnewinmaven4.html)
- [Migration guide](https://maven.apache.org/guides/mini/guide-migration-to-mvn4.html)
- [4.0.0 RCs](https://downloads.apache.org/maven/maven-4/)

The goal would be making the minimum changes so that SC can be build with the latest Maven 4 RC, right?

rzo1:
Yes. We should be ready for 4 :-) (but still be able to build with 3 for now)

PR #1716, opened 2025-11-09
#1662 - Migrate to Maven 4

mvolikas:
#1662 Changes for Maven 4 compatibility following the official migration guidelines [1](https://maven.apache.org/whatsnewinmaven4.html)|[2](https://maven.apache.org/guides/mini/guide-migration-to-mvn4.html).

Steps followed:

1. I started by installing [Apache Maven 4.0.0-rc-4](https://downloads.apache.org/maven/maven-4/) locally.
2. Used the built in upgrade tool ([mvnup](https://maven.apache.org/tools/mvnup.html)) to make the initial changes with `mvnup apply --model-version 4.1.0 --all`
3. Removed as much explicit sub-module definitions as possible. Only two were really necessary because of the nested archetype modules.
4. Followed the new conventions for parent inference.
5. Built with Maven 4 and reviewed the generated targets including the new consumer POMs.


mvolikas:
In some POMs we're using tabs for indentation, while in others we have double spaces. Should we agree on a common style and update them?

rzo1:
Thx. Can we update the action to run with an actual Maven 4 binary? Most likely other ASF projects are doing it similar or https://github.com/s4u/setup-maven-action?tab=readme-ov-file or alike are already Maven 4 ready?

I have no preference regarding formatting.

external/pom.xml
```diff
@@ -19,20 +19,30 @@ specific language governing permissions and limitations
...
-	</parent>
+<project xmlns="http://maven.apache.org/POM/4.1.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.1.0 https://maven.apache.org/xsd/maven-4.1.0.xsd">
+	
+	<parent/>
 
 	<artifactId>stormcrawler-external</artifactId>
 	<packaging>pom</packaging>
 
-    <dependencies>
+	<subprojects>
```
mvolikas: In RC 5, you can no longer have `submodules` together with jar packaging, so I updated accordingly.

PR #1720, opened 2025-11-15
Fix formatting issues reported by checkstyle (Part 1)

sigee:


core/src/main/java/org/apache/stormcrawler/parse/DocumentFragmentBuilder.java
```diff
@@ -93,8 +94,9 @@ private void copyAttributes(org.jsoup.nodes.Node source, Element el) {
             for (Attribute attribute : source.attributes()) {
                 // valid xml attribute names are: ^[a-zA-Z_:][-a-zA-Z0-9_:.]
                 String key = attribute.getKey().replaceAll("[^-a-zA-Z0-9_:.]", "");
```
rzo1: I guess the patterns below would be candidates for pre-compiled pattern.

PR #1728, opened 2025-11-17
Added PiiBolt and PresidioRedacter implementation

klockla:
This PR adds a bolt which allows removal of Personally Identifiable Information (PII).
The PiiBolt is to be used with a class implementing the PiiInterface and which will provide the actual implementation of PII.

This PR implements also the PresidioRedactor class which uses Microsoft Presidio
( https://microsoft.github.io/presidio/ ) as a PII back-end.
It can be configured for different PII entities (names, phones, location, etc...) and different languages according to how you deployed the back-end.

### DIFF
--- a/core/src/main/java/org/apache/stormcrawler/pii/PiiBolt.java
+++ b/core/src/main/java/org/apache/stormcrawler/pii/PiiBolt.java
@@ -1,0 +1,195 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.stormcrawler.pii;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.storm.task.OutputCollector;
+import org.apache.storm.task.TopologyContext;
+import org.apache.storm.topology.OutputFieldsDeclarer;
+import org.apache.storm.topology.base.BaseRichBolt;
+import org.apache.storm.tuple.Fields;
+import org.apache.storm.tuple.Tuple;
+import org.apache.storm.tuple.Values;
+import org.apache.stormcrawler.Metadata;
+import org.apache.stormcrawler.util.ConfUtils;
+import org.apache.stormcrawler.util.InitialisationUtil;
+import org.slf4j.LoggerFactory;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Map;
+
+/**
+ * StormCrawler bolt that performs PII redaction on the content of web pages
+ * before they are passed to the indexing or persistence bolt.<br>
+ * If enabled, the HTML content will be overwritten with a dummy HTML page (containing just "REDACTED")<br><br>
+ * <b>pii.redacter.class</b> is the name of the class implementing the PiiInterface interface (e.g. org.apache.stormcrawler.pii.PresidioRedacter)<br>
+ * <b>pii.language.field</b>, if set, allows to set the name of a Metadata field that contains the language to be passed to the PII redacter instance
+ * 
+ */
+@SuppressWarnings("serial")
+public class PiiBolt extends BaseRichBolt {
+
+	private static final org.slf4j.Logger LOG = LoggerFactory.getLogger(PiiBolt.class);
+
+	/*
+	 *  Name of config field defining the PII Redacter class
+	 * (This class must implement the PiiRedacter interface
+	 */
+	public static final String PII_REDACTER_CLASS_PARAM = "pii.redacter.class";
+
+	/*
+	 * Name of the field for configurating language detection
+	 */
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/apache/stormcrawler/pii/PiiBolt.java
```diff
@@ -0,0 +1,194 @@
...
+					piiRedacter.redact(text);
+
+			if (redacted == null) {
+				throw new Exception("PII Redacter returned null");
+			}
+
+			metadata.addValue("pii.processed", "true");
+			
+			// Force the binary content to a dummy content
+			emitTuple(input, url, REDACTED_BYTES, metadata, redacted);
```
rzo1: Why do we force the binary content to dummy content here, if we have `redacted`? Shouldn't we set it to `redacted.getBytes(...)` ?

core/src/main/java/org/apache/stormcrawler/pii/PiiBolt.java
```diff
@@ -0,0 +1,194 @@
...
+		} catch (Exception e) {
+			LOG.error("Error during PII redaction for URL {}: {}", url, e.getMessage());
+			metadata.addValue("pii.error", e.getMessage());
+			
+			// How to handle the content in case of error ?
+			emitTuple(input, url, originalBytes, metadata, text);
```
rzo1: Fail soft imho. So just return the original data + the added metadata?

core/src/test/java/org/apache/stormcrawler/pii/PiiBoltTest.java
```diff
@@ -0,0 +1,202 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.stormcrawler.pii;
+
+import static org.junit.jupiter.api.Assertions.*;
```
rzo1: please avoid star imports.

external/presidio/README.md
```diff
@@ -0,0 +1,59 @@
+# stormcrawler-aws
```
rzo1: This README needs an update. Looks wrong to me.

external/presidio/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+	<url>https://github.com/apache/stormcrawler/tree/master/external/presidio</url>
+	<description>Interface with Microsoft Presisio for Data Protection and De-identification</description>
+
+	<properties>
+		<aws.version>1.12.792</aws.version>
+	</properties>
+
+	<dependencies>
+		<!-- Java 8 Datatypes -->
+		<dependency>
```
rzo1: For what reason do we need jdk8 datatypes? We are on Java 11 as a baseline atm, so no need for that one, imho with normal ObjectMapper Jackson usage?

core/src/main/java/org/apache/stormcrawler/pii/PiiBolt.java
```diff
@@ -0,0 +1,194 @@
...
+	private static final String FIELD_URL = "url";
+	private static final String FIELD_CONTENT = "content";
+	private static final String FIELD_METADATA = "metadata";
+	private static final String FIELD_TEXT = "text";
+
+
+	// Default value for language metadata field
+	private String languageFieldName = "parse.lang";
+
+	OutputCollector _collector;
```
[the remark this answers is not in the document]
klockla: > and why in core?

My idea was to have the abstraction layer in core and the specific implementation(s) in external. Please let me know where to move it

klockla:
> Hi @klockla,
> 
> Thanks for the PR and for proposing an abstraction for PII removal during crawls. I’ve added a few comments.
> 
> I also have a couple of questions:
> 
> What was the reasoning behind choosing a bolt instead of a (parse) filter for the redaction step?
> 
> Since this is a larger contribution, we’ll likely need an [ICLA](https://www.apache.org/licenses/contributor-agreements.html) before we can accept it.

Hi @rzo1 

I didn't really think about implementing it as a parse filter but as the process (text analysis by the NLP engine in Presidio) is quite consuming, I think it may be better to have this in a separate bolt to have better measures about tuple processing time/latency. 

I will fix the points related to your other comments and will need to have a look at this ICLA thing.

core/src/main/java/org/apache/stormcrawler/pii/PiiBolt.java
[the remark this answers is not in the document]
rzo1: I was looking at `OutputCollector _collector;` :)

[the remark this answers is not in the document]
klockla: ok, will change it to protected

external/presidio/pom.xml
rzo1: Current version of Jackson in SC is **2.18.1** (always aligned with Storm), but if it is needed, I am ok with it.

klockla: > Current version of Jackson in SC is **2.18.1** (always aligned with Storm), but if it is needed, I am ok with it.

yes, sorry the 11 was a typo

jnioche:
> I didn't really think about implementing it as a parse filter but as the process (text analysis by the NLP engine in Presidio) is quite consuming, I think it may be better to have this in a separate bolt to have better measures about tuple processing time/latency.

I haven't looked at the details yet but I agree with @rzo1 that this feels like it should be a ParseFilter.

We log [processing times in ParseFilters](https://github.com/apache/stormcrawler/blob/15b09c1a3b92d500146365ba2db7c39df97ffbc7/core/src/main/java/org/apache/stormcrawler/parse/ParseFilters.java#L146) but could extend that to send proper metrics

jnioche:
> @jzonthemtn Could you take a look, particularly regarding how well the abstraction in "core" interoperates with other PII-redaction engines (such as Philter)?

https://github.com/philterd/philter

PR #1893, opened 2026-04-29
Bump mysql-connector-j to 9.7.0 and fix flaky scheduler test

dpol1:
As the title say, fixes #1889


jnioche (approved):
I would have put the fix in a separate PR from the dependency update but given it is not a large change, I think this is acceptable

jnioche:
thanks @dpol1

PR #1897, opened 2026-05-04
feat(playwright): add PageAction extension point with built-in actions

rzo1:
Introduces a configurable post-navigate chain loaded from JSON via playwright.page.actions.config.file, so site-specific DOM transformations no longer require subclassing the protocol. Ships built-ins for overlay dismissal, clickable expansion, scroll-to-bottom, wait-for-selector, JS evaluation, and screenshot capture.


jnioche:
This is great, reminds me of the NavigationFilter mechanism we have for Selenium.

jnioche (approved):

rzo1:
> This is great, reminds me of the NavigationFilter mechanism we have for Selenium.

Yes. It is inspired by it and we faced the lack of it in Playwright during Lisa's bachelor thesis :)

PR #1926, opened 2026-06-02
Fix missing space between block and inline sibling elements in JSoupTextExtractor

### LINKED ISSUE #1925: JSoupTextExtractor: missing space between block and inline sibling elements
When extracting text from HTML, `JSoupTextExtractor` does not insert a space between a block-level element and an immediately following inline element.

For example, given this HTML from https://www.acai-island.com/contact:

```html
<div>
  <h3>Email</h3>
  <a href="mailto:acaiisland1300@gmail.com">acaiisland1300@gmail.com</a>
</div>
```

The extracted text is `Emailacaiisland1300@gmail.com` instead of `Email acaiisland1300@gmail.com`.

The issue is in the `tail()` method of the `NodeVisitor` inside `JSoupTextExtractor.text()`. Currently, a space is only appended after a block element when the next sibling is a `TextNode`:

```java
if (element.isBlock()
        && (node.nextSibling() instanceof TextNode)
        && !lastCharIsWhitespace(accum)) {
    accum.append(' ');
}
```

This misses cases where the next sibling is an inline `Element` (like `<a>`, `<span>`, etc.), which causes the text content of the block and the inline element to be concatenated without any separator.

This also affects other common patterns like `<div>Phone</div><span>(631) 656-0088</span>` producing `Phone(631) 656-0088`.

nicoscandolo:
Fixes #1925

**Problem**

When a block-level element (e.g. `<h3>`) is immediately followed by an inline element (e.g. `<a>`, `<span>`), the extracted text concatenates both without any space. For instance, this HTML from a real page (https://www.acai-island.com/contact):

```html
<h3>Email</h3>
<a href="mailto:acaiisland1300@gmail.com">acaiisland1300@gmail.com</a>
```

Produces `Emailacaiisland1300@gmail.com` instead of `Email acaiisland1300@gmail.com`.

**Root cause**

In `tail()`, the space after a block element is only appended when `nextSibling() instanceof TextNode`. This misses cases where the next sibling is an inline Element.

**Fix**

Changed the condition to `nextSibling() != null`, so a space is appended after any block element regardless of the sibling type. The existing `lastCharIsWhitespace` guard prevents duplicate spaces.

**Tests**

Added two test cases covering block+inline combinations (`<h3>`+`<a>` and `<div>`+`<span>`).

### DIFF
--- a/core/src/main/java/org/apache/stormcrawler/parse/JSoupTextExtractor.java
+++ b/core/src/main/java/org/apache/stormcrawler/parse/JSoupTextExtractor.java
@@ -139,7 +139,7 @@ public class JSoupTextExtractor implements TextExtractor {
 
                     public void tail(Node node, int depth) {
                         // make sure there is a space between block tags and immediately
-                        // following text nodes <div>One</div>Two should be "One Two".
+                        // following siblings <div>One</div><a>Two</a> should be "One Two".
                         if (node instanceof Element) {
                             Element element = (Element) node;
                             if (element == excluded) {
--- a/core/src/main/java/org/apache/stormcrawler/parse/JSoupTextExtractor.java
+++ b/core/src/main/java/org/apache/stormcrawler/parse/JSoupTextExtractor.java
@@ -146,7 +146,7 @@ public class JSoupTextExtractor implements TextExtractor {
                                 excluded = null;
                             }
                             if (element.isBlock()
-                                    && (node.nextSibling() instanceof TextNode)
+                                    && node.nextSibling() != null
                                     && !lastCharIsWhitespace(accum)) {
                                 accum.append(' ');
                             }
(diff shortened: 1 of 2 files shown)


rzo1 (commented):
Thanks for the fix. Overall, it looks good to me.

The only thing I would like to change is, that tests/comments embed a real small business's website and email address (acaiisland1300@gmail.com, acai-island.com). For an Apache codebase that lives forever, I'd suggest anonymizing to info@example.com / generic text. The issue reference (#1925) already preserves the real-world provenance.

In addition, the existing comment above the condition still reads "a space between block tags and   immediately following text nodes" - it could be updated to "following siblings" to match the new behavior.

rzo1 (approved):

dpol1 (approved):
LGTM. `lastCharIsWhitespace` already guards against double spaces so widening to `!= null` is safe. comment + tests good.

jnioche (approved):
thanks @nicoscandolo and our reviewers

PR #1928, opened 2026-06-04
Fix thread safety, resource leaks, and error handling across core and external modules  

dpol1:
## Summary  
  
This PR addresses a set of code quality, correctness, and robustness issues found across the `core` and `external` modules.  
  
### Thread Safety  
- Replace non-thread-safe `SimpleDateFormat` with `DateTimeFormatter` in `FileResponse` and `DateUtils.parseDate()` in `CookieConverter`  
- Replace shared mutable `Matcher` in `RefreshTag` with a `Pattern` constant and per-call `Matcher` instances  
  
### Resource Leaks  
- Use try-with-resources for `FileInputStream` in `FileResponse` and `InputStream` in `RegexURLFilterBase`  
- Add `cleanup()` to `DebugParseFilter` to close the output stream  
- Ensure Playwright tracing is always stopped via a `finally` block in `HttpProtocol`, even when an exception is thrown  
  
### Error Handling & Robustness  
- Wrap `Long.parseLong()` / `Integer.parseInt()` in `FetcherBolt` with try-catch to handle invalid metadata values gracefully  
- Add null check for classpath resource stream in `RegexURLFilterBase`  
- Fix duplicate `setExpiryDate()` call and add null guard in `CookieConverter`  
- Fix misleading error message in `CloudSearchUtils` ("must be score" → "must NOT be score")  
  
### Logging & Code Cleanup  
- Replace all `e.printStackTrace()` calls with proper SLF4J `LOG.error()`  
- Replace `URLEncoder.encode(url, "UTF-8")` (unnecessary checked exception) with the `StandardCharsets.UTF_8` overload  
- Replace manual `MessageDigest` boilerplate with `DigestUtils.sha512Hex()`  
- Use actual document charset in `JsRenderingDetector` instead of hardcoded UTF-8  
  
### No behavior changes intended  
All changes are backward-compatible. The crawl logic and output remain the same.

### DIFF
--- a/core/src/main/java/org/apache/stormcrawler/ConfigurableTopology.java
+++ b/core/src/main/java/org/apache/stormcrawler/ConfigurableTopology.java
@@ -30,8 +30,12 @@ import org.apache.storm.topology.TopologyBuilder;
 import org.apache.storm.utils.Utils;
 import org.apache.stormcrawler.persistence.Status;
 import org.apache.stormcrawler.util.ConfUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public abstract class ConfigurableTopology {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ConfigurableTopology.class);
 
     protected Config conf = new Config();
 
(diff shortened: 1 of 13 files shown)


jnioche:
Thanks @dpol1. Having a PR affecting loads of different files and covering such different aspects does not make it easy to review. Instead of a large dump, having one PR per type of issue (e.g. Resource Leaks) means that changes can be reviewed quicker, merged in isolation and also reviewed in parallel.
Any chance you could break it down a bit?

dpol1:
Sure, sorry about that — got a bit carried away. Will split it into three PRs: thread-safety, resource leaks, and fault tolerance/cleanup.

PR #1929, opened 2026-06-04
Fix thread safety: replace shared Matcher and SimpleDateFormat instances  

dpol1:
`SimpleDateFormat` and shared `Matcher` instances are not thread-safe.  
In a multi-threaded Storm topology this can cause silent data corruption  
or incorrect results under concurrent load.  
  
- `RefreshTag`: replace static shared `Matcher` with a `Pattern` constant  
  and a per-call `Matcher` instance  
- `FileResponse`: replace `SimpleDateFormat` with `DateTimeFormatter`  
  (also use try-with-resources for `FileInputStream` in the same class)  
- `CookieConverter`: replace `SimpleDateFormat` with `DateUtils.parseDate()`;  
  also fix a duplicate `setExpiryDate()` call and add a null guard on the  
  parsed date

### DIFF
--- a/core/src/main/java/org/apache/stormcrawler/util/CookieConverter.java
+++ b/core/src/main/java/org/apache/stormcrawler/util/CookieConverter.java
@@ -18,8 +18,6 @@
 package org.apache.stormcrawler.util;
 
 import java.net.URL;
-import java.text.ParseException;
-import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


rzo1 (approved):

core/src/main/java/org/apache/stormcrawler/util/CookieConverter.java
```diff
@@ -110,17 +108,17 @@ public static List<Cookie> getCookies(String[] cookiesStrings, URL targetURL) {
             // check expiration
             if (expires != null) {
                 try {
-                    Date expirationDate = DATE_FORMAT.parse(expires);
-                    cookie.setExpiryDate(expirationDate);
-
-                    // check that it hasn't expired?
-                    if (cookie.isExpired(new Date())) {
-                        continue;
+                    Date expirationDate = org.apache.http.client.utils.DateUtils.parseDate(expires);
```
jnioche, line 111: we have dependencies on o.a.httpclient in various places, it is acceptable to have it here too

jnioche (approved):

PR #1943, opened 2026-06-14
#466 - Handle text/plain content in JSoupParserBolt

### LINKED ISSUE #466: JSoup parser to handle text/plain
There is no parsing involved so we might as well do it.

2017-05-22 11:16:54.269 c.d.s.b.JSoupParserBolt Thread-10-parse-executor[7 7] [ERROR] Exception content-type text/plain for http://www.murfreesboropost.com/candidates-for-state-house-district-34-cms-23820
2017-05-22 11:16:55.921 c.d.s.b.JSoupParserBolt Thread-42-parse-executor[5 5] [ERROR] Exception content-type text/plain for http://www.murfreesboropost.com/candidates-for-state-house-district-62-cms-23821
2017-05-22 11:16:57.049 c.d.s.b.JSoupParserBolt Thread-4-parse-executor[8 8] [ERROR] Exception content-type text/plain for http://www.murfreesboropost.com/rutherford-county-sheriff-candidates-cms-23826

rzo1:
text/plain content requires no markup parsing, so instead of raising an error (the default with jsoup.treat.non.html.as.error) the bolt now uses the decoded content directly as the extracted text and emits no outlinks.

Adds a unit test and documentation updates.

Closes #466

### For all changes

### For code changes

### Note

### DIFF
--- a/core/src/main/java/org/apache/stormcrawler/bolt/JSoupParserBolt.java
+++ b/core/src/main/java/org/apache/stormcrawler/bolt/JSoupParserBolt.java
@@ -116,6 +116,19 @@ public class JSoupParserBolt extends StatusEmitterBolt {
 
     private TextExtractor textExtractor;
 
+    /**
+     * Honor {@link TextExtractor#NO_TEXT_PARAM_NAME} on the text/plain path. No extractor is
+     * invoked there, so the value is read directly.
+     */
+    private boolean plainTextNoText;
+
+    /**
+     * Character cap for stored plain text, read from {@link TextExtractor#TEXT_MAX_TEXT_PARAM_NAME}
+     * ({@code -1} = unbounded). Bounds the emitted text only; the raw fetched size is bounded by
+     * {@code http.content.limit}.
+     */
+    private int plainTextMaxSize;
+
     private String protocolMetadataPrefix;
 
     private boolean robotsHeaderSkip;
@@ -188,6 +201,12 @@ public class JSoupParserBolt extends StatusEmitterBolt {
             LOG.warn("Cannot instantiazr textextractor.class '{}'.", clazz, e);
             throw new RuntimeException(e);
         }
+
+        // The text/plain path does not run the TextExtractor (there is no markup to
+        // extract from), so the two size-related knobs are read here and applied
+        // directly. The include/exclude knobs require markup and have no effect.
+        plainTextNoText = ConfUtils.getBoolean(conf, TextExtractor.NO_TEXT_PARAM_NAME, false);
+        plainTextMaxSize = ConfUtils.getInt(conf, TextExtractor.TEXT_MAX_TEXT_PARAM_NAME, -1);
     }
 
     @Override
@@ -202,6 +221,10 @@ public class JSoupParserBolt extends StatusEmitterBolt {
         // check that its content type is HTML
         // look at value found in HTTP headers
         boolean contentTypeIsOk = false;
+
+        // text/plain content needs no markup parsing: the decoded bytes are
+        // the text, there are no outlinks (see issue #466)
+        boolean isPlainText = false;
 
         String mimeType =
                 metadata.getFirstValue(HttpHeaders.CONTENT_TYPE, this.protocolMetadataPrefix);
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/stormcrawler/bolt/JSoupParserBolt.java
```diff
@@ -272,79 +280,91 @@ public void execute(Tuple tuple) {
...
-            // and we are in strict mode
-            if (robotsTags.isNoFollow() && robotsNoFollowStrict) {
+            if (isPlainText) {
+                // no markup to parse: the decoded content is the text itself and
+                // there are no outlinks. An empty shell document is kept so that
+                // the downstream redirection check and parse filters still work.
+                jsoupDoc = org.jsoup.nodes.Document.createShell(url);
                 slinks = new HashMap<>(0);
+                robotsTags.normaliseToMetadata(metadata);
+                text = html;
```
dpol1: the plain text path stores `html` directly, so `textextractor.skip.after` and `textextractor.no.text` don't apply here. fine for typical .txt, but with `http.content.limit: -1` (default) a large file gets stored in full. worth truncating to maxTextSize, or noting the limitation in the docs?

dpol1 (approved):
looks good. the shell-doc trick keeps redirects and filters working and leaves the html path untouched, nice.

core/src/main/java/org/apache/stormcrawler/bolt/JSoupParserBolt.java
rzo1: Good catch. The plain-text branch does `text = html` directly, so neither `textextractor.no.text` nor `textextractor.skip.after` applies, and with the default `http.content.limit: -1` a large `.txt` is stored in full. I see two ways to bound it. Both reuse the existing `TextExtractor` config keys so there is no new knob to document.

### Option A: truncate the raw decoded string

Read the two limits in `prepare()` and apply them in the branch:

```java
// fields
private boolean plainTextNoText;
private int plainTextMaxSize;

// prepare()
plainTextNoText  = ConfUtils.getBoolean(conf, TextExtractor.NO_TEXT_PARAM_NAME, false);
plainTextMaxSize = ConfUtils.getInt(conf, TextExtractor.TEXT_MAX_TEXT_PARAM_NAME, -1);

// in the isPlainText branch
if (plainTextNoText) {
    text = "";
} else if (plainTextMaxSize > 0 && html.length() > plainTextMaxSize) {
    text = html.substring(0, plainTextMaxSize);
} else {
    text = html;
}
```

**Pros:** preserves the original plain-text formatting (newlines, spacing, indentation), which is usually the point of a `.txt`; trivially cheap.
**Cons:** re-reads config keys that conceptually belong to the `TextExtractor`, so a custom `textextractor.class` with different semantics would not be consulted; the cap is on character count, not on the extractor's accumulation logic.

### Option B: route the text through the configured extractor

Wrap the decoded content in the shell document and let the extractor produce the text:

```java
jsoupDoc = Document.createShell(url);
jsoupDoc.body().appendChild(new TextNode(html));
slinks = new HashMap<>(0);
robotsTags.normaliseToMetadata(metadata);
text = textExtractor.text(jsoupDoc.body());
```

**Pros:** a single text-extraction path for HTML and plain text; automatically honors `no.text`, `skip.after`, exclude tags, and any custom `textextractor.class`.
**Cons:** the extractor normalizes whitespace (`appendNormalisedText`), collapsing newlines and runs of spaces, so the stored plain text loses its original layout. Fine for indexing, lossy if the raw structure matters.

### Trade-off in one line

A keeps plain text faithful but special-cases the limits; B unifies the code path but mangles whitespace. Either way I would add a test asserting a `>skip.after` input is truncated.

WDYT?

dpol1: I'd go with Option A. A plain-text file is already its own text, which is the whole point of #466, and the test asserts it's stored verbatim, newlines and all. Option B runs the content through `appendNormalisedText`, which collapses whitespace, so logs, source files and tabular dumps lose their layout. It would also force relaxing that test. For the one format where whitespace is the content, that's a poor trade.

  The downside of Option A, re-reading the two config keys, is smaller than it looks. For plain text only `no.text` and `skip.after` have any effect; `include.pattern` and `exclude.tags` need markup, so there's nothing else to honor. If the custom `textextractor.class` case still bothers you, a cleaner option is a `text(String)` overload on the extractor that applies those two limits without normalizing, so both code paths share one implementation. I'd treat that as a follow-up, not part of this PR.

  One open question: should plain text use the `TextExtractor` knobs at all, or is `http.content.limit` the more honest bound for raw bytes? Either way, Option A plus a test for the `skip.after` truncation works for me.

rzo1: I went for Option A now. Happy for re-review.

dpol1 (approved):
👍

jnioche (approved):

jnioche:
thanks @rzo1 and @dpol1

Issue #867, opened 2021-03-26
Send host info on a specific stream 


jnioche:
Could be used to notify that a status HTTP 429 has been caught (see ##784 on that topic) or that the robots.txt specified a delay.
A bespoke bolt could then notify a URL Frontier via [setDelay](https://github.com/crawler-commons/url-frontier/blob/master/API/urlfrontier.proto#L55) or [blockQueueUntil](https://github.com/crawler-commons/url-frontier/blob/master/API/urlfrontier.proto#L45)

see #504 for a related discussion

dpol1:
Now that #1973  covers the explicit case (429/503 with a Retry-After header), I'd like to tackle the case we probably see most in the wild: servers that rate-limit without sending any header at all. Today we just fetch the host again and collect another 429.

The idea, kept short:

FetcherBolt already emits on the `hostinfo` stream when there's a header. I'd widen that signal with a `kind` field, so a headerless 429/503 becomes a "pressure" event. The consumer bolt keeps a small per-host counter and blocks the queue in the frontier for a growing duration: 60s the first time, then 2 minutes, 4, 8, capped at a day. If the host stays quiet for half an hour the counter is forgotten and the next incident starts from scratch.

Nothing new under the sun: Nutch has done this since 1.19 ( [NUTCH-2946](https://issues.apache.org/jira/browse/NUTCH-2946), same doubling curve, same 30 minute reset). The difference is where the state lives.Nutch keeps it inside the fetcher and later had to patch around fetching going stale when only backed-off queues remain ([NUTCH-3114](https://issues.apache.org/jira/browse/NUTCH-3114)). With the state in the frontier a blocked host never reaches the fetcher in the first place, so that class of problem can't happen. Workers stay stateless, which is also the reason #1944 was closed.

One subtlety worth calling out. When the first 429 arrives there may be a dozen URLs of that host already sitting in the fetcher's internal queues. They will fail one after the other, and each of those failures must not escalate the counter, otherwise a single incident jumps straight to the 24h cap. So: escalate at most once per block window. Same trick TCP uses, where multiple losses inside one RTT count as a single congestion event.

A couple of questions before I write any code:

[...]

PR #1944, opened 2026-06-14
#784 - Support Retry-After in FetcherBolt

### LINKED ISSUE #784: support retry-after in FetcherBolt
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After

if a site is asking for a delay, we should enforce it in the internal queues of the FetcherBolt.

rzo1:
Honour the Retry-After HTTP response header by delaying the next fetch from the affected internal queue. The header value is parsed as either a number of seconds or an HTTP date. A new fetcher.max.retry.after config caps the honoured delay (-1, the default, honours it as-is).

### For all changes

### For code changes

### DIFF
--- a/core/src/main/java/org/apache/stormcrawler/bolt/FetcherBolt.java
+++ b/core/src/main/java/org/apache/stormcrawler/bolt/FetcherBolt.java
@@ -98,8 +98,8 @@ public class FetcherBolt extends StatusEmitterBolt {
     public static final String FETCH_TIMEOUT_PARAM_KEY = "fetcher.thread.timeout";
 
     /**
-     * Maximum delay in seconds the fetcher will wait when a server requests a back-off via the
-     * <a href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After">Retry-After</a>
+     * Maximum delay in seconds the fetcher will wait when a server requests a back-off via the <a
+     * href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After">Retry-After</a>
      * HTTP response header. A value of {@code -1} (the default) means the delay requested by the
      * server is honored as-is, with no upper bound.
      */
--- a/core/src/main/java/org/apache/stormcrawler/bolt/FetcherBolt.java
+++ b/core/src/main/java/org/apache/stormcrawler/bolt/FetcherBolt.java
@@ -24,7 +24,10 @@ import java.net.InetAddress;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.UnknownHostException;
+import java.time.DateTimeException;
 import java.time.Instant;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


jnioche:
thanks @rzo1
my concern is that this would cause unnecessary tuple timeouts. If we come across a retry after, we should probably purge the queue instead.
A more robust approach  would be to implement it via the host stream (see #867) and have a host aware spout implementation. I had made a start in the branch 851

rzo1:
>  I had made a start in the branch 851 

Contains SOLR changes only, so most likely gone.

--

I agree, that purging the queue for the given host might be beneficial. Let me later look into that in more depth again.

jnioche:
> > I had made a start in the branch 851
> 
> Contains SOLR changes only, so most likely gone.

was actually in branch 990, sorry

rzo1:
Agreed, the in-queue hold is the fragile part. A long `Retry-After` (and the default is uncapped) would park the queue's siblings past `topology.message.timeout.secs` and trigger replays.

For this PR, I would take the **purge** route. On a `Retry-After`, re-emit the affected queue's pending items rather than holding them, mirroring the existing `crawl-delay-too-long` path at `FetcherBolt.java:682` so the frontier reschedules them. One thing I'd like your take on: should the re-emitted URLs go out as `Status.ERROR` (reuses the existing path, but carries error/retry semantics), or should we set an explicit future `nextFetchDate` so the scheduler honors the exact back-off?

I'd treat the **host stream / host-aware spout** design (#867, your branch 990) as the proper long-term home for this, happy to follow up there once we settle the short-term behaviour. Does that split sound right to you?

jnioche:
yes, the split makes sense

> One thing I'd like your take on: should the re-emitted URLs go out as `Status.ERROR` (reuses the existing path, but carries error/retry semantics), or should we set an explicit future `nextFetchDate` so the scheduler honors the exact back-off?

`Status.ERROR` is not the right status: it indicates an irremediable problem with the content of the document, like a pdf that would be unparsable for instance or a URL blocked by robots.txt

Could set an explicit `nextFetchDate` but I think just mimicking what is done via `crawl-delay-too-long` would be good enough.

rzo1:
> Status.ERROR is not the right status: it indicates an irremediable problem with the content of the document, like a pdf that would be unparsable for instance or a URL blocked by robots.txt

Yes this was my reasoning too but wanted to get some confirmation on it :)

core/src/main/java/org/apache/stormcrawler/bolt/FetcherBolt.java
sebastian-nagel, line 487: If the next fetch time is in the future, the queue cannot be released for now. Otherwise the retry-after is not honored for new fetch items of the same site arriving through the topology.

See [Nutch's FetchItemQueues](https://github.com/apache/nutch/blob/8e03a3e998c1aac32ae3eb06f9f7cdf116c0c5f8/src/java/org/apache/nutch/fetcher/FetchItemQueues.java#L193) code to release queues in combination with the exponential back-off.

> or should we set an explicit future `nextFetchDate` so the scheduler honors the exact back-off?

This would only help if there are not too many items from the same site.

> I'd treat the host stream / host-aware spout design (https://github.com/apache/stormcrawler/issues/867, your branch 990) as the proper long-term home for this,

Definitely. Avoiding back-pressure is important. And in a broad crawl the number of queues, which need to be kept to ensure the retry-after delay, can grow large.

rzo1:
Thanks @jnioche and @sebastian-nagel for the review and discussion.

Summary of where we landed: honouring `Retry-After` by holding the internal
queue inside `FetcherBolt` is workable, but fragile. To make it correct we'd
also need to make queue reaping back-off aware (don't reap a queue while its
`nextFetchTime` is in the future, see @sebastian-nagel's note and Nutch's
`FetchItemQueues`), and even then a long delay risks back-pressure and tuple
timeouts, and the number of held queues can grow large in a broad crawl.

Given that #784 has been open for a long time without concrete user demand,
investing in the interim in-bolt workaround doesn't seem worth the added
complexity. The proper long-term home for this is the host-aware spout / host
stream design (#867), which avoids the back-pressure problem entirely.

So I'll close this PR in favour of pursuing #867. #784 stays open and we can
revisit `Retry-After` there as part of the host-aware implementation.

PR #1954, opened 2026-06-16
#1951 refactor: replace linear proxy scan with lookup map in MultiProxyManager

### LINKED ISSUE #1951: [Improvement] Replace linear proxy scan in MultiProxyManager with a lookup map

  Follow-up to #1935. `MultiProxyManager.getConfiguredProxy` linearly scans the configured proxies on every metadata-override request, to reuse the configured `SCProxy` instance for usage accounting. O(n) per override request.

  Low priority. A miss is harmless (the metadata proxy is still used, just as a throwaway instance), the override path is opt-in per URL, and `getLeastUsed` already does a comparable per-request scan. This is cleanup, not a measured bottleneck. To be clear, `getLeastUsed` is inherently O(n) and is not a target here; I mention it only because O(n) is already accepted on this path.

  ### How should we improve?

  The lookup only exists to reuse the configured `SCProxy` so its usage counter keeps incrementing. The cleanest fix puts proxy identity on the type instead of rebuilding it in a separate key.

  1. Give `SCProxy` an `equals`/`hashCode` based on its identity fields only: protocol and address lowercased, plus port, username and password. Leave out `country`,  `area`, `location`, `status` and `usage`. None of the identity fields have setters, so the hash stays stable for the life of the object. (A blind "generate equals/hashCode" from the IDE pulls in every field, which is wrong here.)
  2. Build a `Map<SCProxy, SCProxy>` in both `configure()` methods and turn `getConfiguredProxy` into `Optional.ofNullable(map.get(proxy))`, which makes it O(1).
  3. `ProxyUtils.isSameProxy` is only called by `getConfiguredProxy`, so after step 2 it has no callers. Fold its logic into `SCProxy.equals` and delete `isSameProxy` along with its private `normalize` helper. Keep `ProxyUtils` itself, since it still holds `validatePortRange`. A normalized string key would also work, but it duplicates the same five-field comparison and leaves two definitions of proxy identity to keep in sync.

  No behavior change. This is a code-quality cleanup (single identity definition, less dead code), not a performance fix.

leedongkyu0407:
Closes #1951

## Summary
Replace the O(n) linear scan in `getConfiguredProxy()` with an O(1) map lookup by putting proxy identity on the type itself.

## Changes
- Added `equals`/`hashCode` to `SCProxy` based on identity fields only: `protocol` and `address` (lowercased), `port`, `username`, `password`. Excludes `country`, `area`, `location`, `status`, and `usage`.
- Built a `Map<SCProxy, SCProxy>` in both `configure()` overloads and replaced `getConfiguredProxy` with `Optional.ofNullable(map.get(proxy))`
- Removed `ProxyUtils.isSameProxy` and its private `normalize` helper, folding the logic into `SCProxy.equals`

## Notes
- No behavior change
- Single definition of proxy identity, less dead code
- `MultiProxyManagerTest`: 13/13 passed
- `ProtocolTest.testBlocking` failure is pre-existing and unrelated to this change

### For all changes
### For code changes
- [ ] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under ASF 2.0?

### DIFF
--- a/core/src/main/java/org/apache/stormcrawler/proxy/MultiProxyManager.java
+++ b/core/src/main/java/org/apache/stormcrawler/proxy/MultiProxyManager.java
@@ -21,6 +21,8 @@ import java.io.File;
 import java.io.IOException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Random;
 import java.util.Scanner;
@@ -44,6 +46,7 @@ public class MultiProxyManager implements ProxyManager {
     private SCProxy[] proxies;
     private ProxyRotation rotation;
     private final AtomicInteger lastAccessedIndex = new AtomicInteger(0);
+    private Map<SCProxy, SCProxy> proxyLookupMap;
 
     private static final org.slf4j.Logger LOG = LoggerFactory.getLogger(HttpProtocol.class);
 
@@ -146,6 +149,11 @@ public class MultiProxyManager implements ProxyManager {
 
         // assign proxies to class variable
         this.proxies = fileProxies;
+
+        this.proxyLookupMap = new HashMap<>();
+        for (SCProxy proxy : this.proxies) {
+            this.proxyLookupMap.put(proxy, proxy);
+        }
     }
 
     public void configure(ProxyRotation rotation, String[] proxyList) throws RuntimeException {
@@ -170,6 +178,11 @@ public class MultiProxyManager implements ProxyManager {
 
         // assign proxies to class variable
         this.proxies = fileProxies;
+
+        this.proxyLookupMap = new HashMap<>();
+        for (SCProxy proxy : this.proxies) {
+            this.proxyLookupMap.put(proxy, proxy);
+        }
     }
 
     private SCProxy getRandom() {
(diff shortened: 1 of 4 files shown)


rzo1:
Thanks for the refactor, the O(n) to O(1) lookup and folding proxy identity into `SCProxy.equals`/`hashCode` looks correct and consistent with the removed `isSameProxy`. Two things I'd like to see addressed before merge:

**1. Missing tests for the new `equals`/`hashCode`.** The identity contract is now the heart of the lookup, but it has no coverage. Could you add unit tests covering:
- equal on identical identity fields, and `hashCode` equal for equal objects
- case-insensitivity of `protocol`/`address`
- not-equal on differing `port` / `username` / `password`
- `country` / `area` / `location` / `status` are excluded from identity
- null / different-type handling
- a `MultiProxyManagerTest` case confirming metadata-driven lookup resolves to the configured instance through the new map (the path that replaced the linear scan)

**2. Mutable fields used in the hash key.** `username` and `password` are non-final, yet they are part of `equals`/`hashCode` and these objects are used as `HashMap` keys. It is safe today because they are only assigned in the constructor, but mutating either after insertion would corrupt the map lookup. Could you either make them `final` (preferred, they are never reassigned after construction) or add a short comment noting the invariant?

Minor: the diff also carries unrelated reformatting in `MultiProxyManager.java` (Javadoc reflow, reduced continuation indents, blank line after imports) that does not match the project's `git-code-format` style and will fail the format check. Running `mvn git-code-format:format-code -Dgcf.globPattern="**/*" -Dskip.format.code=false` should bring the diff down to just the real change.

dpol1 (approved):
lgtm — Map lookup preserves `isSameProxy` identity semantics, `username/password` final is the right fix for hash key, tests cover it — Thanks for the contribution @leedongkyu0407

leedongkyu0407:
Thank you for the review and guidance!

rzo1 (approved):

PR #1956, opened 2026-06-17
  #1955 - Make FileSpout work in distributed mode

### LINKED ISSUE #1955: [Bug report] FileSpout does not work in distributed mode
### Version

main branch

### Describe what's wrong

It relies on the constructors but in distributed mode Storm expects variables to have been serialised and the logic to be done in `open()`.

### Error message and/or stacktrace

nothing happens - the logs are not showing any activity

### How to reproduce

any version - this has been the case since the beginning of SC

### Additional context

_No response_

dpol1:
FileSpout read the seed directory and filled its input queue inside the constructor. In distributed mode the constructor runs only on the client that submits the topology, so the resolved queue (usually empty, since the seed files live on the workers) was serialised and shipped out. The spout then sat idle with no activity in the logs.

  The constructors now only keep the values they are given: the directory and filter, or the explicit file list. The seeds are resolved in `open()`, which runs on each worker. `WARCSpout` already calls `super.open()`, so it gets the same behaviour without any change.

  While moving the code I also fixed a log call that used printf-style `%s` formatting instead of SLF4J `{}` placeholders.

  Tests: a regression test builds the spout against an empty directory and creates the seed file only after construction, before `open()`. It fails on the old code and passes now. There is also a directory happy-path test and a serialisation round-trip.

  Closes #1955

### For all changes

### For code changes

### Note

### DIFF
--- a/core/src/main/java/org/apache/stormcrawler/spout/FileSpout.java
+++ b/core/src/main/java/org/apache/stormcrawler/spout/FileSpout.java
@@ -61,6 +61,9 @@ public class FileSpout extends BaseRichSpout {
     public static final int BATCH_SIZE = 10000;
     public static final Logger LOG = LoggerFactory.getLogger(FileSpout.class);
     private final Queue<String> inputFiles;
+    private final String seedDir;
+    private final String fileFilter;
+    private final String[] seedFiles;
     protected SpoutOutputCollector collector;
     protected Scheme scheme = new StringTabScheme();
     protected LinkedList<byte[]> buffer = new LinkedList<>();
@@ -68,7 +71,7 @@ public class FileSpout extends BaseRichSpout {
     protected int totalTasks;
     protected int taskIndex;
     private BufferedReader currentBuffer;
-    private boolean withDiscoveredStatus = false;
+    private final boolean withDiscoveredStatus;
 
     /**
      * @param dir containing the seed files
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/stormcrawler/spout/FileSpout.java
```diff
@@ -94,18 +97,10 @@ public FileSpout(String... files) {
...
-                inputFiles.add(inputFile);
-                LOG.info("Input : {}", inputFile);
-            }
-        } catch (IOException ioe) {
-            LOG.error("IOException: %s%n", ioe);
-        }
+        this.seedDir = dir;
+        this.fileFilter = filter;
+        this.seedFiles = null;
+        this.inputFiles = new LinkedList<>();
```
jnioche: no need to instantiate it here - I would  do that in` open()` just before the call to `populateInputFiles()`

core/src/main/java/org/apache/stormcrawler/spout/FileSpout.java
```diff
@@ -115,12 +110,14 @@ public FileSpout(String dir, String filter, boolean withDiscoveredStatus) {
...
         if (files.length == 0) {
             throw new IllegalArgumentException("Must configure at least one inputFile");
         }
-        inputFiles = new LinkedList<>();
-        Collections.addAll(inputFiles, files);
+        this.withDiscoveredStatus = withDiscoveredStatus;
+        this.seedDir = null;
+        this.fileFilter = null;
+        this.seedFiles = files;
+        this.inputFiles = new LinkedList<>();
```
jnioche: same as above

core/src/main/java/org/apache/stormcrawler/spout/FileSpout.java
```diff
@@ -61,14 +61,17 @@ public class FileSpout extends BaseRichSpout {
     public static final int BATCH_SIZE = 10000;
     public static final Logger LOG = LoggerFactory.getLogger(FileSpout.class);
     private final Queue<String> inputFiles;
+    private final String seedDir;
+    private final String fileFilter;
+    private final String[] seedFiles;
     protected SpoutOutputCollector collector;
```
jnioche: might as well mark collector as `transient`

jnioche: in fact, anything handled in `open()` could be marked as transient
totalTasks
taskIndex

jnioche (commented):
looks good, a few minor suggestions

rzo1 (approved):

jnioche:
Thanks @dpol1

PR #1981, opened 2026-07-07
Port QueueBolt (queue stream) to the opensearch-java module

rzo1:
Ports the queue stream feature (#1974) from `external/opensearch` to the newer `external/opensearch-java` module, which uses the opensearch-java client instead of the legacy REST high-level client.


jnioche (approved):

jnioche:
thanks @rzo1

Issue #1979, opened 2026-07-07
Propagate robots.txt crawl-delay to URLFrontier via setDelay


dpol1:
Third part of #867, after the Retry-After handling (#784, done in #1973) and the adaptive back-off for headerless rate limits (#1106).

Today a robots.txt `Crawl-delay` above `fetcher.max.crawl.delay` has no good outcome: by default the URL is marked `Status.ERROR` (`error.cause=crawl_delay`) and the host is never crawled, with `fetcher.max.crawl.delay.force` it is fetched at a pace that violates what robots asked. Short delays are fine where they are, inside the fetcher queues; it's the long tail that loses either way. And these sites are exactly the ones that mean it - crawl-delay is the most-complied-with robots directive in the wild (https://arxiv.org/abs/2505.21733).

The frontier can do better. With `force` enabled the fetcher would report the requested delay in the status metadata, and the queue-stream consumer (`QueueRegulatorBolt`) would pass it to URLFrontier via [setDelay](https://github.com/crawler-commons/url-frontier/blob/master/API/src/main/protobuf/urlfrontier.proto#L68): the host becomes crawlable at the pace robots asked for, and the compliance that `force` alone breaks is restored at the source. Same split as Nutch-style crawlers moved to long ago: politeness state lives with the queue owner, not the workers.

Two things worth calling out. Unlike `blockQueueUntil`, `setDelay` does not expire on its own, and `setDelay(key, 0)` erases a server-side default for that queue - so when a site later removes its `Crawl-delay`, the old pace simply stays (throughput lost on that host, never impoliteness) rather than guessing at a reset. And the value needs a cap of its own: SC disables the parser-side limit on crawl-delays, and sites do ask for absurd numbers.

PR #1984, opened 2026-07-08
Adaptive back-off for rate-limited hosts via the queue stream

### LINKED ISSUE #1106: Fetcher: optionally slow down fetching from hosts with repeated exceptions
See [NUTCH-2946](https://issues.apache.org/jira/browse/NUTCH-2946)

> The fetcher holds for every fetch queue a counter which counts the number of observed "exceptions" seen when fetching from the host (resp. domain or IP) bound to this queue.
> 
> As an improvement to increase the politeness of the crawler, the counter value could be used to dynamically increase the fetch delay for hosts where requests fail repeatedly with exceptions or HTTP status codes mapped to ProtocolStatus.EXCEPTION (HTTP 403 Forbidden, 429 Too many requests, 5xx server errors, etc.) Of course, this should be optional. The aim to reduce the load on such hosts already before the configured max. number of exceptions (property fetcher.max.exceptions.per.queue) is hit.

dpol1:
Stage 2 of #867, closes #1106. #1973 handled the explicit case, a 429/503 carrying a Retry-After header. This covers the one we probably see most in the wild: servers that rate-limit without sending any header. Today we just fetch the host again and collect another 429.

A headerless 429/503 arriving on the queue stream now blocks the host's queue in URLFrontier for a growing duration: 60s the first time, doubled per new incident, capped at a day, forgiven after 30 quiet minutes. Same curve as Nutch since 1.19 ([NUTCH-2946](https://issues.apache.org/jira/browse/NUTCH-2946)), but the state lives with the queue owner: a blocked host never reaches the fetcher at all, so the stale fetching Nutch had to patch around ([NUTCH-3114](https://issues.apache.org/jira/browse/NUTCH-3114)) can't happen here.

Everything from the #867 discussion is in:

* a random fraction on every computed block (`urlfrontier.backoff.jitter`, default 0.1) so hosts blocked on the same schedule don't all come back at once. Thanks @sebastian-nagel, the daily error spike was exactly the failure mode I had dismissed.
* configurable trigger codes (`urlfrontier.backoff.status.codes`, default 429 and 503; 403 stays out since it so often just means "forbidden"). They now gate the Retry-After path of #1973 as well, with one behaviour change: a malformed or zero Retry-After escalates as a headerless incident instead of being ignored.
* fetch exceptions as incidents behind `urlfrontier.backoff.on.exceptions`, off by default: a timeout is a noisier signal than a clean 429, I'd flip the default once this has seen some real crawls.
* URLs already in the fetcher's internal queues just fail and reschedule. Failures during an active block count as the same incident, so a draining backlog escalates at most once per window instead of jumping to the cap.

One rename against what I posted on #867: the config prefix is `urlfrontier.backoff.*`, not `fetcher.backoff.*`. The fetcher plays no part in this, and it sits next to Phase 1's `urlfrontier.max.retry.after`.

How it's built:

* the decision logic is a package-private `HostBackoff`, same split as `RetryAfterParser` in #1973; the bolt keeps the RPC and nothing else.
* state is one Caffeine cache with per-entry expiry at block end + decay: memory stays bounded and the 24h cap stays reachable through escalation.
* a block is only ever extended, never shrunk by a later, shorter signal. A signal suppressed by that guard or by the anti-burst one re-asserts the stored block with an idempotent RPC, so a lost fire-and-forget call heals on the next signal instead of leaving a phantom block.
* `HostBlockBolt` becomes `QueueRegulatorBolt` in a separate mechanical commit: with #1979 lined up next, blocking is one of the actions this bolt takes on a queue, not the only one.
* same `metadata.persist` caveat as #1973 (`fetch.statusCode`, plus `fetch.exception` when the flag is on); the bolt warns at startup when they wouldn't survive the filter.

Tests: escalation math against a controllable ticker and a seeded RNG for the jitter, regressions for the burst and only-extend guards, and two Testcontainers runs against a real frontier (explicit Retry-After and headerless 429).

### For all changes

- [ ] Does your PR title start with `#XXXX` where `XXXX` is the issue number you are tryin
- [ ] Is your initial contribution a single, squashed commit? (mechanical rename kept as idiff stays reviewable)

### For code changes

- [ ] If applicable, have you updated the LICENSE file, including the main LICENSE file? (n/a)
- [ ] If applicable, have you updated the NOTICE file, including the main NOTICE file? (n/a)


external/urlfrontier/src/main/java/org/apache/stormcrawler/urlfrontier/Constants.java
```diff
@@ -51,4 +51,65 @@ private Constants() {}
     public static final String URLFRONTIER_MAX_RETRY_AFTER_KEY = "urlfrontier.max.retry.after";
 
     public static final int URLFRONTIER_MAX_RETRY_AFTER_DEFAULT = 86400;
+
```
rzo1, line 54: All of these need to land int the docs :)

rzo1:
Thanks for the PR. Aside from the docs thing already mentioned, I have two things while skimming through the PR:

1. HostBackoff imports com.github.benmanes.caffeine.cache.*, but external/urlfrontier/pom.xml doesn't declare caffeine as dependency it; it only compiles via the transitive compile-scope dep from stormcrawler-core. Add com.github.ben-manes.caffeine:caffeine to the urlfrontier pom.
2.  backoff.max.secs doesn't cap the first block since the first-incident branch (backoffSecs == 0 ? baseSecs : min(...)) skips the cap, so with max.secs < base.secs the first block uses the full base while later ones are capped lower; a negative max.secs blocks once at base and then never again. I would clamp the first step against maxSecs too (or clamp maxSecs >= baseSecs in the constructor), consistent with the clamping already done for factor, decay and jitter.

external/urlfrontier/src/main/java/org/apache/stormcrawler/urlfrontier/Constants.java
dpol1: Absolutely!

dpol1:
@jnioche I would love to hear your opininion too - when you have time. Just a bit of more context: the Stage 3 is on the way related to #1979 issue, I will share it when this lands

external/urlfrontier/src/main/java/org/apache/stormcrawler/urlfrontier/HostBackoff.java
```diff
@@ -0,0 +1,333 @@
...
+        this.jitter =
+                Math.max(
+                        0f,
+                        ConfUtils.getFloat(
+                                conf,
+                                Constants.URLFRONTIER_BACKOFF_JITTER_KEY,
+                                Constants.URLFRONTIER_BACKOFF_JITTER_DEFAULT));
+        this.jitterSource = jitterSource;
+        this.hosts =
+                Caffeine.newBuilder()
```
jnioche, line 189: Does not this re-implements Caffeine's built-in Expiry.writing(...) factory?

`Expiry.writing((k, s) -> Duration.ofSeconds(s.ttlSecs()))` should be semantically identical: the function applies on create and update, reads keep the current duration.

jnioche (commented):
LGTM

dpol1:
Thanks, both points are addressed.

jnioche (approved):

jnioche:
thanks @dpol1 and reviewers

PR #1987, opened 2026-07-09
Pluggable robot rule parser

jnioche:
# Make the robots.txt parser implementation configurable

Currently `AbstractHttpProtocol` hardcodes `HttpRobotRulesParser` as the class used to fetch and parse robots.txt for every HTTP-based Protocol implementation. This makes it impossible to customize robots.txt handling (e.g. to support directives that crawler-commons doesn't recognise, such as the emerging [Content-Signal ](https://contentsignals.org/) proposal) without forking core classes.

This PR makes the implementation pluggable via configuration, following the same pattern already used for `http.protocol.implementation` and `http.proxy.manager`:

- Added a new config key, `http.robots.parser.class`, defaulting to `org.apache.stormcrawler.protocol.HttpRobotRulesParser `(unchanged default behaviour).
- `AbstractHttpProtocol` now instantiates the configured class via `InitialisationUtil.initializeFromQualifiedName(...),` typed against the abstract `RobotRulesParser` rather than the concrete `HttpRobotRulesParser`.
- `HttpRobotRulesParser's` no-arg constructor is now public (required for reflective instantiation).
- Added `RobotRules.getWrapped()`, returning the wrapped `BaseRobotRules. HttpRobotRulesParser.getRobotRulesSet() `always wraps whatever `parseRules() `returns in a plain `RobotRules` for content-length/cache-hit tracking; without this accessor, a custom `RobotRulesParser` that returns its own `BaseRobotRules` subclass (to carry extra parsed data) would have that subtype hidden by the outer wrapper with no way to retrieve it.
- Documented the new setting in `crawler-default.yaml`.

No behavioural change with the default configuration; existing tests pass unchanged.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -20,9 +20,13 @@ on:
   push:
     branches:
       - main
+    paths-ignore:
+      - 'docs/**'
   pull_request:
     branches:
       - main
+    paths-ignore:
+      - 'docs/**'
 
 jobs:
   rat:
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -20,13 +20,9 @@ on:
   push:
     branches:
       - main
-    paths-ignore:
-      - 'docs/**'
   pull_request:
     branches:
       - main
-    paths-ignore:
-      - 'docs/**'
 
 jobs:
   rat:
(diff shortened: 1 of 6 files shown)


.github/workflows/maven.yml
rzo1: Dont think this should be in this PR.

core/src/main/java/org/apache/stormcrawler/protocol/AbstractHttpProtocol.java
```diff
@@ -121,7 +121,15 @@ public void configure(Config conf) {
             customHeaders.add(KeyValue.build(h));
         }
 
-        robots = new HttpRobotRulesParser(conf);
+        String robotsParserImplementation =
+                ConfUtils.getString(
+                        conf,
+                        "http.robots.parser.class",
```
rzo1, line 127: Should be added to docs too

.github/workflows/maven.yml
jnioche: yes, it leaked from a different branch which has since been merged in main.
Will remove so that it is cleaner

jnioche:
@rzo1 thanks for your comments, should now be addressed

rzo1 (approved):

jnioche:
thanks @rzo1

PR #2004, opened 2026-07-18
bugfix - max content not applied for robots.txt when redirecting

jnioche:
Trivial fix

`http.robots.content.limit` is used on the original fetch attempt but not on redirects


sigee (approved):
LGTM

rzo1 (approved):

PR #2005, opened 2026-07-18
Fix #304: Use HTML test document instead of sitemap XML in SubDocumentsFilterTest

### LINKED ISSUE #304: SubDocumentsFilterTest should not use a sitemap file
this currently uses the JSoupParser - which works only because we allow it to parse when no mimetype is available. We should have a minimalistic HTML doc to do the same OR use the RSS feed and generate subdocuments for each feed item

AsifQureshi04:
## Summary

Fixes #304.

The `SubDocumentsFilterTest` was using a sitemap XML file (`stormcrawler.sitemap.xml`) as test input for `JSoupParserBolt`, which is an HTML parser. This only worked because MIME type detection was disabled — a fragile workaround.

**Changes:**
- Created a minimal HTML test document (`subdocuments.html`) with 6 anchor links
- Rewrote `SubDocumentsParseFilter` to extract links from the W3C DOM (`DocumentFragment`) instead of re-parsing raw bytes as XML with XPath
- Removed the `detect.mimetype = false` hack from the test

## Test plan

- [x] `SubDocumentsFilterTest` passes (7 emitted tuples: 6 subdocuments + 1 parent)
- [x] All other parse filter tests pass (`XPathFilterTest`, `CollectionTaggerTest`, `CSVMetadataFilterTest`)
- [x] Code formatted with `mvn git-code-format:format-code`

### DIFF
--- a/core/src/test/java/org/apache/stormcrawler/parse/filter/SubDocumentsFilterTest.java
+++ b/core/src/test/java/org/apache/stormcrawler/parse/filter/SubDocumentsFilterTest.java
@@ -18,8 +18,6 @@
 package org.apache.stormcrawler.parse.filter;
 
 import java.io.IOException;
-import java.util.HashMap;
-import java.util.Map;
 import org.apache.stormcrawler.Metadata;
 import org.apache.stormcrawler.bolt.JSoupParserBolt;
 import org.apache.stormcrawler.parse.ParsingTester;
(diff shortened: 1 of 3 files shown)


jnioche:
ParseFilters are not specific to JSoupParseFilter: it is used by all [Parsers](https://github.com/search?q=repo%3Aapache%2Fstormcrawler%20Parsefilters&type=code), including SitemapParserBolt, which is why we had this test.

AsifQureshi04:
Thanks for the context @jnioche . I understand that ParseFilters are used across all parsers, including SiteMapParserBolt.

I followed the approach you suggested in the issue — using a minimalistic HTML document instead. The rewritten SubDocumentsParseFilter now works with the W3C DOM (DocumentFragment) provided by JSoupParserBolt, rather than re-parsing the raw bytes as XML.

Would you prefer I also add a separate test that demonstrates SubDocumentsParseFilter working with SiteMapParserBolt? Happy to adjust based on your feedback.

jnioche:
> Thanks for the context @jnioche . I understand that ParseFilters are used across all parsers, including SiteMapParserBolt.
> 
> I followed the approach you suggested in the issue — using a minimalistic HTML document instead. The rewritten SubDocumentsParseFilter now works with the W3C DOM (DocumentFragment) provided by JSoupParserBolt, rather than re-parsing the raw bytes as XML.

Sorry @AsifQureshi04, I opened the issue 10 years ago (!?!) and needed a bit of time to refresh my memory. Your PR is fine as it is.

> Would you prefer I also add a separate test that demonstrates SubDocumentsParseFilter working with SiteMapParserBolt? Happy to adjust based on your feedback.

That's what I was wondering about but actually it is probably not needed and would mean that the document would be reparsed as SitemapParserBolt does not generate a DOM. 

Thanks for your contribution to StormCrawler

jnioche (approved):

PR #2012, opened 2026-07-22
Allow indexer bolts to use a metadata value as the document ID

sigee:
The change should cover the discussion in the https://github.com/apache/stormcrawler/issues/671 issue.


core/src/main/java/org/apache/stormcrawler/indexing/AbstractIndexerBolt.java
```diff
@@ -259,12 +272,32 @@ protected Map<String, String[]> filterMetadata(Metadata meta) {
      *
      * @param metadata The {@link Metadata}.
      * @param normalisedUrl The normalised url.
-     * @return Return the normalised url SHA-256 digest as String.
+     * @return The value found in the metadata for the key configured with {@link
+     *     #DOC_ID_METADATA_PARAM_NAME}, if any, otherwise the normalised url SHA-256 digest as
+     *     String.
      */
     protected String getDocumentID(Metadata metadata, String normalisedUrl) {
+        final String fromMetadata = getDocumentIDFromMetadata(metadata);
```
jnioche, line 293: what about a sha256Hex(fromMetadata)? like we do for URLs?

jnioche (commented):
Looks good, forgot to mention that we should add the new `indexer.md.docid` to `core/src/main/resources/crawler-default.yaml`
with a default value of "".

rzo1 (commented):
Changes look good to me, but we should update the docs alongside this PR too, so people get an updated doc version ;-)

docs/src/main/asciidoc/configuration.adoc
```diff
@@ -318,6 +318,7 @@ The values below are used by sub-classes of `AbstractIndexerBolt`.
 
 | indexer.canonical.name | canonical | Metadata key for the canonical URL. Used to replace the URL with its canonical form before indexing.
 | indexer.ignore.empty.fields | false | If true, skip fields with empty values when indexing.
+| indexer.md.docid | - | Metadata key whose value, if present, is used as the document ID instead of the default SHA-256 digest of the URL. Useful e.g. for content-based deduplication when a `ParseFilter` stores a content hash in the metadata.
```
dpol1: Two precisions for this row: the value is digested, not used raw (SHA-256 here, SHA-512 via `CloudSearchUtils.getID` for CloudSearch); and the SOLR/SQL indexers never call `getDocumentID()`, so the setting is silently ignored there. Worth stating the supported backends.

core/src/main/java/org/apache/stormcrawler/indexing/AbstractIndexerBolt.java
```diff
@@ -75,8 +75,20 @@ public abstract class AbstractIndexerBolt extends BaseRichBolt {
...
 
+    /**
+     * Metadata key whose value should be used to derive the document ID, if present. The value is
+     * passed through the same SHA-256 digest used for the URL-based ID, so this can be used e.g. by
+     * a {@link org.apache.stormcrawler.parse.ParseFilter} generating a hash of the content so that
+     * duplicate content ends up under the same document ID regardless of the URL(s) it was found
+     * at. Falls back to the default (a SHA-256 digest of the URL) if not set or if the metadata
+     * does not contain a value for the configured key.
+     */
+    public static final String DOC_ID_METADATA_PARAM_NAME = "indexer.md.docid";
```
dpol1, line 98: Worth documenting: on the deletion path (gone pages) there's no parsing, so the DeletionBolt only sees this key if it's in `metadata.persist` and round-trips through the status index. Otherwise it falls back to `sha256(url)` and the document indexed under the metadata ID is silently orphaned.

dpol1 (commented):
Implementation looks correct and backward compatible. One design question for the maintainers: with content-based dedup, N URLs share one document. When a single URL goes away, the DeletionBolt deletes the document even if the content is still live under the other URLs, until one gets re-fetched. Acceptable as a documented limitation, or should deletion be skipped when the ID comes from metadata?

Inline comments cover a `metadata.persist` prerequisite and backend coverage of the new setting.

jnioche (approved):
thanks @sigee

PR #2052, opened 2026-08-16
#315 Adaptative URL filter to normalize URLs based on canonical tag  

### LINKED ISSUE #315: Adaptative URL filter to normalize URLs based on canonical tag
Such a filter could compare the parameters of a URL with the canonical tag found in the page (if any) and  determine after a while which parameters can be safely removed in order to normalise the URL.

The aim is similar to the [clean-param extension of the robots protocol by Yandex](https://yandex.com/support/webmaster/controlling-robot/robots-txt.xml#clean-param) where sites can specify how URLs can be normalised.

TODO compare with [research.google.com/pubs/archive/35210.pdf]

akash-manna-sky:
Adaptative URL filter to normalize URLs based on canonical tag
  Fixes #315 

### For all changes

### For code changes

### Note

---

> [!IMPORTANT]
> **🛠️ Maintainer triage note for @akash-manna-sky** · by `@dpol1` · 2026-09-06 16:47 UTC
>
> Helpful heads-up from the maintainers — please address before this PR can be reviewed (see our [Pull Request quality criteria](https://github.com/apache/stormcrawler/blob/main/CONTRIBUTING.md)):
> - :x: **Merge conflicts** with `main` — please rebase locally and push again.
> - :x: **Failing CI checks**: the `rat` job (license / code-format validation) is red on the current head.
> - :warning: **Review feedback pending**: the open points from `@jnioche` (Aug 19) and `@dpol1` (Sep 1) in the comments above have no reply yet.
>
> **The ball is in your court** — you've been assigned to this PR. Fix the above, then mark it **Ready for review**. This is not a rejection, just an invitation to bring the PR up to standard. No rush.
>
> <sub>_Automated triage — may be imperfect; a maintainer takes the next look._</sub>


dpol1 (approved):

rzo1 (changes requested):
Thanks for sticking with this. A few things need fixing before merge:

**Blocking**
- **Redirects loop back:** redirect targets go through the same filter. Once `sid` is learned, `/c` → 302 `/c?sid=Y` is rewritten to `/c` and dropped as a duplicate, so new pages on that site stop being fetched. If the filtered target equals the source, please return it unchanged.
- **Rules apply to the whole host:** evidence ignores the path, so product pages can remove `brand` from listing pages where it matters. Links to `/about?productId=1..5` can also get `productId` removed host-wide. Since rules are final, this never recovers. Please scope rules by path, or let a "kept" canonical cancel a rule.
- **Wrong canonical:** `filter()` runs before the parse filters, so it only sees a persisted or transferred canonical. It then marks the source as known, and the real canonical is ignored. `CanonicalParamLearner` also reads `canonicals[0]`, the oldest value. I'd remove learning from `filter()` and read the last value.

**Smaller**
- The refcount never reaches zero: both `LinkParseFilter`s, Tika's `ParserBolt` and `DelegateRefresher` don't call `cleanup()`, so the next topology in a local JVM inherits the rules.
- `scope=domain` uses `getPLD`, which ignores private suffixes, so `alice.github.io` and `bob.github.io` share rules.
- Queries are only split on `&`. Removing `sid` from `sid=abc;item=7` also removes `item`.
- The ~30 MB estimate leaves out `knownSources`, and parameter names have no length cap.

**Tests/docs**
- `testConfidenceThreshold` never reaches the confidence check; the suite stays green without it. The `max.params` cap and the synchronized blocks aren't really covered either.
- Please add a test that runs the parser pipeline (`JSoupParserBolt` + `XPathFilter` + learner). It would have caught the canonical issue.
- `internals.adoc` says the filter "does nothing on its own", but `filter()` still learns.

PR #2074, opened 2026-08-27
Honour ignoreOutsideDomain when set on its own in HostURLFilter

rzo1:
`HostURLFilter.configure()` read the `ignoreOutsideDomain` parameter but tested the `ignoreOutsideHost` node for null. A configuration naming only `ignoreOutsideDomain` was silently discarded and the filter let every URL through; one naming only `ignoreOutsideHost: false` threw a `NullPointerException` at topology startup. `ignoreOutsideHost: true` was never affected, as it short-circuits before the domain parameter is read. Each parameter is now tested on its own.

Topologies setting only `ignoreOutsideDomain: true` will start restricting outlinks to the source domain, as configured. Configurations setting both keys, including all three archetypes, are unaffected.

### For all changes

### For code changes

- [ ] Have you ensured that the full suite of tests is executed via `mvn clean verify`? - `core` module tests only (348 run, all green)
- [x] Have you written or updated unit tests to verify your changes? - two new cases in `HostURLFilterTest`, both fail without the fix

### DIFF
--- a/core/src/main/java/org/apache/stormcrawler/filtering/host/HostURLFilter.java
+++ b/core/src/main/java/org/apache/stormcrawler/filtering/host/HostURLFilter.java
@@ -62,7 +62,7 @@ public class HostURLFilter extends URLFilter {
         // always the same
         if (!ignoreOutsideHost) {
             JsonNode filterByDomainNode = filterParams.get("ignoreOutsideDomain");
-            if (filterByHostNode == null) {
+            if (filterByDomainNode == null) {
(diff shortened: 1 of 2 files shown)


dpol1 (approved):
LGTM. One thing for the description: only `ignoreOutsideHost: false` on its own NPE'd, `true` never reached the domain branch. Worth being precise, since this and the "only `ignoreOutsideDomain` now restricts" sentence will end up in the 4.0 release notes.

jnioche (approved):

PR #2075, opened 2026-08-27
Scope cookies to the host that set them

rzo1:
`CookieConverter` only checked applicability for cookies carrying a `Domain` attribute. A cookie without one, which is the usual session cookie, skipped the check and was sent to whatever host was being fetched. The domain check also passed on exception and accepted a `Domain` covering every host under it, such as `com` or `co.uk`.

A three-argument `getCookies` taking the originating URL is added; the two-argument method is deprecated and kept. A `Domain` attribute is now normalised with `IDN.toASCII` and validated before use, so the unicode and punycode forms of a name are interchangeable and a malformed value such as `com..` is rejected instead of matching every host under `com`. The public suffix list already shipped with crawler-commons, private section included, decides whether a domain may cover subdomains at all; a domain the list does not know about, e.g. an internal name under `.local`, falls back to the plain suffix match so unlisted domains keep working. A domain which cannot own subdomains is ignored and the cookie is bound to the host, per RFC 6265 5.3. An address is matched only by itself, per RFC 6265 5.1.3, where `Domain=2.3.4` previously covered `9.2.3.4`.

Two smaller ones found on the way: a `Set-Cookie` header whose first token has no `=` threw `StringIndexOutOfBoundsException` out of `getCookies` and failed the whole fetch, and is now skipped; cookies dropped on the path, secure and expiry checks are logged like the other ones.

The protocols record which host set the cookies. When a response carries `Set-Cookie`, okhttp and playwright write its URL alongside as `set-cookie-origin`, dropping any inbound key of that name first so a response cannot choose where its cookies are sent. That key has to travel with `protocol.set-cookie` for host-only cookies to be sent back, so it is added to the `metadata.persist` and `metadata.transfer` examples and a warning naming both keys is logged once when cookies arrive without it. Also corrects the `metadata.transfer` example in `internals.adoc`, which named `set-cookie` where the key is `protocol.set-cookie`.

### For all changes

### For code changes

- [ ] Have you ensured that the full suite of tests is executed via `mvn clean verify`? - `core` module tests only
- [x] Have you written or updated unit tests to verify your changes? - `CookieConverterTest` extended

### DIFF
--- a/core/src/main/java/org/apache/stormcrawler/protocol/okhttp/HttpProtocol.java
+++ b/core/src/main/java/org/apache/stormcrawler/protocol/okhttp/HttpProtocol.java
@@ -23,6 +23,7 @@ import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.MalformedURLException;
 import java.net.Proxy;
+import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.security.cert.CertificateException;
 import java.util.ArrayList;
@@ -35,6 +36,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLSession;
@@ -102,6 +104,9 @@ public class HttpProtocol extends AbstractHttpProtocol {
 
     // track the time spent for each URL in DNS resolution
     private final Map<String, Long> DNStimes = new ConcurrentHashMap<>();
+
+    // makes sure that a missing cookie origin is reported once and not for every url
+    private final AtomicBoolean missingCookieOriginLogged = new AtomicBoolean();
 
     private OkHttpClient.Builder builder;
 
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/apache/stormcrawler/util/CookieConverter.java
```diff
@@ -129,6 +168,16 @@ public static List<Cookie> getCookies(String[] cookiesStrings, URL targetURL) {
...
     }
 
+    /** Checks whether a cookie domain is made of a single label, e.g. "com". */
+    private static boolean isSingleLabel(String domain) {
+        String d = domain;
+        if (d.startsWith(".")) {
+            d = d.substring(1);
+        }
+        int dot = d.indexOf('.');
+        return dot < 1 || dot == d.length() - 1;
```
dpol1: `com..` gets through: first dot at 3, `length - 1` is 4, so it counts as multi-label; then `split("\\.")` in `checkDomainMatchToUrl` drops the empty tokens and it matches every `.com` host. `getCookies({"sid=x; Domain=com.."}, http://evil.com, http://victim.com)` returns the cookie. `Domain=co.uk` has the same shape, `evil.co.uk` reaches `bank.co.uk`. Could we reject malformed domains and known public suffixes with the PSL support already in crawler-commons, with a test for unlisted/internal domains so those keep working?

core/src/main/java/org/apache/stormcrawler/protocol/okhttp/HttpProtocol.java
```diff
@@ -286,8 +286,10 @@ private void addCookiesToRequest(Builder rb, String url, Metadata md) {
             return;
         }
         try {
+            // the host whose response set the cookies is not kept in the metadata,
+            // hence the null origin: cookies without a domain attribute are dropped
             final List<Cookie> cookies =
-                    CookieConverter.getCookies(cookieStrings, URLUtil.toURL(url));
+                    CookieConverter.getCookies(cookieStrings, null, URLUtil.toURL(url));
```
dpol1: With `null` here every host-only cookie is dropped, in practice most session cookies, login via 302 + Set-Cookie included, and nothing is logged. Could we record the origin instead? Only when a response actually carries `Set-Cookie`, write a sibling `set-cookie-origin` = fetched URL next to it in `responsemetadata`, and it has to travel with `protocol.set-cookie` in `metadata.transfer` (`putAll` alone won't do that). Then `getFirstValue(origin, protocolMetadataPrefix)` here. Should be about 15 lines. If not in this PR: a WARN at configure time, fix the cookies paragraph in `internals.adoc` and open a follow-up, or someone will spend an afternoon looking for their cookies.

dpol1 (changes requested):
Two things before merging: the new domain guard is bypassable, and the `null` origin disables every host-only cookie (no `Domain`, in practice most session cookies) for anyone running `http.use.cookies`. Threads.

Side note on "both protocol implementations": httpclient went with #2071, only okhttp calls this now.

core/src/main/java/org/apache/stormcrawler/util/CookieConverter.java
rzo1: Replaced the label check with normalisation plus the crawler-commons PSL. Empty labels are kept when splitting, so `com..`, `com...` and `.com..` are rejected instead of matching every host under `com`, and a domain which is itself a suffix (`com`, `co.uk`, `pvt.k12.ma.us`, `github.io` from the private section, the `*.ck` wildcard) can no longer cover subdomains — it is ignored and the cookie binds to the host, per RFC 6265 5.3. Domains the list does not know, `.local` and `.internal` included, fall back to the plain suffix match and are covered by tests, as is a single-label intranet domain.

core/src/main/java/org/apache/stormcrawler/protocol/okhttp/HttpProtocol.java
rzo1: Recorded rather than deferred. When a response carries `Set-Cookie`, okhttp and playwright write its URL alongside as `set-cookie-origin`, dropping any inbound key of that name first so a response cannot pick where its cookies go. `addCookiesToRequest` reads it back with the protocol prefix. It has to travel with `protocol.set-cookie` in `metadata.transfer`, so it is in the examples and a warning naming both keys is logged once when cookies arrive without it; the cookies paragraph in `internals.adoc` is fixed too. Host-only cookies round-trip in the new `HttpProtocolCookieTest`, including via a 302.

core/src/main/java/org/apache/stormcrawler/util/CookieConverter.java
rzo1: Follow-up: the hand-rolled normalisation is gone in favour of `IDN.toASCII`, which is what crawler-commons calls internally (`EffectiveTldFinder.normalizeName` and `BasicURLNormalizer` are both private, so there is nothing public to reuse directly). It rejects the empty-label family itself — `com..`, `com...`, `..example.com`, `some..url.com` all throw — and makes the unicode and punycode forms interchangeable, so `Domain=münchen.de` now matches `xn--mnchen-3ya.de` either way round. Also closed the address case while here: an address is matched only by itself per RFC 6265 5.1.3, where `Domain=2.3.4` set by `1.2.3.4` used to reach `9.2.3.4`.

dpol1 (approved):

PR #2076, opened 2026-08-27
Escape the metadata column of the SQL status store

rzo1:
`StatusUpdaterBolt` wrote the metadata column by concatenating `\t` + key + `=` + value with no escaping, and the read side split on those same characters, so a value containing a tabulation came back as one or more extra keys. The column is now written in an escaped form behind a `v1` marker; rows written by earlier versions are still decoded exactly as before, so no migration is needed. Note a downgrade is not clean, and existing rows are not repaired: worth grepping the metadata column of an existing table for tabulations.

Also guards `Date.from(Instant.parse(...))` in `AbstractStatusUpdaterBolt`, which sat outside the try/catch, so an unparsable `status.store.as.is.with.nextfetchdate` no longer takes the bolt down; it is logged and the URL goes through the normal scheduler. `SQLSpout` no longer routes through `StringTabScheme`, whose behaviour is unchanged for the seed-file spouts that rely on it.

### For all changes

### For code changes

- [ ] Have you ensured that the full suite of tests is executed via `mvn clean verify`? - `core` and the full `external/sql` suite against the MySQL container (18 tests, all green), not a full `mvn clean verify`
- [x] Have you written or updated unit tests to verify your changes? - round-trip tests for the new encoding and for the date guard

### DIFF
--- a/core/src/main/java/org/apache/stormcrawler/persistence/AbstractStatusUpdaterBolt.java
+++ b/core/src/main/java/org/apache/stormcrawler/persistence/AbstractStatusUpdaterBolt.java
@@ -21,6 +21,7 @@ import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import java.time.Instant;
 import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


dpol1 (approved):
LGTM. Checked the legacy path against `StringTabScheme` on a pile of random rows, identical, so "no migration" holds.

Two one-liners, not blocking. The new catch around `Instant.parse` in `AbstractStatusUpdaterBolt` only takes `DateTimeParseException`, but `Date.from` throws `IllegalArgumentException` on overflow, so `+1000000000-12-31T23:59:59Z` still escapes `execute()`. And the loop in `MetadataColumn.encode` NPEs on an empty key: `getValues("")` returns null while `keySet()` lists it, a legacy `\t=x` row gets you there.

PR #2077, opened 2026-08-27
Only follow robots.txt redirects on the same host

rzo1:
`HttpRobotRulesParser` resolved the `Location` header of a robots.txt redirect and re-fetched it without checking scheme, host or port, so the fetch could end up on any host the redirect named.

A redirect which only replaces http with https on the same host and port is still followed, whatever this is configured to, so the usual `http://…/robots.txt` -> `https://…/robots.txt` is unaffected. Anything else is now followed only when the target is http or https and shares scheme, host and port with the URL it was reached from; `http.robots.redirect.crossorigin.allow` (default false) restores the old behaviour for http and https targets.

`getCacheKey` derives the default port from the scheme, so an upgraded target has a different key although no other host is involved, and refusing it would leave that very common redirect with no rules at all. Only that direction is exempt — https to http would put the configured request headers on the wire in clear and let rules read over plain text replace the ones the redirect was reached from.

A redirect which is not followed no longer ends as allow-all in the success cache. It yields no rules for the host, cached in the error cache like a 429 or a 5xx, so nothing is crawled there until the redirect is resolved; `http.robots.redirect.refused.allow: true` restores allow-all. The refusal is logged at warn level with the target, the reason and the setting that changes it.

Following a redirect to another origin re-fetches robots.txt with the headers configured through `http.basicauth.*` and `http.custom.headers`, as any other request does. That is now stated where the setting is defined and logged once at configuration time when either is set. `http.filter.ipaddress.exclude` gained a `linklocal` rule to restrict which addresses those fetches may reach — link-local is matched by neither `loopback` nor `sitelocal`, so the documented `"localhost,sitelocal"` example did not cover it.

Sites serving robots.txt through a redirect to another host, a CDN for instance, are affected. Documented in `crawler-default.yaml` and `configuration.adoc`.

### For all changes

### For code changes

- [ ] Have you ensured that the full suite of tests is executed via `mvn clean verify`? - `core` module tests only
- [x] Have you written or updated unit tests to verify your changes? - new `HttpRobotRulesParserRedirectTargetTest`; the existing redirect test sets the new key, since its chain crosses ports

### DIFF
--- a/core/src/main/java/org/apache/stormcrawler/protocol/HttpRobotRulesParser.java
+++ b/core/src/main/java/org/apache/stormcrawler/protocol/HttpRobotRulesParser.java
@@ -21,6 +21,7 @@ import com.github.benmanes.caffeine.cache.Cache;
 import com.google.common.primitives.Ints;
 import crawlercommons.robots.BaseRobotRules;
 import java.net.URL;
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
@@ -44,6 +45,10 @@ public class HttpRobotRulesParser extends RobotRulesParser {
 
     protected boolean allow5xx = false;
 
+    protected boolean allowCrossOriginRedirects = false;
+
+    protected boolean allowRefusedRedirect = false;
+
     protected Metadata fetchRobotsMd;
 
     private static final int MAX_NUM_REDIRECTS = 5;
@@ -63,6 +68,36 @@ public class HttpRobotRulesParser extends RobotRulesParser {
         int robotsTxtContentLimit = ConfUtils.getInt(conf, "http.robots.content.limit", -1);
         fetchRobotsMd.addValue("http.content.limit", Integer.toString(robotsTxtContentLimit));
         allow5xx = ConfUtils.getBoolean(conf, "http.robots.5xx.allow", false);
+        allowCrossOriginRedirects =
+                ConfUtils.getBoolean(conf, "http.robots.redirect.crossorigin.allow", false);
+        allowRefusedRedirect =
+                ConfUtils.getBoolean(conf, "http.robots.redirect.refused.allow", false);
+        if (allowCrossOriginRedirects) {
+            logForwardedRequestHeaders(conf);
+        }
+    }
+
+    /**
+     * Logs which of the configured request headers are sent to the target of a robots.txt redirect
+     * to another host. The headers are configured per protocol instance and the redirect is
+     * followed by re-fetching the target through the same instance, so every hop of the chain is
+     * requested with them.
+     */
+    private static void logForwardedRequestHeaders(Config conf) {
+        List<String> sources = new ArrayList<>();
+        if (StringUtils.isNotBlank(ConfUtils.getString(conf, "http.basicauth.user", null))) {
+            sources.add("http.basicauth.user");
+        }
+        if (!ConfUtils.loadListFromConf("http.custom.headers", conf).isEmpty()) {
+            sources.add("http.custom.headers");
+        }
+        if (!sources.isEmpty()) {
+            LOG.warn(
+                    "http.robots.redirect.crossorigin.allow is set: a robots.txt redirect is fetched "
+                            + "from the host named in the Location header, and the request headers "
+                            + "configured by {} are sent to it as to any other host",
+                    String.join(" and ", sources));
+        }
     }
 
     /** Compose unique key to store and access robot rules in cache for given URL. */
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/apache/stormcrawler/protocol/HttpRobotRulesParser.java
```diff
@@ -44,6 +44,8 @@ public class HttpRobotRulesParser extends RobotRulesParser {
 
     protected boolean allow5xx = false;
 
+    protected boolean allowCrossHostRedirects = false;
```
dpol1: Two cases hide behind `false`. `getCacheKey` treats `http:h:80` and `https:h:443` as different authorities, so the scheme upgrade on the same host dies on the first hop; no other host is involved, I'd always follow that one. Apex -> www and CDNs are a real host change: the robots RFC says a crawler should follow those (hence our five-hop cap), but it's a SHOULD, and an egress restriction is defensible under our security model. Keep it opt-in if you want, but exempt the same-host scheme upgrade?

core/src/main/java/org/apache/stormcrawler/protocol/HttpRobotRulesParser.java
```diff
@@ -149,7 +170,18 @@ public BaseRobotRules getRobotRulesSet(Protocol http, URL url) {
...
+                    URL target = URLUtil.resolveUrl(redir, redirection);
+                    if (!followRedirect(redir, target)) {
+                        LOG.debug(
+                                "Robots for {} redirected to {} which is not fetched "
+                                        + "(not on the same host as {})",
+                                url,
+                                target,
+                                redir);
+                        // handled like any other response which does not provide rules
+                        break;
```
dpol1, line 269: The new tests make rejected redirect -> allow-all explicit, and that 3xx is then cached as a successful `EMPTY_RULES` result for 6h. I don't think an egress restriction should silently turn into permission to crawl without rules. Could we at least keep it out of the success cache and decide explicitly between fail-closed and allow-all?

core/src/test/java/org/apache/stormcrawler/protocol/HttpRobotRulesParserRedirectTest.java
```diff
@@ -79,6 +79,8 @@ void setUp() {
         mockServer7.start();
         mockServer8.start();
         conf.put("http.agent.name", "this_is_only_a_test");
+        // the redirect chains tested below point at other ports of the same host
+        conf.put("http.robots.redirect.crosshost.allow", true);
```
dpol1: With the flag on here nothing runs end to end with the default, and `localhost:8090 -> :8091` is exactly the chain that now ends as allow-all. One case without the flag? Same host `http -> https` in the new test would be good too.

dpol1 (changes requested):
Not convinced by the default. As it stands `http://h/robots.txt -> https://h/robots.txt` (same host, only the scheme changes) gets refused, and from there we crawl the site with no rules at all. Tried it on a scratch test. Threads.

core/src/main/java/org/apache/stormcrawler/protocol/HttpRobotRulesParser.java
rzo1: Exempted, one way only. `isSchemeUpgrade` follows http -> https on the same host when the port is unchanged or both sides use their scheme default, so `http://h/robots.txt -> https://h/robots.txt` and the `:443` / explicit-same-port variants are followed by default; `http://h:8080 -> https://h:9443` stays refused. The downgrade is deliberately not exempt, since it would put the configured `http.basicauth.*` / `http.custom.headers` on the wire in clear and let rules read over plain text replace the ones the redirect was reached from. Also renamed the flag to `http.robots.redirect.crossorigin.allow` while it is still unreleased — it governs scheme and port as well as host, and one scheme change is no longer cross-anything.

core/src/main/java/org/apache/stormcrawler/protocol/HttpRobotRulesParser.java
rzo1: Agreed, and decided explicitly rather than by fall-through. A refused redirect now sets `cacheRule = false`, so it lands in the error cache like a 429 or a 5xx instead of the success cache, and the hosts on the chain get nothing stored under their keys either. Default is fail-closed: no rules for that host, so nothing is crawled there until the redirect is resolved. `http.robots.redirect.refused.allow: true` restores allow-all for anyone who wants it. It is logged at warn with the target, why it was not followed and which setting changes it, so the two outcomes are distinguishable from a genuinely absent robots.txt.

core/src/test/java/org/apache/stormcrawler/protocol/HttpRobotRulesParserRedirectTest.java
rzo1: Added. `HttpRobotRulesParserRedirectTest` now has `testRedirectOnTheSameAuthorityIsFollowedByDefault` and `testRedirectToAnotherAuthorityIsNotFollowedByDefault`, both on the default conf, plus `testHeadersNotSentToAnUnfollowedRedirectTarget`. The same-host http -> https case is covered at unit level in `HttpRobotRulesParserRedirectTargetTest` (upgrade, explicit `:443`, explicit same port, downgrade, other host, other port) rather than end to end — core has no keystore test setup, and a genuine upgrade compares `http:h:80` against `https:h:443`, which the wiremock harness cannot represent.

dpol1 (approved):
LGTM. The one-way scheme-upgrade exemption and the refused redirect going to the error cache instead of allow-all cover both points, and `refused.allow` keeps the old behaviour one line away for whoever needs it. Only the description: it still opens with "shares scheme, host and port", the http -> https exemption deserves the first sentence since that's what most people will hit.

PR #2110, opened 2026-08-29
Make the WARC digest algorithm configurable

abhinav-phi:
This PR addresses #2106 and fixes it.

## Problem

`WARCRecordFormat.getDigestSha1()` computed a SHA-1 digest over the bytes and returned it as `sha1:<base32>`. That fixed value was used for the `WARC-Payload-Digest` and `WARC-Block-Digest` headers of response, resource, request and metadata records. There was no configuration key for the algorithm, so an operator who wants SHA-256 digests in their archives — which WARC 1.1 allows — could not have them without rewriting the records afterwards.

SHA-1 digests are the convention across the WARC ecosystem and identify content for deduplication rather than authenticate it, so this is not urgent — but it was still a fixed choice in a place where the digested bytes come from the network. Operators whose own policy rules out SHA-1, or whose downstream tooling wants SHA-256, had no option.

## Solution

Add a configuration key `warc.digest.algorithm` and thread the configured algorithm through all record formats:

```
warc.digest.algorithm: sha256
```

* **Accepted values**: `sha1` and `sha256`. The value is matched case-insensitively, an optional hyphen is ignored (`sha256`, `SHA-256`, `SHA256` are all accepted) and surrounding whitespace is trimmed.
* **`sha1` remains the default**: CDX indexes and revisit tooling downstream assume `sha1:` base32, and changing the default would break them. Deciding on a different default can be done separately from adding the option.
* **Fail fast on invalid values**: an unsupported value (e.g. `md5`) raises an `IllegalArgumentException` with a descriptive message when the bolt is prepared, instead of silently producing digests with a different algorithm than the one configured.
* **Digest values without Base32 padding**: the WARC digest fields define the digest value as a token (ISO 28500 WARC 1.1), which does not allow the `=` padding character, so digest values are written as unpadded Base32 (e.g. suggested in review). SHA-1 digests are exactly 32 characters without padding and are unaffected; SHA-256 digests would otherwise end in `====`.

### Changes

**`external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java`**

* New configuration key constants: `DIGEST_ALGORITHM_PARAM` (`warc.digest.algorithm`) and the supported values `DIGEST_ALGORITHM_SHA1` / `DIGEST_ALGORITHM_SHA256`.
* New constructor `WARCRecordFormat(String protocolMDprefix, String digestAlgorithm)`; the existing single-argument constructor is unchanged in behaviour and keeps defaulting to SHA-1.
* New instance methods `getDigest(byte[])` and `getDigest(byte[], byte[])` emit the matching `sha1:` / `sha256:` prefix. The digest used for records without content (`digestNoContent`) is now derived from the configured algorithm instead of a static SHA-1 constant.
* The static helpers `getDigestSha1(byte[])` and `getDigestSha1(byte[], byte[])` are removed (suggested in review): 4.0.0 is a major release, so the deprecated API can go straight away. Use the instance `getDigest` methods instead.

**`external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRequestRecordFormat.java`** and **`MetadataRecordFormat.java`**

* New constructors accepting the digest algorithm; both use the configurable `getDigest` instance methods so request and metadata records use the same algorithm as response and resource records.

**`external/warc/src/main/java/org/apache/stormcrawler/warc/WARCHdfsBolt.java`**

* Reads `warc.digest.algorithm` from the topology configuration (default `sha1`) and passes it to every record format it instantiates, so a single setting governs the whole WARC output.

### Documentation

* `external/warc/README.md`: documents the new key, the accepted values and the compatibility considerations around SHA-1.
* Both also document that digest values are written as unpadded Base32, as required by the WARC digest field grammar.
* `docs/src/main/asciidoc/configuration.adoc`: adds `warc.digest.algorithm` to the WARC section of the configuration reference.

### Note on #2034

As requested in the issue, I checked PR #2034 ("WARC writer: WARC-Protocol header to follow WARC field proposals"): it is about the `WARC-Protocol` / cipher-suite headers and does not touch digest computation, so this is not a duplicate. It modifies the same file, so whichever merges second may need a trivial rebase.

## Testing

* New `WARCDigestAlgorithmTest` (11 tests):
  * SHA-1 is the default, including for a `null` algorithm value;
  * SHA-256 digests for `getDigest(byte[])` and `getDigest(byte[], byte[])`, including empty content;
  * acceptance of value variants (`SHA256`, `SHA-256`, ` sha256 `, `SHA-1`);
  * `IllegalArgumentException` for unsupported values (`md5`, `sha512`) on all three record format classes;
  * blank (`""`, whitespace) algorithm values select the default SHA-1 like `null`;
  * digest values contain no Base32 padding (`=`), for both SHA-1 and SHA-256;
  * `getDigest` rejects null byte arrays (`NullPointerException`);
  * full response, resource, request and metadata records verifying that `WARC-Payload-Digest` / `WARC-Block-Digest` carry the `sha256:` prefix and the correct digest value, computed independently of the code under test.
* The pre-existing `WARCRecordFormatTest` SHA-1 digest tests now exercise the instance methods, covering the default algorithm (plain, empty, two byte arrays, robots.txt digests).
* `WARCHdfsBoltTest.testDigestAlgorithmConfig` prepares the bolt with `warc.digest.algorithm: sha256` and verifies that the written records (warcinfo, request, response) use `sha256:` digests.
* `mvn -pl external/warc test`: 23 tests run, 0 failures, 0 errors; `checkstyle:check` reports 0 violations.

### DIFF
--- a/external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
+++ b/external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
@@ -181,15 +181,16 @@ public class WARCRecordFormat implements RecordFormat {
         return switch (digestAlgorithm.trim().toLowerCase(Locale.ROOT).replace("-", "")) {
             case DIGEST_ALGORITHM_SHA1 -> "SHA-1";
             case DIGEST_ALGORITHM_SHA256 -> "SHA-256";
-            default -> throw new IllegalArgumentException(
-                    "Unsupported value ["
-                            + digestAlgorithm
-                            + "] for "
-                            + DIGEST_ALGORITHM_PARAM
-                            + ", supported algorithms: "
-                            + DIGEST_ALGORITHM_SHA1
-                            + ", "
-                            + DIGEST_ALGORITHM_SHA256);
+            default ->
+                    throw new IllegalArgumentException(
+                            "Unsupported value ["
+                                    + digestAlgorithm
+                                    + "] for "
+                                    + DIGEST_ALGORITHM_PARAM
+                                    + ", supported algorithms: "
+                                    + DIGEST_ALGORITHM_SHA1
+                                    + ", "
+                                    + DIGEST_ALGORITHM_SHA256);
         };
     }
 
@@ -218,10 +219,10 @@ public class WARCRecordFormat implements RecordFormat {
     }
 
     /**
-     * Base32-encode a digest value without the trailing &quot;=&quot; padding characters: the
-     * WARC digest fields define the digest value as a token, which does not allow the padding
-     * character (cf. ISO 28500 WARC 1.1, WARC-Block-Digest / WARC-Payload-Digest). SHA-1 digests
-     * are unaffected (32 characters without padding), while e.g. SHA-256 digests would end in
+     * Base32-encode a digest value without the trailing &quot;=&quot; padding characters: the WARC
+     * digest fields define the digest value as a token, which does not allow the padding character
+     * (cf. ISO 28500 WARC 1.1, WARC-Block-Digest / WARC-Payload-Digest). SHA-1 digests are
+     * unaffected (32 characters without padding), while e.g. SHA-256 digests would end in
      * &quot;====&quot;.
      */
     private static String base32Unpadded(byte[] digest) {
--- a/external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
+++ b/external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
@@ -129,19 +129,108 @@ public class WARCRecordFormat implements RecordFormat {
             Pattern.compile("(?i)(?:Content-(?:Encoding|Length)|Transfer-Encoding)");
     protected static final String X_HIDE_HEADER = "X-Crawler-";
 
+    /**
+     * Configuration key setting the algorithm used to compute the WARC-Payload-Digest and
+     * WARC-Block-Digest fields. Supported values are {@value #DIGEST_ALGORITHM_SHA1} (the default)
+     * and {@value #DIGEST_ALGORITHM_SHA256}.
+     *
+     * <p>Note: SHA-1 is the convention across the WARC ecosystem and downstream tooling (CDX
+     * indexes, revisit record handling) may expect it. Change the default deliberately, not
+     * casually.
+     */
+    public static final String DIGEST_ALGORITHM_PARAM = "warc.digest.algorithm";
+
+    public static final String DIGEST_ALGORITHM_SHA1 = "sha1";
+
+    public static final String DIGEST_ALGORITHM_SHA256 = "sha256";
+
     private static final Base32 base32 = new Base32();
-    private static final String digestNoContent = getDigestSha1(new byte[0]);
 
     protected final String protocolMDprefix;
 
+    /** JCA name of the message digest algorithm, e.g. &quot;SHA-1&quot;. */
+    private final String digestJCAName;
+
+    /** Algorithm prefix of the WARC digest fields, e.g. &quot;sha1:&quot;. */
+    private final String digestPrefix;
+
+    private final String digestNoContent;
+
     public WARCRecordFormat(String protocolMDprefix) {
-        this.protocolMDprefix = protocolMDprefix;
+        this(protocolMDprefix, DIGEST_ALGORITHM_SHA1);
     }
 
+    public WARCRecordFormat(String protocolMDprefix, String digestAlgorithm) {
+        this.protocolMDprefix = protocolMDprefix;
+        this.digestJCAName = getDigestJCAName(digestAlgorithm);
+        this.digestPrefix = digestJCAName.toLowerCase(Locale.ROOT).replace("-", "") + ":";
+        this.digestNoContent = getDigest(new byte[0]);
+    }
+
+    /**
+     * Resolve the configured digest algorithm to the JCA name of the message digest. The value is
+     * matched case-insensitively and an optional hyphen is ignored, i.e. &quot;sha256&quot;,
+     * &quot;SHA-256&quot; etc. are all accepted.
+     *
+     * @throws IllegalArgumentException if the value is not a supported algorithm
+     */
+    private static String getDigestJCAName(String digestAlgorithm) {
+        if (digestAlgorithm == null) {
+            return "SHA-1";
+        }
+        return switch (digestAlgorithm.trim().toLowerCase(Locale.ROOT).replace("-", "")) {
+            case DIGEST_ALGORITHM_SHA1 -> "SHA-1";
+            case DIGEST_ALGORITHM_SHA256 -> "SHA-256";
+            default -> throw new IllegalArgumentException(
+                    "Unsupported value ["
+                            + digestAlgorithm
+                            + "] for "
+                            + DIGEST_ALGORITHM_PARAM
+                            + ", supported algorithms: "
+                            + DIGEST_ALGORITHM_SHA1
+                            + ", "
+                            + DIGEST_ALGORITHM_SHA256);
+        };
+    }
+
+    /**
+     * Compute the digest of the given bytes with the configured algorithm.
+     *
+     * @return digest in the form &quot;&lt;algorithm&gt;:&lt;base32&gt;&quot;, e.g.
+     *     &quot;sha1:...&quot;
+     */
+    public String getDigest(byte[] bytes) {
+        MessageDigest md = DigestUtils.getDigest(digestJCAName);
+        return digestPrefix + base32.encodeAsString(md.digest(bytes));
+    }
+
+    /**
+     * Compute the digest of the concatenation of the two given byte arrays with the configured
+     * algorithm.
+     *
+     * @return digest in the form &quot;&lt;algorithm&gt;:&lt;base32&gt;&quot;, e.g.
+     *     &quot;sha1:...&quot;
+     */
+    public String getDigest(byte[] bytes1, byte[] bytes2) {
+        MessageDigest md = DigestUtils.getDigest(digestJCAName);
+        md.update(bytes1);
+        return digestPrefix + base32.encodeAsString(md.digest(bytes2));
+    }
+
+    /**
+     * @deprecated use {@link #getDigest(byte[])} instead; the algorithm is set by {@link
+     *     #DIGEST_ALGORITHM_PARAM} and no longer fixed to SHA-1
+     */
+    @Deprecated
     public static String getDigestSha1(byte[] bytes) {
         return "sha1:" + base32.encodeAsString(DigestUtils.sha1(bytes));
     }
 
+    /**
+     * @deprecated use {@link #getDigest(byte[], byte[])} instead; the algorithm is set by {@link
+     *     #DIGEST_ALGORITHM_PARAM} and no longer fixed to SHA-1
+     */
+    @Deprecated
     public static String getDigestSha1(byte[] bytes1, byte[] bytes2) {
         MessageDigest sha1 = DigestUtils.getSha1Digest();
         sha1.update(bytes1);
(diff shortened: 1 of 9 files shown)


external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
```diff
@@ -93,19 +93,108 @@ public class WARCRecordFormat implements RecordFormat {
...
+
+    /**
+     * Compute the digest of the given bytes with the configured algorithm.
+     *
+     * @return digest in the form &quot;&lt;algorithm&gt;:&lt;base32&gt;&quot;, e.g.
+     *     &quot;sha1:...&quot;
+     */
+    public String getDigest(byte[] bytes) {
+        MessageDigest md = DigestUtils.getDigest(digestJCAName);
+        return digestPrefix + base32.encodeAsString(md.digest(bytes));
```
dpol1: One thing worth deciding before merge: `Base32.encodeAsString` pads, so sha256 digests end in `====`, which strictly speaking falls outside WARC 1.1's `token` grammar for digest values. jwarc copes, but would a strict CDX tool? Stripping the trailing `=` here would be cheap now, less so once archives exist. What do you think?

dpol1 (approved):
LGTM - default stays byte-identical sha1, all record formats share the configured algorithm, bad names fail at startup. Left one question inline about the sha256 padding.

external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
abhinav-phi: Agreed — good catch, thanks for flagging it before merge.

The spec backs this up: in ISO 28500 WARC 1.1 the digest fields are `labelled-digest = algorithm ":" digest-value` with `digest-value = token`, and the spec note explicitly says *"The grammar for digest-value disallows the character = which is used for padding in Base32"*. So `sha256:...====` is formally invalid, and as you say it's cheap to fix now and much less so once padded digests are in existing archives.

Commit a6e24d04 strips the trailing padding: both `getDigest` methods now route the Base32 output through a small helper that drops any trailing `=`. Two notes on the impact:

- SHA-1 is unaffected: a 20-byte digest is exactly 32 Base32 characters, so the default output is byte-for-byte identical to before. Only SHA-256 (52 characters + 4 `=`) changes.
- Unpadded Base32 is also what the rest of the ecosystem writes (e.g. heritrix's WARC writer), so strict CDX tooling gets what it expects, while lenient readers like jwarc are fine either way.

The test expectations were updated to the unpadded values and there is now an explicit assertion that the produced digest value never contains `=` for either algorithm; README and the configuration docs mention the unpadded encoding too.

external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
```diff
@@ -129,19 +129,120 @@ static String sanitizeWarcFieldValue(String value) {
...
+     */
+    private static String base32Unpadded(byte[] digest) {
+        return StringUtils.stripEnd(base32.encodeAsString(digest), "=");
+    }
+
+    /**
+     * @deprecated use {@link #getDigest(byte[])} instead; the algorithm is set by {@link
+     *     #DIGEST_ALGORITHM_PARAM} and no longer fixed to SHA-1
+     */
+    @Deprecated
```
rzo1: We are going to do a new major release anyway, so if this is deprecated it can also be removed imho.

external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
```diff
@@ -129,19 +129,120 @@ static String sanitizeWarcFieldValue(String value) {
...
     public static String getDigestSha1(byte[] bytes) {
         return "sha1:" + base32.encodeAsString(DigestUtils.sha1(bytes));
     }
 
+    /**
+     * @deprecated use {@link #getDigest(byte[], byte[])} instead; the algorithm is set by {@link
+     *     #DIGEST_ALGORITHM_PARAM} and no longer fixed to SHA-1
+     */
+    @Deprecated
     public static String getDigestSha1(byte[] bytes1, byte[] bytes2) {
```
rzo1: We are going to do a new major release anyway, so if this is deprecated it can also be removed imho.

what do others think? @jnioche ?

external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
```diff
@@ -129,19 +129,120 @@ static String sanitizeWarcFieldValue(String value) {
...
+    }
+
+    /**
+     * Compute the digest of the given bytes with the configured algorithm.
+     *
+     * @return digest in the form &quot;&lt;algorithm&gt;:&lt;base32&gt;&quot;, e.g.
+     *     &quot;sha1:...&quot;
+     */
+    public String getDigest(byte[] bytes) {
+        MessageDigest md = DigestUtils.getDigest(digestJCAName);
```
rzo1, line 228: input validation?

external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
```diff
@@ -129,19 +129,120 @@ static String sanitizeWarcFieldValue(String value) {
...
 
+    /**
+     * Resolve the configured digest algorithm to the JCA name of the message digest. The value is
+     * matched case-insensitively and an optional hyphen is ignored, i.e. &quot;sha256&quot;,
+     * &quot;SHA-256&quot; etc. are all accepted.
+     *
+     * @throws IllegalArgumentException if the value is not a supported algorithm
+     */
+    private static String getDigestJCAName(String digestAlgorithm) {
+        if (digestAlgorithm == null) {
```
rzo1: Shouldnt we treat `digestAlgorithmn.isBlank()` as SHA-1 too ?

external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
```diff
@@ -129,19 +129,120 @@ static String sanitizeWarcFieldValue(String value) {
...
+        this.digestJCAName = getDigestJCAName(digestAlgorithm);
+        this.digestPrefix = digestJCAName.toLowerCase(Locale.ROOT).replace("-", "") + ":";
+        this.digestNoContent = getDigest(new byte[0]);
     }
 
+    /**
+     * Resolve the configured digest algorithm to the JCA name of the message digest. The value is
+     * matched case-insensitively and an optional hyphen is ignored, i.e. &quot;sha256&quot;,
+     * &quot;SHA-256&quot; etc. are all accepted.
+     *
```
rzo1, line 193: documentation for the param? what happens if it is null -> return sha1

external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
```diff
@@ -129,19 +129,120 @@ static String sanitizeWarcFieldValue(String value) {
...
+    /** Algorithm prefix of the WARC digest fields, e.g. &quot;sha1:&quot;. */
+    private final String digestPrefix;
+
+    private final String digestNoContent;
+
     public WARCRecordFormat(String protocolMDprefix) {
+        this(protocolMDprefix, DIGEST_ALGORITHM_SHA1);
+    }
+
+    public WARCRecordFormat(String protocolMDprefix, String digestAlgorithm) {
```
rzo1, line 182: add javadoc

external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
```diff
@@ -129,19 +129,120 @@ static String sanitizeWarcFieldValue(String value) {
...
 
+    /** JCA name of the message digest algorithm, e.g. &quot;SHA-1&quot;. */
+    private final String digestJCAName;
+
+    /** Algorithm prefix of the WARC digest fields, e.g. &quot;sha1:&quot;. */
+    private final String digestPrefix;
+
+    private final String digestNoContent;
+
     public WARCRecordFormat(String protocolMDprefix) {
```
rzo1, line 166: add javadoc

external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
```diff
@@ -129,19 +129,120 @@ static String sanitizeWarcFieldValue(String value) {
...
+
+    /**
+     * Compute the digest of the concatenation of the two given byte arrays with the configured
+     * algorithm.
+     *
+     * @return digest in the form &quot;&lt;algorithm&gt;:&lt;base32&gt;&quot;, e.g.
+     *     &quot;sha1:...&quot;
+     */
+    public String getDigest(byte[] bytes1, byte[] bytes2) {
+        MessageDigest md = DigestUtils.getDigest(digestJCAName);
```
rzo1, line 245: input validation?

external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRequestRecordFormat.java
```diff
@@ -45,6 +45,10 @@ public WARCRequestRecordFormat(String protocolMDprefix) {
         super(protocolMDprefix);
     }
 
+    public WARCRequestRecordFormat(String protocolMDprefix, String digestAlgorithm) {
```
rzo1, line 59: add javadoc

rzo1 (changes requested):

external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
abhinav-phi: Removed in 727ec1df — agreed, no point deprecating across a major release.

external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
abhinav-phi: Removed in 727ec1df. Happy to reintroduce shims if @jnioche would rather keep them for downstream users, but for a major release dropping the deprecated statics outright seems right.

external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
abhinav-phi: Done in 727ec1df: getDigest(byte[]) now rejects null input via Objects.requireNonNull (it only ever receives arrays here — format() guards the content == null case — but the method is public API). The algorithm itself is validated once at construction time, which keeps the per-call path free of it.

external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
abhinav-phi: Done in 727ec1df: null or blank now selects SHA-1, same as an unset value.

external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
abhinav-phi: Done in 727ec1df: the javadoc now documents the parameter, including the null/blank → SHA-1 behaviour.

external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
abhinav-phi: Done in 727ec1df — javadoc added with @param/@throws.

external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
abhinav-phi: Done in 727ec1df — javadoc added.

external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
abhinav-phi: Done in 727ec1df: same as above, both arrays are null-checked.

external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRequestRecordFormat.java
abhinav-phi: Done in 727ec1df — javadoc added (and matching ones on MetadataRecordFormat).

rzo1 (approved):

external/warc/src/main/java/org/apache/stormcrawler/warc/WARCRecordFormat.java
jnioche: > We are going to do a new major release anyway, so if this is deprecated it can also be removed imho.
> 
> what do others think? @jnioche ?

perfect time to remove it indeed

PR #2111, opened 2026-08-29
CommaSeparatedToMultivaluedMetadata: replace quadratic token append with a single bulk write

abhinav-phi:
## Summary

`CommaSeparatedToMultivaluedMetadata` splits a single metadata value containing comma separated entries into multiple values for the same key, but it did so by calling `Metadata.addValue(key, token)` once per token. `addValue` grows the underlying array by one on every call — allocating a new array and copying all previously stored values — so a value with *n* tokens costs 1 + 2 + ... + n = **n(n+1)/2 array copies**. This is quadratic in the length of the value: a keyword tag string of tens of thousands of tokens blocks the parser bolt for over a second on a single metadata entry, and the cost scales with the square of `http.content.limit`.

### Changes

**1. `CommaSeparatedToMultivaluedMetadata` — single bulk write**

The filter now splits the value, drops blank tokens (the behaviour `Metadata.addValue` had), and stores everything with one `Metadata.setValues(key, values)` call, which does not copy per token. Appending one token at a time copies the whole array for every token; writing the complete array once does not.

**2. New optional `maxTokens` parameter**

Since a page can influence the content it gets parsed, an unbounded split lets a page generate an unbounded amount of metadata. The number of values produced from a single entry is now capped at `maxTokens` (128 by default, configurable per filter config); when the cap is exceeded the first `maxTokens` values are kept, the rest are dropped, and a warning is logged with the key and the counts (the raw value is not logged, it can be huge).

The cap is applied **after** blank tokens are dropped, so it counts real values: with `maxTokens: 2` and the value `",,,a,b,c"` the result is `[a, b]`, matching what the old code would have kept.

**3. `Metadata.addValues(String, String[])` — append with a single array copy**

`addValues` appended the new values with one `addValue` call per element. When the key already has values, it now allocates one array of the final size and does two `System.arraycopy` calls. The `values == null || values.length == 0` and absent-key cases are unchanged.

Note on semantics: previously, appending to an *existing* key skipped blank values (because it went through `addValue`), while setting the values of an *absent* key stored blanks as-is. `addValues` now treats both cases the same and stores blanks either way; `addValue` keeps its blank-skipping behaviour for single values.

### Performance

Measured with the reproduction from the issue (`CountingMetadata` counting `addValue` calls; 32768 tokens, 65535 chars, one warm-up round discarded, `-Xmx4g`, JDK 25):

| | before | after |
|---|---|---|
| 32768 tokens (cold run) | 1223 ms | 272 ms |
| 32768 tokens (warm runs) | ~1200 ms | ~30 ms |
| `addValue` calls for 1000 tokens | 1000 | 0 |

### Testing

- New unit test `CommaSeparatedToMultivaluedMetadataTest`: bulk-append assertion (≤ 1 `addValue` call for 1000 tokens), timing at 32768 tokens, blank-token handling, cap applied after blank filtering, default cap trimming a 200-token value to 128.
- Existing `CSVMetadataFilterTest` (8 keyword values through the parser bolt) still passes.
- `mvn -pl core clean test`: 419 tests, 0 failures, 0 errors.
- `mvn checkstyle:check test-compile` across the reactor: clean.

### Docs

The `maxTokens` parameter and default are documented in the filter's javadoc and in the CommaSeparatedToMultivaluedMetadata entry of `docs/src/main/asciidoc/internals.adoc`.

### DIFF
--- a/core/src/main/java/org/apache/stormcrawler/parse/filter/CommaSeparatedToMultivaluedMetadata.java
+++ b/core/src/main/java/org/apache/stormcrawler/parse/filter/CommaSeparatedToMultivaluedMetadata.java
@@ -19,7 +19,6 @@ package org.apache.stormcrawler.parse.filter;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -35,7 +34,8 @@ import org.w3c.dom.DocumentFragment;
 
 /**
  * Rewrites single metadata containing comma separated values into multiple values for the same key,
- * useful for instance for keyword tags.
+ * useful for instance for keyword tags. The number of values produced from a single entry is
+ * bounded by the optional {@code maxTokens} parameter, 128 by default.
  */
 public class CommaSeparatedToMultivaluedMetadata extends ParseFilter {
 
@@ -43,11 +43,10 @@ public class CommaSeparatedToMultivaluedMetadata extends ParseFilter {
             LoggerFactory.getLogger(CommaSeparatedToMultivaluedMetadata.class);
 
     /**
-     * Default upper bound on the number of tokens a single value is split into. It is above what a
-     * page fetched with the archetype's default 65536 byte http.content.limit can produce, yet
-     * bounds the metadata a page can generate when the content limit is raised.
+     * Default upper bound on the number of values a single comma separated entry can produce, so
+     * that a page cannot generate an unbounded amount of metadata.
      */
-    private static final int MAX_TOKENS_DEFAULT = 65536;
+    private static final int MAX_TOKENS_DEFAULT = 128;
 
     private final Set<String> keys = new HashSet<>();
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/stormcrawler/parse/filter/CommaSeparatedToMultivaluedMetadata.java
```diff
@@ -60,9 +83,23 @@ public void filter(String url, byte[] content, DocumentFragment doc, ParseResult
             }
             m.remove(key);
             String[] tokens = val.split(" *, *");
+            if (tokens.length > maxTokens) {
```
dpol1: The cap runs before blanks are filtered, so with a low cap `,,,a` loses the real token where the old code kept it - was that intended? Dropping blanks first would avoid it.

dpol1 (commented):
The O(n²) fix looks right - I checked blanks, all-blank values and locked metadata against the old path and behaviour matches. Curious about the cap though: #2107 didn't ask for one and `maxTokens` isn't documented anywhere. Do we want it in this PR? Happy either way - if it stays, a docs line plus the inline point; if not, the PR gets simpler.

core/src/main/java/org/apache/stormcrawler/parse/filter/CommaSeparatedToMultivaluedMetadata.java
```diff
@@ -33,8 +39,20 @@
...
 
+    private static final Logger LOG =
+            LoggerFactory.getLogger(CommaSeparatedToMultivaluedMetadata.class);
+
+    /**
+     * Default upper bound on the number of tokens a single value is split into. It is above what a
+     * page fetched with the archetype's default 65536 byte http.content.limit can produce, yet
+     * bounds the metadata a page can generate when the content limit is raised.
+     */
+    private static final int MAX_TOKENS_DEFAULT = 65536;
```
jnioche: this is ridiculously high. I am not against setting a max number of tokens if we are paranoid about websites generating a ludicrous amount of data to trip the crawler but this value is too large to be really useful and also introduces confusion as it refers to a limit which is expressed by bytes, not token count.
Please add this new behaviour to the documentation + have a default of 64 or 128 but not more.
Also would be good to remove the confusing reference to http.content.limit.
Thanks!

jnioche (changes requested):

dpol1 (approved):
lgtm

jnioche (approved):
thanks @abhinav-phi

PR #2117, opened 2026-09-02
Use batch endpoint in URLFrontier module for DISCOVERED urls

### LINKED ISSUE #2062: [Improvement] Use batch endpoint in URLFrontier module for DISCOVERED urls

URLFrontier 2.6 added a batch endpoint for sending  DISCOVERED urls, which should greatly improve the performance. Known URLs should continue to use the streaming mode.
In addition, a better control flow mechanism is available compared to the existing polling.
See https://github.com/crawler-commons/url-frontier/blob/master/client/src/main/java/crawlercommons/urlfrontier/client/PutURLs.java
for best practices

### How should we improve?

_No response_

abhinav-phi:
## Summary

URLFrontier [2.6](https://github.com/crawler-commons/url-frontier/releases/tag/2.6) added a batched endpoint for ingesting discovered URLs, `PutDiscovered(stream DiscoveredBatch) returns (stream BatchAck)`. The per-message cost of the streaming `PutURLs` endpoint is what caps the ingestion rate, and discovered URLs — the outlinks of the pages being parsed — are the bulk of what a crawl writes. This PR routes them through the batched endpoint, grouping up to `urlfrontier.batch.size` URLs (default 100) into one message, while known URLs (fetched, redirections, errors...) keep using the streaming mode, as suggested in #2062. The control-flow mechanism follows the best practices of the client shipped with URLFrontier ([`PutURLs.java`](https://github.com/crawler-commons/url-frontier/blob/master/client/src/main/java/crawlercommons/urlfrontier/client/PutURLs.java)).

Fixes #2062

## What the bolt does now

`StatusUpdaterBolt` keeps its overall structure: a `waitAck` cache holds the tuples per URL until the frontier acks them, and a semaphore bounds the messages in flight. What changes is the path each URL takes:

**Discovered URLs → `PutDiscovered`.** A DISCOVERED URL is appended to a buffer instead of being sent on its own. The buffer leaves as one `DiscoveredBatch` message when:
- it reaches `urlfrontier.batch.size` entries;
- a known URL arrives — the outlinks buffered so far belong to the page whose status is now being updated, a natural end-of-page boundary for the batch (the release notes for 2.6 explicitly call the outlinks of a page a natural batch);
- a partially filled batch has been open for more than a second (checked by a scheduled flusher every 100 ms), so that acks are not delayed when the crawl tails off;
- the bolt is throttled because the semaphore is exhausted — pushing the batch out frees the permits sooner.

Each batch carries a unique ID (`batch-N`); the frontier echoes it in the `BatchAck` with **one status per URL, in the order they were sent**, so every URL can be resolved against `waitAck` and its tuples acked (or failed on `FAIL`) exactly as the streaming path did. Permits are released in bulk per URL, unchanged.

**Known URLs → `PutURLs`**, one message per URL, untouched behaviour.

**Flow control without polling.** `store()` used to busy-poll `tryAcquire` with a sleep. It now waits on a monitor (`flow`) and is woken by `notifyAll()` from the places that change its conditions: permit releases (acks, evictions), transport on-ready notifications, and stream errors. The wait timeout (`urlfrontier.throttling.time.msec`, unchanged, 10 ms) is a backstop that also drives the periodic `waitAck.cleanUp()` which prevents the deadlock the old code guarded against. This mirrors the `isReady()`/on-ready pattern of the reference client instead of polling.

**Fallback for older frontiers.** If the frontier does not implement `PutDiscovered` (i.e. predates 2.6), the stream fails with `UNIMPLEMENTED`. The bolt detects this, switches batching off permanently for its lifetime, and re-sends whatever was buffered or still in flight individually on the streaming endpoint — no URLs are lost, and a pre-2.6 frontier behaves exactly as before this PR. A dead batch stream (any other error) drops the pending batch IDs and opens a fresh stream on the next flush; the affected tuples are failed locally so Storm replays them.

## Configuration

New key, documented in the module README:

```yaml
# max number of discovered URLs per PutDiscovered message (default 100, 0 disables batching)
urlfrontier.batch.size: 100
```

The module now depends on `urlfrontier-API` **2.6** (bumped from 2.5). The schema stays wire-compatible with 2.5 apart from the semantics noted in the 2.6 release notes, and the fallback covers servers that do not implement the new RPC, so no frontier upgrade is strictly required — though the batching only pays off against a 2.6 server.

## Tests

- `StatusUpdaterBoltTest.acknowledgesDiscoveredURLsSentInBatches` — 6 discovered URLs against a 2.6 frontier with batch size 2: all acked, ≥ 3 batches sent, nothing failed;
- `StatusUpdaterBoltTest.sendsDiscoveredURLsIndividuallyWhenBatchingDisabled` — `urlfrontier.batch.size: 0` sends individually and never opens a batch;
- `StatusUpdaterBoltTest.acksKnownURLsThroughStreamingEndpoint` — known URLs still ack through the streaming endpoint;
- `StatusUpdaterBoltFallbackTest` — runs the bolt against a `crawlercommons/url-frontier:2.5` container: the `UNIMPLEMENTED` error is caught, batching is switched off, and discovered URLs are acked through the streaming fallback;
- the existing `StatusUpdaterBoltTest` cases (ack with metadata, queue-stream emission, semaphore recovery after a frontier restart) pass unchanged on the new implementation.

All 64 module tests pass locally against real frontier containers (testcontainers), checkstyle, forbiddenapis and the google-java-format check are clean. The `StatusUpdaterBolt` JaCoCo ratios configured for this module still pass.

## Notes for reviewers

- **Dependency availability.** The url-frontier 2.6 release published its artifacts to Sonatype Central via a workflow that completed successfully on 2026-08-19, but the deployment is not yet visible on Maven Central (`urlfrontier-API` still lists 2.5 as latest; the portal deployment likely awaits the manual publish step, since `central-publishing-maven-plugin` defaults to `autoPublish=false`). This PR builds against the 2.6 API regardless — the protobuf/gRPC API is stable in the 2.6 tag — but CI will need `urlfrontier-API:2.6` to be resolvable; it may be worth pinging the crawler-commons maintainers to complete the release on the portal. If 2.6 cannot be published soon, an alternative is to keep the dependency at 2.5 and build the batch messages from the proto (not possible: `DiscoveredBatch`/`BatchAck` and the `putDiscovered` stub only exist in 2.6), so the version bump is inherent to the feature.
- The `waitAck`-cache eviction semantics, the semaphore accounting, the queue-stream emission and the channel management (multi-address assignment, reconnect on `TRANSIENT_FAILURE`) are carried over unchanged from the previous implementation.
- The `PutDiscovered` stream is deliberately not re-created on channel `TRANSIENT_FAILURE` (unlike the `PutURLs` stream): with `waitForReady` it survives connection blips, and abandoning it would orphan the batches already sent on it.

### DIFF
--- a/external/urlfrontier/src/main/java/org/apache/stormcrawler/urlfrontier/StatusUpdaterBolt.java
+++ b/external/urlfrontier/src/main/java/org/apache/stormcrawler/urlfrontier/StatusUpdaterBolt.java
@@ -18,6 +18,8 @@
 package org.apache.stormcrawler.urlfrontier;
 
 import static org.apache.stormcrawler.urlfrontier.Constants.URLFRONTIER_ADDRESS_KEY;
+import static org.apache.stormcrawler.urlfrontier.Constants.URLFRONTIER_BATCH_SIZE_DEFAULT;
+import static org.apache.stormcrawler.urlfrontier.Constants.URLFRONTIER_BATCH_SIZE_KEY;
 import static org.apache.stormcrawler.urlfrontier.Constants.URLFRONTIER_CACHE_EXPIREAFTER_SEC_KEY;
 import static org.apache.stormcrawler.urlfrontier.Constants.URLFRONTIER_CRAWL_ID_KEY;
 import static org.apache.stormcrawler.urlfrontier.Constants.URLFRONTIER_DEFAULT_HOST;
@@ -37,6 +39,8 @@ import crawlercommons.urlfrontier.CrawlID;
 import crawlercommons.urlfrontier.URLFrontierGrpc;
 import crawlercommons.urlfrontier.URLFrontierGrpc.URLFrontierStub;
 import crawlercommons.urlfrontier.Urlfrontier.AckMessage;
+import crawlercommons.urlfrontier.Urlfrontier.BatchAck;
+import crawlercommons.urlfrontier.Urlfrontier.DiscoveredBatch;
 import crawlercommons.urlfrontier.Urlfrontier.DiscoveredURLItem;
 import crawlercommons.urlfrontier.Urlfrontier.KnownURLItem;
 import crawlercommons.urlfrontier.Urlfrontier.StringList;
@@ -45,7 +49,12 @@ import crawlercommons.urlfrontier.Urlfrontier.URLInfo;
 import crawlercommons.urlfrontier.Urlfrontier.URLItem;
 import io.grpc.ConnectivityState;
 import io.grpc.ManagedChannel;
+import io.grpc.StatusRuntimeException;
+import io.grpc.stub.ClientCallStreamObserver;
+import io.grpc.stub.ClientResponseObserver;
 import io.grpc.stub.StreamObserver;
+import java.util.ArrayDeque;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
@@ -53,8 +62,11 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantLock;
 import org.apache.storm.task.OutputCollector;
 import org.apache.storm.task.TopologyContext;
@@ -73,15 +85,39 @@ import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * Persists the status of URLs in a URLFrontier service.
+ *
+ * <p>Known URLs (fetched, redirections, errors...) are sent one message at a time on the streaming
+ * {@code PutURLs} endpoint. Discovered URLs, which are the bulk of what a crawl writes, are grouped
+ * into batches and pushed on the batched {@code PutDiscovered} endpoint introduced in URLFrontier
+ * 2.6, which amortises the per-message cost that limits the ingestion rate. A partially filled
+ * batch is sent after a second at the latest, so that acks are not delayed when the crawl tails
+ * off. If the frontier does not implement {@code PutDiscovered}, the bolt detects it and falls back
+ * to sending discovered URLs individually on the streaming endpoint.
+ *
+ * <p>Flow control follows the client implementation shipped with URLFrontier
+ * (crawlercommons.urlfrontier.client.PutURLs): the sends wait on a monitor woken by the acks and by
+ * the transport's on-ready notifications, instead of polling.
+ */
 public class StatusUpdaterBolt extends AbstractStatusUpdaterBolt
         implements RemovalListener<String, List<Tuple>>,
                 StreamObserver<crawlercommons.urlfrontier.Urlfrontier.AckMessage> {
 
     private static final Logger LOG = LoggerFactory.getLogger(StatusUpdaterBolt.class);
 
+    /** how long a partially filled batch is held back before it is sent anyway */
+    private static final long BATCH_FLUSH_DELAY_MS = 1000;
+
+    /** how often the flusher checks whether a batch is due */
+    private static final long FLUSH_CHECK_INTERVAL_MS = 100;
+
     private ManagedChannel channel;
     private URLPartitioner partitioner;
-    private StreamObserver<URLItem> requestObserver;
+    private volatile URLFrontierStub frontier;
+    private volatile StreamObserver<URLItem> requestObserver;
+    private volatile StreamObserver<DiscoveredBatch> batchRequestObserver;
+    private volatile ClientCallStreamObserver<DiscoveredBatch> batchTransport;
 
     private Cache<String, List<Tuple>> waitAck;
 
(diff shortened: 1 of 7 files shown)


rzo1 (commented):

The structure is clear and the `UNIMPLEMENTED` fallback to 2.5 frontiers has a container test, which is more than I expected. Three paths lose tuples; see the inline comments.

Smaller points:

- `store()` checks `inFlightSemaphore.tryAcquire()` outside `synchronized (flow)`, so a `notifyAll` between the failed acquire and entering the monitor is missed. The timeout makes that latency rather than deadlock, but please add a comment saying it was considered.
- `flushBatch` calls `newPutDiscoveredStream()` while holding `batchLock`, and a synchronous `onError` from gRPC re-enters `onBatchError` and `batchLock` on the same thread. Re-entrant, so it works. Please document the `batchLock` then `sendLock` ordering, since nothing enforces it.
- `urlfrontier.batch.size` is in the module README but not in `docs/src/main/asciidoc/configuration.adoc`.
- Format check: the `Executors.newSingleThreadScheduledExecutor(` block in `prepare()`, and the `batchedBolt.prepare(...)` / `streamingBolt.prepare(...)` lines in `StatusUpdaterBoltTest`, will be rewritten by the formatter. CI runs it with `-Dskip.format.code=false`.

external/urlfrontier/src/main/java/org/apache/stormcrawler/urlfrontier/StatusUpdaterBolt.java
```diff
@@ -418,7 +751,142 @@ public void store(
...
+                    }
+                }
+            }
+            synchronized (sendLock) {
+                stream.onNext(batch);
+            }
+            eventCounter.scope("batched").incrBy(items.size());
+            eventCounter.scope("batches").incrBy(1);
+            LOG.debug("Sent batch {} with {} discovered URL(s).", batchID, items.size());
+        } catch (InterruptedException e) {
```
rzo1, line 850: This strands a batch.

`pendingBatches.put(batchID, items)` on line 829 happens before the `awaitTransport` wait. If that wait is interrupted, which `disableBatchingAndResend` causes through `batchFlusher.shutdownNow()`, this catch only re-sets the interrupt flag. The batch is never sent, stays in `pendingBatches` forever, and its tuples sit in `waitAck` until cache expiry (default one hour) fails them. Storm will have timed them out long before.

```suggestion
        } catch (InterruptedException e) {
            Thread.currentThread().interrupt();
            synchronized (batchLock) {
                pendingBatches.remove(batchID);
            }
            for (URLItem failed : items) {
                failTupleLocally(failed.getID());
            }
```

external/urlfrontier/src/main/java/org/apache/stormcrawler/urlfrontier/StatusUpdaterBolt.java
```diff
@@ -253,10 +403,168 @@ public void onNext(final crawlercommons.urlfrontier.Urlfrontier.AckMessage confi
...
+
+    /** Stops batching and pushes everything buffered or in flight through the streaming endpoint. */
+    private void disableBatchingAndResend() {
+        final List<URLItem> toResend = new ArrayList<>();
+        synchronized (batchLock) {
+            batching = false;
+            for (List<URLItem> items : pendingBatches.values()) {
+                toResend.addAll(items);
+            }
+            pendingBatches.clear();
```
rzo1, line 491: Same expiry-window problem as the interrupt path, and inconsistent with the `RuntimeException` path in `flushBatch`, which calls `failTupleLocally` straight away.

Relying on `waitAck` eviction means the tuples are only failed after `urlfrontier.cache.expireafter.sec`, long after Storm has timed them out.

```suggestion
        final List<URLItem> orphaned = new ArrayList<>();
        synchronized (batchLock) {
            for (List<URLItem> items : pendingBatches.values()) {
                orphaned.addAll(items);
            }
            pendingBatches.clear();
            batchRequestObserver = null;
            batchTransport = null;
        }
        for (URLItem item : orphaned) {
            failTupleLocally(item.getID());
        }
```

external/urlfrontier/src/main/java/org/apache/stormcrawler/urlfrontier/StatusUpdaterBolt.java
```diff
@@ -294,43 +615,55 @@ public void store(
...
-                        waitAckLock.unlock();
+                        Thread.currentThread().interrupt();
                     }
                 }
-            } catch (InterruptedException e) {
-                LOG.warn(
-                        "InterruptedException - (approx.) {} messages in flight.",
-                        inFlightSemaphore.getQueueLength());
-                Thread.currentThread().interrupt();
+                eventCounter.scope("timeSpentThrottling").incrBy(throttleTimeMS);
```
rzo1, line 639: `timeSpentThrottling` is now incremented by the full `throttleTimeMS` even when `flow.wait` returned immediately because an ack woke it, which is the whole point of this change.

Measure the elapsed time around the `wait` instead, otherwise the metric operators use to tune `urlfrontier.updater.max.messages` stops meaning anything.

external/urlfrontier/src/main/java/org/apache/stormcrawler/urlfrontier/StatusUpdaterBolt.java
abhinav-phi: Fixed in 41ea0ba5: the interrupt handler now removes the batch from `pendingBatches` and fails its tuples locally so Storm replays them - exactly as suggested. The batch had never been handed to the transport at that point, so nothing that reached the frontier is dropped.

external/urlfrontier/src/main/java/org/apache/stormcrawler/urlfrontier/StatusUpdaterBolt.java
abhinav-phi: Fixed in 41ea0ba5 as suggested: the batches still tracked in `pendingBatches` are drained under `batchLock` and their tuples failed immediately, consistent with the RuntimeException path in `flushBatch`. Nothing depends on the `waitAck` cache expiry anymore.

external/urlfrontier/src/main/java/org/apache/stormcrawler/urlfrontier/StatusUpdaterBolt.java
abhinav-phi: Fixed in 41ea0ba5: the wait is bracketed with `System.nanoTime()` and the metric now records the time actually spent in `flow.wait` - an ack waking the wait early no longer bills the full throttle interval. The 30-second warn threshold uses the measured sum as well.

dpol1 (approved):
LGTM.

PR #2118, opened 2026-09-03
Upgrade to Tika 4.0.0

### LINKED ISSUE #2063: Upgrade to Tika 4.0
Tim Allison <tallison@apache.org>
03:42 (6 hours ago)
to announce, <dev@tika.apache.org>, Tika

The Apache Tika project is pleased to announce the release of Apache
Tika 4.0.0. The release contents have been pushed out to the main
Apache release site and to the Maven Central sync.

Apache Tika is a toolkit for detecting and extracting metadata and
structured text content from various documents using existing parser
libraries.

Apache Tika 4.0.0 is the first stable release of the 4.x line. It
switches the default output to Markdown, moves parsing into
crash-isolated forked processes (tika-server, tika-app's async/fork
modes and tika-grpc all parse out-of-process via tika-pipes), and adds
vision-language-model parsers (Claude, Gemini, OpenAI). Configuration
is now JSON rather than XML (tika-app's --convert-config-xml-to-json
offers a starting point for migrating a 3.x parsers config).
tika-app, tika-server and tika-eval now ship as runnable zip
distributions (with drop-in pf4j pipes plugins) in place of shaded
jars, and the Docker "latest" tags now track 4.x. Tika 4.x
requires Java 17 or later.

Metadata keys have been renamed for consistency and provenance
with an opt-in compatibility filter that restores the 3.x key names
where direct mapping exists. The key-for-key rename tables are
in the metadata migration guide
(https://tika.apache.org/docs/4.0.x/migration-to-4x/metadata-changes-4x.html),
and tika-core bundles the same
mapping in a machine-readable form (metadata-migration-3x-4x.json).

As a new major version, 4.0.0 includes breaking changes. Please see
the migration guides (https://tika.apache.org/docs/4.0.x/migration-to-4x/)
and the notable changes page (https://tika.apache.org/4.0.0/) before
upgrading. The 3.3.x line remains supported and continues to receive
bug fixes and dependency upgrades.

Details can be found in the
changes file: https://www.apache.org/dist/tika/4.0.0/CHANGES-4.0.0.txt
and in our 4.x docs site: https://tika.apache.org/docs/4.0.x/

abhinav-phi:
## Upgrade to Apache Tika 4.0.0 — Closes #2063

This PR upgrades StormCrawler from Tika 3.3.2 to the first stable release of the Tika 4.x line ([announcement](https://lists.apache.org/thread/fo8ozc8hkkzmrbfmvojtocml4fbm2p3q)), tracking the 4.0.0 milestone. Tika 4.0 ships a number of breaking changes; this PR migrates the codebase, the bundled configuration and the documentation accordingly. No behaviour change is intended for users of the `JSoupParserBolt` and the Tika `ParserBolt` beyond the Tika upgrade itself.

### Dependency changes

- `tika.version` `3.3.2` → `4.0.0` in the root POM. Tika 4.x requires Java 17+, which StormCrawler already targets (the project builds on JDK 17-25, CI runs JDK 25), so no compiler settings were needed.
- In Tika 4.0, `tika-parsers-standard-package` changed from a **jar** to a **POM** that aggregates the individual `tika-parser-*-module` artifacts. `external/tika` now consumes it with `<type>pom</type>`, which pulls in all standard parser modules transitively — same effective set of parsers as before, including the OCR module.
- Added explicit `tika-core` and `tika-serialization` dependencies to `external/tika`: `tika-serialization` provides the new `TikaLoader` used to read the JSON configuration (see below). The existing exclusions (`asm`, `slf4j-log4j12`, and the BSD-licensed `jai-imageio-core`, still a compile dependency of the image parser module) are preserved.
- `THIRD-PARTY.txt` regenerated with `license:aggregate-add-third-party` (Tika modules now 4.0.0; new Tika artifacts: `tika-serialization`, `tika-encoding-detector-html`, `tika-encoding-detector-mojibuster`, `tika-ml-core`, `tika-ml-junkdetect`, `tika-parser-datauri-commons`; removals: `tika-parsers-standard-package` jar, `vorbis-java-tika`, POI `poi-ooxml-full`, PDFBox `jempbox`/`xmpbox` 3.0.8, `org.tukaani:xz` — replaced by Tika 4's internal dependencies).

### Code migration

**`core/JSoupParserBolt.java`** (only core module using Tika, via `tika-core`):

- `TikaConfig` was **removed** in Tika 4. The mime-type detector, previously obtained from `TikaConfig.getDefaultConfig().getDetector()`, is now built with `new DefaultDetector()`, which is exactly what the removed default configuration wired (bytecode of Tika 4's `Tika()` facade confirms it constructs `new DefaultDetector()`).
- `Detector.detect()` now takes `(TikaInputStream, Metadata, ParseContext)`; `guessMimeType()` wraps the content bytes with `TikaInputStream.get(content)` and passes an empty `ParseContext`.
- `Metadata` no longer implements `HttpHeaders`, so the `Metadata.CONTENT_TYPE` / `Metadata.CONTENT_LENGTH` constants used for **Tika metadata** are now taken from `org.apache.tika.metadata.HttpHeaders`. Lookups in StormCrawler's own `Metadata` (String keys) keep using `org.apache.http.HttpHeaders`. Code referencing `TikaCoreProperties` constants is unaffected (the `RESOURCE_NAME_KEY` usage stays).

**`external/tika/ParserBolt.java`**:

- XML configurations are no longer supported in Tika 4; configuration moved to JSON. `instantiateTika()` now loads the configuration through `TikaLoader.load(path, classLoader)` (from the new `tika-serialization` module) and wires the result into the `Tika` facade via `new Tika(loader.loadDetectors(), loader.loadAutoDetectParser())` — the equivalent of the removed `new TikaConfig(url, classLoader)` + `new Tika(tikaConfig)` pair.
- A configuration that is missing from the classpath or fails to load now **fails fast** (throws from `prepare()`, aborting the topology) instead of silently degrading to the Tika defaults — a silent downgrade would re-activate parsers the configuration excluded, e.g. OCR via `TesseractOCRParser`.
- The `ParseContext` used for every parse is now seeded from the **`parse-context`** section of the Tika configuration via `TikaLoader.loadParseContext()` + `ParseContext.copyFrom()` (per-parse contexts are copies, since a bolt is multi-threaded); previously that section was silently ignored. This is what makes Tika 4's new JSON-only parse-context components (`output-limits`, `timeout-limits`, `embedded-limits`, …) reachable for embedded parsing.
- `TikaLoader` can only read configuration files from the filesystem. When the configuration is bundled inside a jar (the common case for topologies built as fat jars), it is copied to a temporary file, which is deleted as soon as `TikaLoader.load()` returns (the file is read eagerly; nothing is kept for the worker's lifetime).
- The parse itself now uses `TikaInputStream.get(content)` in a try-with-resources block, replacing the manual `ByteArrayInputStream` + `finally` close (Tika 4's `Parser.parse()` requires a `TikaInputStream`, and `TikaInputStream` holds `TemporaryResources` that must be released).
- Unless `parser.extract.embedded` is set, an `EmptyParser` is now bound as `Parser.class` in the `ParseContext`. This fixes a pre-existing bug (present in 3.x as well): `AutoDetectParser` sets *itself* as the embedded parser when the context has none, so embedded documents were parsed (and their text emitted into the parent's XHTML) even though the configuration said not to extract them. See `DelegatingParser.getDelegateParser()`, which falls back to `context.get(Parser.class, EmptyParser.INSTANCE)`.
- Default value of `parser.tika.config.file` is now `tika-config.json`.

### Configuration migration

- `external/tika/src/main/resources/tika-config.xml` → `tika-config.json`. The old XML (a) excluded `TesseractOCRParser` from `DefaultParser` and (b) set `service-loader` warning handlers, which no longer exist as configurable handlers in Tika 4 at all — the `service-loader` section of Tika 4's JSON config deserializes into an empty POJO, so dropping it is correct rather than a behaviour change. Using Tika's own `XmlToJsonConfigConverter` semantics, this becomes:

  ```json
  {
    "parsers": [
      {
        "default-parser": {
          "exclude": ["tesseract-ocr-parser"]
        }
      }
    ]
  }
  ```

  (OCR stays disabled by default in StormCrawler — Tesseract is an optional native dependency.)

- Archetype `crawler-conf.yaml` files (core, OpenSearch, Solr) updated to `parser.tika.config.file: "tika-config.json"` — this also fixes a pre-existing typo in those files (`parse.tika.config.file`), which the bolt never read.

### Documentation

- `external/tika/README.md` and `docs/configuration.adoc` updated for the JSON configuration and the new default file name, and the README now states the fail-fast behaviour.

### Testing

- `mvn verify` with `CI_ENV=true`:
  - `stormcrawler-core`: **426 tests, 0 failures** (includes `JSoupParserBoltTest` mime-type detection).
  - `stormcrawler-tika`: **9 tests, 0 failures**:
    - new `TikaConfigTest` proves what is loaded rather than assuming it: the bundled configuration really excludes `TesseractOCRParser` (walks the `CompositeParser`'s component parsers), the `parse-context` section reaches the per-parse `ParseContext` (`writeLimit` assertion against `OutputLimits.get(context)`), and a broken or missing configuration aborts `prepare()`;
    - new `ParserBoltTest#testEmbeddedNotParsedByDefault` pins the `EmptyParser` behaviour — parsing `test_recursive_embedded.docx` with `parser.extract.embedded=false` emits a single document whose text does **not** contain the embedded documents' text (before the fix, it did);
    - existing tests (recursive embedded `.docx` through the `TikaLoader`/JSON-config path, mime-type whitelist, empty-subdocument handling from #2112) all green.
  - `stormcrawler-urlfrontier`, `stormcrawler-langid`, archetypes, docs: green.
- `-Prat -DskipTests verify -Dskip.format.code=false` (CI's license + format job): the new `tika-config.json` is approved by RAT; google-java-format validation passes on all touched Java files.
- Two test failures encountered locally are pre-existing Windows-environment issues, reproduced identically on a pristine checkout of `main` and unrelated to this change: `HttpRobotRulesParserRedirectTest` (WireMock cannot bind port 8089 on Windows) and `WARCHdfsBoltTest` (Hadoop requires `winutils.exe` on Windows); Testcontainers-based tests (OpenSearch/Solr/SQL) require Docker, unavailable locally. CI runs on Ubuntu with Docker where these all pass.

### Notes for reviewers

- The Tika 4 metadata key renames (`tk:` prefix) only affect keys **emitted by parsers**, which this bolt copies with a `parse.` prefix into StormCrawler metadata. Users with ParseFilters/indexing rules matching specific `parse.*` Tika keys may be affected by the renames; the upstream [migration guide](https://tika.apache.org/4.0.0/index.html) documents an opt-in legacy-key filter (`metadata-migration-3x-4x.json`) for that case.
- HTML remains the domain of `JSoupParserBolt`; the "default content handler is now Markdown" change does not affect StormCrawler, which always supplies its own `BodyContentHandler`/`LinkContentHandler`. Likewise, process-isolation/timeouts in Tika 4 (`PipesForkParser`, tika-pipes) are out of scope here: StormCrawler parses in-process, so `timeout-limits` only applies if/when Tika surfaces it for in-process parsing.
- Licensing of the new Tika artifacts pulled in by the aggregated POM (`tika-ml-junkdetect`, `tika-encoding-detector-mojibuster`, `tika-ml-core`): all three are Apache Tika (ASF) jars under the Apache License 2.0, verified against their bundled `META-INF/LICENSE`/`NOTICE`/`DEPENDENCIES`. The only third-party runtime dependency added is `org.commonmark:commonmark` (BSD-2-Clause) inside `tika-ml-junkdetect`, which is already an approved license category in `THIRD-PARTY.txt`. The bundled model resources (`junkdetect.bin`, `nb-bigram.bin`) are generated ASF artifacts excluded from RAT by Tika itself.


external/tika/src/main/resources/tika-config.json
```diff
@@ -0,0 +1,9 @@
+{
```
rzo1, line 1: This drops half of the config it replaces.

The deleted `tika-config.xml` carried two things this does not:

```xml
<service-loader initializableProblemHandler="ignore" loadErrorHandler="warn" />
<parser class="org.apache.tika.parser.CompositeParser">
    <parser-exclude class="org.apache.tika.parser.ocr.TesseractOCRParser"/>
```

Losing the service-loader handlers puts Tika warnings back on stderr in every worker, which the XML deliberately suppressed. Please carry both across, or explain why Tika 4 makes them unnecessary.

external/tika/src/main/java/org/apache/stormcrawler/tika/ParserBolt.java
```diff
@@ -343,6 +343,22 @@ private Tika instantiateTika(Map<String, Object> conf) {
         return tika;
     }
 
+    /**
+     * Returns the configuration file as a path, copying it to a temporary file first if it is
+     * bundled inside a jar, as TikaLoader can only read configs from the filesystem.
+     */
+    private Path urlToPath(URL configUrl) throws IOException, URISyntaxException {
```
rzo1, line 350: This writes a temp file every time `instantiateTika` runs and relies on `deleteOnExit`, which never fires for a killed worker. One file per bolt instance per restart, accumulating in `/tmp` with default permissions.

```suggestion
    private Path urlToPath(URL configUrl) throws IOException, URISyntaxException {
        if ("file".equals(configUrl.getProtocol())) {
            return Paths.get(configUrl.toURI());
        }
        Path tmp = Files.createTempFile("tika-config", ".json");
        try (InputStream is = configUrl.openStream()) {
            Files.copy(is, tmp, StandardCopyOption.REPLACE_EXISTING);
        }
        return tmp;
    }
```

and delete it in a `finally` in `instantiateTika` once `TikaLoader.load` has returned, rather than deferring to JVM exit.

rzo1 (commented):

The `TikaConfig` to `TikaLoader` migration and the XML to JSON config move look right, and dropping GPL/LGPL `juniversalchardet` from `THIRD-PARTY.txt` is a useful side effect.

**`HttpHeaders.CONTENT_TYPE`.** `JSoupParserBolt` and `ParserBolt` both import `org.apache.http.HttpHeaders`. The string values match Tika's so the code is correct, but in a Tika metadata call `org.apache.tika.metadata.HttpHeaders` is the constant that says what is meant. One line each.

**Release compliance.** `tika-parsers-standard-package` with `<type>pom</type>` now pulls in `tika-ml-junkdetect` and `tika-encoding-detector-mojibuster`. Please confirm what model artifacts those ship and under which licence before this goes into a release. I could not settle that from the diff alone.

external/tika/src/main/java/org/apache/stormcrawler/tika/ParserBolt.java
```diff
@@ -317,16 +316,17 @@ public void execute(Tuple tuple) {
     private Tika instantiateTika(Map<String, Object> conf) {
         Tika tika = null;
         String tikaConfigFile =
-                ConfUtils.getString(conf, "parser.tika.config.file", "tika-config.xml");
+                ConfUtils.getString(conf, "parser.tika.config.file", "tika-config.json");
```
rzo1, line 319: Pre-existing bug this PR carries forward: the code reads `parser.tika.config.file`, but all three archetype `crawler-conf.yaml` files write `parse.tika.config.file` (and this PR updates all three). The archetype setting has never had any effect.

Since this PR touches all four places, please fix the key here too, or accept both spellings for one release.

rzo1 (approved):
LGTM, thanks for addressing the review feedback and clarifying the licensing of the new Tika artifacts.

dpol1 (approved):
LGTM. Two more items for the combined upgrade note: embedded documents are no longer parsed unless `parser.extract.embedded` is set (3.x did it by default, and the key is not in configuration.adoc yet), and the Tika key renames come through as renamed `parse.*` keys, e.g. `parse.resourceName` is now `parse.tk:resource-name`.

PR #2132, opened 2026-09-06
#2129 FetcherBolt: lock-free fetch queues ordered by next fetch time

### LINKED ISSUE #2129: FetcherBolt: global lock and linear scan in FetchItemQueues stall the executor with many hosts
FetchItemQueues in FetcherBolt uses a single monitor for `addFetchItem()`, `getFetchItem()` and `finishFetchItem()`, and `getFetchItem()` rotates a `LinkedHashMap` linearly over every queue whose crawl delay has not elapsed.

With many hosts the fetcher threads hold that lock most of the time and the executor thread calling `execute()` stalls on every incoming tuple.

Benchmark (50 fetcher threads, 1 producer thread, 20 URLs per host, `fetcher.server.delay` 1s, 20000 hosts):

```
getFetchItem avg 1.58 ms, p99 5.0 ms
addFetchItem avg 14.1 ms, p99 458 ms, max 571 ms
```

Proposal: keep the queues in a `ConcurrentHashMap` and reference the ones that may be ready from a `DelayQueue` ordered by next fetch time, so that taking an item is O(log n) and adding never waits for the fetcher threads.

Same benchmark after the change: `getFetchItem` avg 24 us, `addFetchItem` p99 0.11 ms, throughput unchanged (bounded by politeness).

GGraziadei:
Fixes #2129.

`FetchItemQueues` used a single monitor for adding, taking and finishing items, and `getFetchItem()` rotated a `LinkedHashMap` linearly over every queue whose crawl delay had not elapsed. With many hosts the fetcher threads held that lock most of the time and the executor thread calling `execute()` stalled on every incoming tuple.

### Change

- Queues live in a `ConcurrentHashMap`; the ones that may have an item ready are referenced from a `DelayQueue` of tickets ordered by their next fetch time. Taking an item is O(log n) and adding never waits for the fetcher threads.
- Per-queue state (size bound, in-progress count, next fetch time, crawl delays) uses atomics; a per-queue monitor is used only for the add/reap race. Empty queues are removed from the map as soon as they drain.
- `FetchItem` is built before any lock is taken, so a DNS lookup in `byIP` mode no longer blocks the fetcher threads.

### Behaviour preserved

Politeness per queue, `asap` release, `fetcher.threads.per.queue` and `fetcher.maxThreads.*`, `fetcher.max.queue.size`, crawl delay overrides from metadata and robots.txt, queue modes, metrics and the debug dump.

### Behaviour change: order in which hosts are served

The old implementation served hosts round-robin in `LinkedHashMap` insertion order, rotating past the ones whose crawl delay had not elapsed. The new one serves them in strict `nextFetchTime` order: the host that became ready the longest time ago goes first, and hosts whose next fetch time falls in the same millisecond are ordered arbitrarily (`DelayQueue` does not guarantee FIFO among equal delays).

In practice:

- A host is never starved: once its delay has elapsed it is ahead of every host that became ready later.
- Newly added hosts are ready immediately, so under a steady stream of new hosts they are served before hosts still waiting on their delay, exactly as before.
- Operators who relied on the insertion-order rotation to spread fetches evenly across a fixed set of hosts will see a different, but not less fair, interleaving. Per-host rate is unchanged: it is still bounded by the crawl delay.

Two smaller observable details, both stated in the javadoc: `getQueueSize()` can transiently over-report by one while an offer is being rejected, and the debug dump produced by the trigger file is no longer a point-in-time snapshot but a smear over the time it takes to iterate.

### Tests

New `FetchItemQueuesTest`: politeness, `asap`, several hosts back to back, queue size bound, multiple threads per queue, crawl delay from metadata, reaping of empty queues, and a concurrent producers/consumers test checking that nothing is lost or duplicated. Existing `FetcherBoltTest` and `SimpleFetcherBoltTest` pass unchanged.

### Benchmark: the case this targets

The cost the change removes is the linear rotation over queues that are *not* ready. It shows when a fetcher instance holds many hosts and most of them are waiting on their crawl delay, which is the steady state of a broad crawl: every `getFetchItem()` walked all of them under the global monitor, and the executor thread calling `execute()` queued behind the fetcher threads for that monitor.

Micro-benchmark on `FetchItemQueues` alone (`main` vs this branch): 50 fetcher threads, 1 producer thread emulating the executor, 20 URLs per host, `fetcher.server.delay` 1s so that almost every queue is waiting, instant fetch, 8s per run, p99 from sampled calls.

| hosts | metric | before | after |
|---|---|---|---|
| 20000 | `getFetchItem` avg | 1.58 ms | 24 us |
| 20000 | `getFetchItem` p99 | 5.0 ms | 24 us |
| 20000 | `addFetchItem` avg | 14.1 ms | 24 us |
| 20000 | `addFetchItem` p99 | 458 ms | 0.11 ms |
| 2000 | `addFetchItem` p99 | 2.5 ms | 0.13 ms |
| 20000 | fetched/s | 20001 | 20000 |

Throughput is unchanged, as expected: it is bounded by politeness. What goes away is the executor stall and the CPU burnt scanning queues that are not ready. With `fetcher.server.delay` 0 (every queue always ready, no rotation) both versions do ~53k fetch/s, i.e. the new structure is not slower on the path where the old one was cheap.

With a few hundred hosts per fetcher instance the difference is not measurable; with tens of thousands it is the numbers above.

### Verification

- `FetcherBoltTest` (8) and `SimpleFetcherBoltTest` (8) pass unchanged; `FetchItemQueuesTest` adds 9 tests including a concurrent producers/consumers run and a deterministic reproduction of the lost-wakeup race fixed in the second commit.
- Full `mvn clean verify` on the whole project: 675 tests, 0 failures, 0 errors (Docker-backed modules included).
- CI on this PR is green.

---

### For all changes

### For code changes

- [x] If adding new dependencies to the code, are these dependencies licensed in a way that is compatible for inclusion under [ASF 2.0](http://www.apache.org/legal/resolved.html#category-a)? (no new dependencies)
- [x] If applicable, have you updated the LICENSE file, including the main LICENSE file? (not applicable)
- [x] If applicable, have you updated the NOTICE file, including the main NOTICE file? (not applicable)

### DIFF
--- a/core/src/main/java/org/apache/stormcrawler/bolt/FetcherBolt.java
+++ b/core/src/main/java/org/apache/stormcrawler/bolt/FetcherBolt.java
@@ -26,22 +26,23 @@ import java.net.URL;
 import java.net.UnknownHostException;
 import java.time.Instant;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashMap;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.concurrent.BlockingDeque;
+import java.util.Queue;
 import java.util.concurrent.CancellationException;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.DelayQueue;
+import java.util.concurrent.Delayed;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
-import java.util.concurrent.LinkedBlockingDeque;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.regex.Pattern;
@@ -138,7 +139,7 @@ public class FetcherBolt extends StatusEmitterBolt {
     }
 
     /** This class described the item to be fetched. */
-    private static class FetchItem {
+    static class FetchItem {
 
         String queueId;
         String url;
@@ -204,58 +205,85 @@ public class FetcherBolt extends StatusEmitterBolt {
      * proto/IP pair). It also keeps track of requests in progress and elapsed time between
      * requests.
      */
-    private static class FetchItemQueue {
-        final BlockingDeque<FetchItem> queue;
+    static class FetchItemQueue {
+        final Queue<FetchItem> queue = new ConcurrentLinkedQueue<>();
+
+        final String id;
+
+        /** Number of items in {@link #queue}; bounded by maxQueueSize. */
+        private final AtomicInteger size = new AtomicInteger();
 
         private final AtomicInteger inProgress = new AtomicInteger();
         private final AtomicLong nextFetchTime = new AtomicLong();
 
-        private long minCrawlDelay;
+        /** Whether a ticket for this queue is currently present in the ready queue. */
+        private final AtomicBoolean scheduled = new AtomicBoolean(false);
+
+        /** Set when the queue has been removed from the map because it was empty. */
+        private boolean removed = false;
+
+        private final int maxQueueSize;
         private final int maxThreads;
 
-        long crawlDelay;
+        volatile long minCrawlDelay;
+        volatile long crawlDelay;
 
         public FetchItemQueue(
-                int maxThreads, long crawlDelay, long minCrawlDelay, int maxQueueSize) {
+                String id, int maxThreads, long crawlDelay, long minCrawlDelay, int maxQueueSize) {
+            this.id = id;
             this.maxThreads = maxThreads;
             this.crawlDelay = crawlDelay;
             this.minCrawlDelay = minCrawlDelay;
-            this.queue = new LinkedBlockingDeque<>(maxQueueSize);
+            this.maxQueueSize = maxQueueSize;
             // ready to start
             setNextFetchTime(System.currentTimeMillis(), true);
         }
 
         public int getQueueSize() {
-            return queue.size();
+            return size.get();
         }
 
         public int getInProgressSize() {
             return inProgress.get();
         }
 
-        public void finishFetchItem(FetchItem it, boolean asap) {
-            if (it != null) {
-                inProgress.decrementAndGet();
-                setNextFetchTime(System.currentTimeMillis(), asap);
-            }
+        long getNextFetchTime() {
+            return nextFetchTime.get();
         }
 
-        public boolean addFetchItem(FetchItem it) {
-            return queue.offer(it);
+        /** Must be called with the monitor of this queue held. */
+        boolean offer(FetchItem it) {
+            if (removed) {
+                return false;
+            }
+            if (size.incrementAndGet() > maxQueueSize) {
+                size.decrementAndGet();
+                return false;
+            }
+            queue.add(it);
+            return true;
         }
 
-        public FetchItem getFetchItem() {
-            if (inProgress.get() >= maxThreads) {
-                return null;
-            }
-            if (nextFetchTime.get() > System.currentTimeMillis()) {
-                return null;
-            }
-            FetchItem it = queue.pollFirst();
+        FetchItem poll() {
+            FetchItem it = queue.poll();
             if (it != null) {
+                size.decrementAndGet();
                 inProgress.incrementAndGet();
             }
             return it;
+        }
+
+        boolean hasFreeSlot() {
+            return inProgress.get() < maxThreads;
+        }
+
+        boolean isReady(long now) {
+            return nextFetchTime.get() <= now;
+        }
+
+        void finish(boolean asap) {
+            inProgress.decrementAndGet();
+            setNextFetchTime(System.currentTimeMillis(), asap);
         }
 
         private void setNextFetchTime(long endTime, boolean asap) {
@@ -268,12 +296,35 @@ public class FetcherBolt extends StatusEmitterBolt {
     }
 
     /**
+     * A ticket in the ready queue: a queue which may have an item to fetch at {@code time}. Kept
+     * separate from the queue itself so that the ordering key is immutable while in the heap.
+     */
+    private record QueueTicket(FetchItemQueue fiq, long time) implements Delayed {
+
+        @Override
+        public long getDelay(TimeUnit unit) {
+            return unit.convert(time - System.currentTimeMillis(), TimeUnit.MILLISECONDS);
+        }
+
+        @Override
+        public int compareTo(Delayed o) {
+            return Long.compare(time, ((QueueTicket) o).time);
+        }
+    }
+
+    /**
      * Convenience class - a collection of queues that keeps track of the total number of items, and
      * provides items eligible for fetching from any queue.
+     *
+     * <p>Queues are kept in a {@link ConcurrentHashMap} and the ones which may have an item ready
+     * are referenced from a {@link DelayQueue} ordered by their next fetch time: taking an item is
+     * O(log n) and does not require a global lock, so the executor thread adding URLs is never
+     * blocked by the fetcher threads.
      */
-    private static class FetchItemQueues {
-        final Map<String, FetchItemQueue> queues =
-                Collections.synchronizedMap(new LinkedHashMap<>());
+    static class FetchItemQueues {
+        final Map<String, FetchItemQueue> queues = new ConcurrentHashMap<>();
+
+        private final DelayQueue<QueueTicket> ready = new DelayQueue<>();
 
         AtomicInteger inQueues = new AtomicInteger(0);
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/stormcrawler/bolt/FetcherBolt.java
```diff
@@ -430,55 +516,47 @@ public synchronized FetchItemQueue getFetchItemQueue(String id, Metadata metadat
...
                 }
-
-                FetchItemQueue fiq = nextEntry.getValue();
-
-                // We remove the entry and put it at the end of the map
-                i.remove();
-
-                // reap empty queues
-                if (fiq.getQueueSize() == 0 && fiq.getInProgressSize() == 0) {
+                if (!fiq.hasFreeSlot()) {
```
rzo1, line 543: Lost wakeup: this can leave a queue with items and a free slot but no ticket.

Interleaving: a fetch on this queue finishes between `hasFreeSlot()` returning false and `scheduled.set(false)` on line 540. `finishFetchItem` calls `schedule`, whose CAS from false to true fails because `scheduled` is still true, and then this thread sets it to false. Nothing re-issues a ticket until an unrelated URL for the same host arrives, so that host stalls.

The two branches below (lines 544 to 556) already clear the flag and then re-check, which is the pattern this one needs.

```suggestion
                if (!fiq.hasFreeSlot()) {
                    // clear first, then re-check: a fetch finishing in between
                    // would otherwise find the flag still set, skip scheduling,
                    // and leave the queue with items but no ticket
                    fiq.scheduled.set(false);
                    if (fiq.hasFreeSlot() && !fiq.queue.isEmpty()) {
                        schedule(fiq, fiq.getNextFetchTime());
                    }
                    continue;
                }
```

Please add a targeted test: `fetcher.threads.per.queue: 1`, take an item, assert the next one becomes available after `finishFetchItem`, driven from two threads so the race is reachable. Without it this failure shows up only as `concurrentProducersAndConsumersLoseNothing` occasionally hitting its 60 s timeout, which reads as flakiness.

rzo1 (commented):

The design holds up: one ticket per queue guarded by the `scheduled` CAS, an immutable ordering key in `QueueTicket`, and `FetchItem.create` (which does DNS in `byIP` mode) moved out of the critical section.

One lost-wakeup window, inline on line 538.

Other comments:

- Fairness changes from `LinkedHashMap` round-robin to strict `nextFetchTime` order, with arbitrary tie-breaking among queues sharing a millisecond. This is a component operators tune by feel, so please put the change in the description.
- `offer` increments `size` before the bound check and decrements on overflow, so `getQueueSize()` can transiently over-report. Harmless, but worth a comment.
- `logQueuesContent` is no longer synchronized. Safe with weakly consistent iterators, but the dump is now a smear rather than a snapshot. Say so in its javadoc.
- This rewrites the hottest path in the crawler. Please include before/after numbers for the case it targets, and confirm the existing `FetcherBolt` integration tests pass.

core/src/main/java/org/apache/stormcrawler/bolt/FetcherBolt.java
GGraziadei: Thanks, added the fix and tests.

core/src/main/java/org/apache/stormcrawler/bolt/FetcherBolt.java
```diff
@@ -204,60 +205,92 @@ public static FetchItem create(URL u, String url, Tuple t, String queueMode) {
...
+                size.decrementAndGet();
+                return false;
             }
-            FetchItem it = queue.pollFirst();
+            queue.add(it);
+            return true;
+        }
+
+        FetchItem poll() {
+            FetchItem it = queue.poll();
```
dpol1, line 299: Reap can run between `queue.poll()` and `inProgress.incrementAndGet()`: with `fetcher.threads.per.queue` > 1 (or an equivalent per-queue override), another fetch on the same queue finishing in that window can observe an empty queue and `inProgress == 0`, and remove the queue while the just-polled item is being dispatched. When that item finishes, `finishFetchItem` either hits the unknown-queue branch or, if the queue has been recreated, decrements the replacement queue's counter. This can undercount active fetches, potentially make the counter negative, and allow `maxThreads` to be exceeded.

Could we increment `inProgress` before `queue.poll()`, undo it on `null`, and add a deterministic regression test for this interleaving? Pausing the poll after it removes the last queued item, then finishing the other active fetch, should leave the original queue in the map; after resuming, the remaining fetch should be counted and finish normally.

dpol1 (changes requested):
One reap/poll race left, inline - *Flagged by Claude*

core/src/main/java/org/apache/stormcrawler/bolt/FetcherBolt.java
GGraziadei: Confirmed and fixed in 2986b7c.

`poll()` now reserves the slot first with a CAS loop (`tryAcquireSlot()`, refusing when `inProgress >= maxThreads`), then dequeues, and releases the slot again if the queue was empty. While an item is in flight `inProgress` is never zero, so `reapIfEmpty` cannot remove the queue from under it. The CAS also makes the bound exact when two threads pass `hasFreeSlot()` together with one slot left, which the old `hasFreeSlot()` + `poll()` pair allowed to overshoot by one. In `getFetchItem` the null-poll branch re-issues the ticket only if a slot is free and the queue is not empty; otherwise the fetch that finishes next schedules it.

Tests, both failing on the previous revision:
- `fetchFinishingDuringPollDoesNotReapTheQueue`: a package-private no-op `afterDequeue()` hook in `poll()` lets the test pause right after the last item is taken. The other active fetch then finishes; the queue must still be in the map. After resuming, the item is returned, `inProgress == 1`, finishing it goes to the original queue, and only then is the queue reaped. Before the fix the queue was gone at the first assertion.
- `pollEnforcesMaxThreadsAtomically`: `maxThreads = 1`, two queued items, second `poll()` returns null without losing the item; after `finish()` it is handed out.

Full `core` verify: 454 tests, 0 failures.

GGraziadei: Hi @dpol1 thanks for your comment. 
The reported path is repeatable and unintentional. I agree with the proposed solution, fixing it.

core/src/main/java/org/apache/stormcrawler/bolt/FetcherBolt.java
```diff
@@ -430,55 +549,52 @@ public synchronized FetchItemQueue getFetchItemQueue(String id, Metadata metadat
...
                 }
-
-                Map.Entry<String, FetchItemQueue> nextEntry = i.next();
-
-                if (nextEntry == null) {
+                final FetchItemQueue fiq = ticket.fiq();
+                final long now = System.currentTimeMillis();
+                if (!fiq.isReady(now)) {
+                    // the delay was extended after the ticket was issued: re-issue it
+                    ready.add(new QueueTicket(fiq, fiq.getNextFetchTime()));
```
dpol1, line 582: Why `return null` here rather than `continue`? Returning here can make this fetcher wait even when another host is ready.

dpol1 (approved):
The slot reservation closes it. One question inline.

jnioche (changes requested):
I like it very much, thanks @GGraziadei 

Here are a few observations from an AI agent I launched.

1. Racy max-update on crawl delays — FetcherBolt.java:540-548

The base version had public synchronized FetchItemQueue getFetchItemQueue(...) (confirmed at 2cb93d1c:357), which made this an atomic max:

if (fiq.minCrawlDelay < minDelay) fiq.minCrawlDelay = minDelay;
if (fiq.crawlDelay < delay)       fiq.crawlDelay = delay;

The refactor drops the method-level lock and only makes the fields volatile, which doesn't help a read-check-write. Two threads adding URLs for the same host with different crawl.delay metadata can both read the old value, both decide theirs is larger, and last-writer-wins — so the queue can settle on the smaller delay instead of the intended maximum, crawling a host faster than configured. It's reachable from more than the executor thread: :783 calls getFetchItemQueue from every fetcher thread. Fix is a small CAS loop (or an AtomicLong with accumulateAndGet(v, Math::max)) on the two fields.

2. poll() decrements size after dequeuing — FetcherBolt.java:285-291

FetchItem it = queue.poll();   // :285  item is gone
...
size.decrementAndGet();        // :291  counter catches up

offer() bounds on size.incrementAndGet() > maxQueueSize under synchronized (fiq) (:264, :424), but poll() at :580 doesn't hold that monitor. In the window between those two lines the queue is at max-1 real items while size still reads max, so a concurrent offer rejects, addFetchItem returns false, and the caller at :1215 fails the tuple — Storm replays a URL that actually had room. Only bites when queues sit near maxQueueSize, i.e. exactly the heavy-load regime the bound exists for.

Worth noting the offer() javadoc at :254-259 asserts the transient over-report is "harmless: the value is only used for metrics and the debug dump" — that's not true, size is the bound check. Cleanest fix is to reserve in the other direction, mirroring the existing tryAcquireSlot pattern: decrement before queue.poll() and increment back if it returns null. Then size transiently under-reports, which can't cause a spurious rejection and can't push the real queue over the bound.

3. Nit: stale ticket returns instead of retrying — FetcherBolt.java:566-570

When the head ticket's queue had its delay extended, the branch re-issues the ticket and return null, while the two sibling race branches (:571-579, :582-590) both continue. Since DelayQueue orders by ticket time rather than the fiq's live nextFetchTime, other tickets may be due right now, and the caller reads null as "nothing to do" and spin-waits 100 ms. The re-added ticket carries a future time so it can't be re-polled immediately — continue is safe here and avoids the stall.

Both real findings are consequences of the same trade in this PR (coarse lock → per-queue atomics); neither invalidates the approach, they're just spots where the old lock was doing work the new code doesn't replace.

core/src/main/java/org/apache/stormcrawler/bolt/FetcherBolt.java
GGraziadei: No reason that survives scrutiny: the ticket was due, only its queue is not, and other tickets may be. Changed to `continue` in 5f40940 (the re-issued ticket has a future time, so it cannot be polled straight back, and the loop is bounded). `staleTicketDoesNotHideAnotherReadyQueue` reproduces the stall: `a.net` at the head with its delay pushed to 10 s, `b.net` due, `getFetchItem` returned null on the previous revision.

jnioche (approved):
LGTM thanks

rzo1 (approved):

PR #2133, opened 2026-09-06
#2130 MetadataTransfer: single pass over the metadata for wildcard keys

### LINKED ISSUE #2130: MetadataTransfer: avoid an intermediate key set per wildcard for every outlink
`MetadataTransfer.filter()` is called for every outlink. For each wildcard key (e.g. `cookie.*`) it calls `Metadata.keySet(prefix)`, which streams over all the metadata keys and collects a new `Set`, then copies each matching key through `getValues()`/`setValues()` with key normalisation on both sides.

That is one intermediate set per wildcard per outlink, plus repeated key normalisation, on the hottest path of the parser bolts.

Proposal: pre-compile the configured keys into exact keys and wildcard prefixes once, and copy the matching entries in a single pass over the metadata map.

Micro-benchmark on a 12-key metadata with 3 wildcards and 2 exact keys: ~780 ns to ~265 ns per outlink.

GGraziadei:
Fixes #2130.

`MetadataTransfer.filter()` runs for every outlink. For each wildcard key (e.g. `cookie.*`) it called `Metadata.keySet(prefix)`, which streams over all the metadata keys and collects a new `Set`, then copied each matching key through `getValues()`/`setValues()` with key normalisation on both sides.

### Change

The configured keys are compiled once into exact keys and wildcard prefixes (cached per set, rebuilt if a subclass modifies the set), and the matching entries are copied in a single pass over the metadata map. Value arrays are shared as before, keys are already normalised.

Micro-benchmark on a 12-key metadata with 3 wildcards and 2 exact keys: ~780 ns to ~265 ns per outlink.

### Tests

`MetadataTransferTest` gains a case checking that wildcard prefixes are matched case-insensitively and selectively (`Cookie.*` matches `cookie.id` but not `cookies`), existing cases unchanged.

---

### For all changes

### For code changes

### DIFF
--- a/core/src/main/java/org/apache/stormcrawler/util/MetadataTransfer.java
+++ b/core/src/main/java/org/apache/stormcrawler/util/MetadataTransfer.java
@@ -17,7 +17,11 @@
 
 package org.apache.stormcrawler.util;
 
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import org.apache.commons.lang3.StringUtils;
@@ -123,7 +127,7 @@ public class MetadataTransfer {
      * the URL path.
      */
     public Metadata getMetaForOutlink(String targetUrl, String sourceUrl, Metadata parentMetadata) {
-        Metadata md = filter(parentMetadata, mdToTransfer);
+        Metadata md = filter(parentMetadata, transferFilter());
 
         // keep the path?
         if (trackPath) {
@@ -150,11 +154,11 @@ public class MetadataTransfer {
      * not necessarily transferred to the outlinks.
      */
     public Metadata filter(Metadata metadata) {
-        Metadata filteredMetadata = filter(metadata, mdToTransfer);
+        Metadata filteredMetadata = filter(metadata, transferFilter());
 
         // add the features that are only persisted but
         // not transferred like __redirTo_
-        filteredMetadata.putAll(filter(metadata, mdToPersistOnly));
+        filteredMetadata.putAll(filter(metadata, persistOnlyFilter()));
 
         return filteredMetadata;
     }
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/stormcrawler/util/MetadataTransfer.java
```diff
@@ -164,19 +168,86 @@ public Metadata filter(Metadata metadata) {
...
+                    prefixList.add(normalised.substring(0, normalised.length() - 1));
+                } else {
+                    exact.add(normalised);
+                }
+            }
+            this.exactKeys = exact;
+            this.prefixes = prefixList.toArray(new String[0]);
+        }
+
+        private boolean isFor(Set<String> filter) {
```
rzo1, line 229:
```java
return sourceSet == filter && sourceSize == filter.size();
```

This detects a size change but not a content change of the same size. `mdToTransfer` is `protected final Set<String>` with mutable contents, so a subclass doing

```java
mdToTransfer.remove("depth");
mdToTransfer.add("mycustom");
```

after the first `filter()` call keeps the size and leaves the compiled filter stale. Every outlink from then on carries the wrong metadata, silently.

The javadoc on line 205 says the cache is "rebuilt if the set has been modified since (e.g. by a subclass)", which is a stronger claim than the code makes good on.

core/src/main/java/org/apache/stormcrawler/util/MetadataTransfer.java
```diff
@@ -164,19 +168,86 @@ public Metadata filter(Metadata metadata) {
...
+            return sourceSet == filter && sourceSize == filter.size();
+        }
+    }
+
+    private volatile CompiledFilter compiledTransfer;
+    private volatile CompiledFilter compiledPersistOnly;
+
+    private CompiledFilter compile(Set<String> filter) {
```
rzo1: The two sets are only ever populated in `configure()`. If that stays true, none of this machinery is needed: compile once at the end of `configure()` into two final fields and drop the lazy cache, the two `volatile`s, the identity dispatch and the staleness heuristic.

That keeps the whole measured win, since the per-outlink cost you removed is the stream and `Set` allocation, not the compile.

If subclass mutation after `configure()` really has to be supported, the check needs to be sound: either key the cache on a copy of the set contents, or give subclasses an explicit `invalidateCompiledFilters()` to call.

As it stands the design pays for both options and is correct under neither.

core/src/main/java/org/apache/stormcrawler/util/MetadataTransfer.java
```diff
@@ -164,19 +168,86 @@ public Metadata filter(Metadata metadata) {
...
+        CompiledFilter compiled =
+                filter == mdToTransfer
+                        ? compiledTransfer
+                        : filter == mdToPersistOnly ? compiledPersistOnly : null;
```
rzo1: `filter(Metadata, Set)` is private and only ever called with the two fields, so this `null` branch is unreachable. It disappears if the compile moves into `configure()`.

core/src/main/java/org/apache/stormcrawler/util/MetadataTransfer.java
```diff
@@ -164,19 +168,86 @@ public Metadata filter(Metadata metadata) {
...
-            if (key.endsWith("*")) {
-                String prefix = key.substring(0, key.length() - 1);
-                for (String k : metadata.keySet(prefix)) {
-                    metadata.copy(filteredMetadata, k);
+        // wildcards: a single pass over the metadata for all the prefixes,
+        // without allocating an intermediate key set per prefix
+        if (compiled.prefixes.length > 0) {
+            for (Map.Entry<String, String[]> entry : source.entrySet()) {
+                final String key = entry.getKey();
+                if (entry.getValue().length == 0 || target.containsKey(key)) {
```
rzo1: Unguarded dereference of `entry.getValue()`.

`Metadata(Map)` wraps a caller-supplied map without validating it, so a `null` value array reaches this line and throws, where the old path called `getValues()`, got `null`, and skipped the key. The exact-key branch above is null-checked; this one is not.

Unlikely, but it is one `!= null` away.

core/src/test/java/org/apache/stormcrawler/util/MetadataTransferTest.java
```diff
@@ -152,4 +152,26 @@ void testFilterWithAsterisk() {
     }
 
     static class MyCustomTransferClass extends MetadataTransfer {}
+
+    @Test
+    void testWildcardPrefixIsCaseInsensitiveAndSelective() throws MalformedURLException {
```
rzo1, line 157: Good addition, and it pins the case-insensitive prefix behaviour.

Nothing covers the caching, which is the part carrying the risk. A test that calls `getMetaForOutlink`, then mutates `mdToTransfer` through a subclass without changing its size, then calls it again, would fail today.

rzo1 (commented):

The single-pass change itself is right. `Metadata.keySet(prefix)` builds a stream and collects a new `Set` per prefix per outlink, and that is worth removing.

I checked the behaviour that is easy to get wrong here, and it is preserved: the old `metadata.copy()` went through `setValues`, which drops `null` and zero-length arrays, and `getValues` returns `null` for a zero-length array. The new `values.length > 0` / `entry.getValue().length == 0` guards match that. Value arrays were shared by reference before too, so the PR description is accurate on that point.

My concern is the caching, not the single pass. See the inline comments on lines 230 and 238.

core/src/main/java/org/apache/stormcrawler/util/MetadataTransfer.java
GGraziadei: Fixed in a9578fd. `CompiledFilter` now keeps a `HashSet` snapshot of the keys it was built from and `isFor` is `snapshot.equals(filter)`, so a same-size content change (`remove("depth"); add("mycustom")`) rebuilds the compiled form. Javadoc adjusted to describe what the code does. Test `testSameSizeMutationOfTransferSetIsHonoured` reproduces the exact scenario and failed on the previous revision.

core/src/main/java/org/apache/stormcrawler/util/MetadataTransfer.java
GGraziadei: I went with the second option (sound check on a copy of the contents) rather than compiling at the end of `configure()`, for one reason: `configure()` is protected and a subclass that does `super.configure(conf); mdToTransfer.add("added.*");` is the natural extension point given the two `protected` sets. Compiling inside the base `configure()` would leave that subclass with a stale filter. `testSubclassCanExtendTransferSetInConfigure` pins that case.

The cache is now built lazily on first use and `isFor` is a `Set.equals` against the snapshot: a size check plus one hash lookup per key with cached `String` hashes, no allocation. The per-outlink win (the stream + intermediate `Set`) is unchanged.

Dropped: the two `volatile`s (all `CompiledFilter` fields are final, and the cache is idempotent), the identity dispatch and the `null` branch. `filter(Metadata, Set)` is now `filter(Metadata, CompiledFilter)` fed by `transferFilter()` / `persistOnlyFilter()`.

core/src/main/java/org/apache/stormcrawler/util/MetadataTransfer.java
GGraziadei: Gone. The private filter now takes a `CompiledFilter` directly, obtained from `transferFilter()` / `persistOnlyFilter()`, so there is no dispatch on set identity anymore.

core/src/main/java/org/apache/stormcrawler/util/MetadataTransfer.java
GGraziadei: Fixed. The wildcard loop reads `entry.getValue()` once and skips `null` as well as empty arrays, matching the old `getValues()` behaviour. `testNullValueArrayIsSkipped` builds a `Metadata` over a map with `null` arrays for both an exact key and a wildcard match; it threw an NPE on the previous revision.

core/src/test/java/org/apache/stormcrawler/util/MetadataTransferTest.java
GGraziadei: Added three tests: `testSameSizeMutationOfTransferSetIsHonoured` (remove/add keeping the size, second call must see the new keys; failed before), `testSubclassCanExtendTransferSetInConfigure` (subclass adds a wildcard after `super.configure()`), and `testNullValueArrayIsSkipped`. Full `core` verify: 447 tests, 0 failures.

dpol1 (approved):
LGTM. Compiling once on first use would still cover the subclass-in-`configure()` case and drop the `snapshot.equals` per outlink, which is ~75 ns on an 8-key set. Do we need the sets editable after that?

rzo1 (approved):
lgtm - think it is sufficient now and we can follow up later if needed ;-)

Thanks for the improvement!!

PR #2135, opened 2026-09-06
#2134 fetcher.thread.timeout: cancel the okhttp call, bounded helper pool for other protocols

### LINKED ISSUE #2134: fetcher.thread.timeout: cancel the okhttp call instead of interrupting a helper thread
`fetcher.thread.timeout` (#996, #1861) wraps `protocol.getProtocolOutput()` in a `Future` run on a single-thread executor owned by each `FetcherThread`, and cancels it with `future.cancel(true)` when the deadline passes.

Two problems with that:

- `cancel(true)` is a `Thread.interrupt()`, which okhttp does not honour while connecting or reading. The helper thread stays blocked on the socket until okhttp's own timeouts fire, and the `FetcherThread` queues behind it at its next fetch because the executor is single-threaded. One host that dribbles bytes can take a `FetcherThread` out of service for as long as `http.timeout`, or `topology.message.timeout.secs` (300s by default) which is the value the okhttp `callTimeout` is currently set from.
- It doubles the thread count of the bolt whenever the option is on: 50 `FetcherThread`s plus 50 `FetcherTimeout` helpers.

okhttp exposes the right primitive: `Call.timeout()` sets a per-call deadline enforced by okio's single shared watchdog thread, and on expiry the call is cancelled, the socket closed, and the fetching thread gets an `InterruptedIOException` immediately.

Proposal: `okhttp.HttpProtocol` applies `fetcher.thread.timeout` as a per-call deadline, and `Protocol` gains a `default boolean supportsFetchTimeout()` (false by default, true for okhttp when configured). `FetcherBolt` and `SimpleFetcherBolt` keep the `Future` path only for protocols that do not support it, creating the helper lazily. For the default okhttp protocol the fetch runs in the `FetcherThread` itself, the timeout is a real cancellation, and no helper threads exist.

GGraziadei:
Design doc available here https://docs.google.com/document/d/1XXwdz8-FQaRurvSpljVUdE3IuRmj-lG6/edit?usp=sharing&ouid=114888462700939773114&rtpof=true&sd=true 

Fixes #2134.

`fetcher.thread.timeout` ran the protocol call on a single-thread executor owned by each `FetcherThread` and abandoned it with `future.cancel(true)` when the deadline passed. okhttp does not honour the interrupt while connecting or reading, so the helper stayed blocked on the socket and the `FetcherThread` queued behind it at its next fetch: one host dribbling bytes could take a thread out of service for as long as `http.timeout`. The option also doubled the bolt's thread count, and the robots.txt lookup was not covered by the timeout at all.

### Change

**okhttp protocol**

- `fetcher.thread.timeout` is applied as a per-call deadline with `Call.timeout()`, enforced by okio's shared watchdog thread: on expiry the call is cancelled, the socket closed and the fetching thread gets an `InterruptedIOException` immediately, classified as "Socket timeout fetching" as before.
- The deadline is clamped to `topology.message.timeout.secs`, so that it can never loosen the client-level `callTimeout` derived from it (a warning is logged).
- With `http.content.partial.as.trimmed` the content received before the deadline is kept and flagged as trimmed for "time", which is what that option already did for the call timeout. Documented.
- `Protocol` gains `default boolean supportsFetchTimeout()` (false), backward compatible for external and user protocols. okhttp returns true when the deadline is configured; `DelegatorProtocol` only when every delegate does, since the bolt cannot know in advance which delegate a URL is routed to.

**Fetcher bolts**

- New package-private `FetchTimeoutHelpers`, owned by `FetcherBolt` and `SimpleFetcherBolt`. Its `call()` runs both the robots.txt lookup and the fetch: on the calling thread when the protocol enforces the timeout itself, otherwise on a helper thread from one bounded pool per bolt (`fetcher.thread.timeout.helpers`, default 2 × `fetcher.threads.number`, 2 for `SimpleFetcherBolt`; threads created on demand, released after a minute idle).
- A deadline on the helper path throws a typed `TimeoutException`; a full pool rejects at once with `SaturatedException`, reported as `FETCH_ERROR` with `fetch.exception` "No fetch helper available". Helpers are shared by all hosts, so a host that never answers can make fetches of other hosts fail this way until its helpers time out; stated in the docs.
- With the default okhttp protocol no helper thread is ever created.
- New `fetchhelpers` gauge and `fetch.timeout` / `fetch.helper.rejected` counters.
- `fetcher.thread.timeout` is now defined in `Constants`; the public alias `FetcherBolt.FETCH_TIMEOUT_PARAM_KEY` (#1861) is removed. The configuration key is unchanged.
- `configuration.adoc` documents `fetcher.thread.timeout`, which was missing, and the new helpers key.

### Behaviour

| protocol | `fetcher.thread.timeout` set | before | after |
|---|---|---|---|
| okhttp (default) | yes | fetch abandoned on a per-thread helper that stays blocked; 50 extra threads | call cancelled, socket closed, no helper threads |
| okhttp | no | unchanged | unchanged |
| other (Playwright, custom) | yes | per-thread helper; a stuck helper pinned its `FetcherThread` | bounded shared pool; stuck helpers do not pin threads; explicit saturation error |
| any | yes | robots.txt lookup outside the timeout | covered |

Exception classification now also maps `InterruptedIOException` (okhttp's own timeouts) to "Socket timeout fetching" regardless of the option; previously such failures carried the exception class name in `fetch.exception`.

### Tests

- `HttpProtocolFetchTimeoutTest`: a 5s response is cancelled at the 1s deadline; a 60s deadline is clamped to a 1s message timeout; partial content kept or failed depending on `http.content.partial.as.trimmed`.
- `FetchTimeoutHelpersTest`: every branch of `call()` (off, self-timing protocol, helper thread, exception propagation, typed timeout with interrupt, saturation, bound from config, shutdown).
- Fetcher bolt tests, run for both bolts, with a test protocol that hangs and ignores interruption: a stuck fetch followed by two fast ones on one fetcher thread yields two pages and one `FETCH_ERROR` (before: all three timed out); bounded pool with one rejection; hanging robots.txt reported at the timeout (before: the thread blocked for good).
- `FetcherBoltTest`: no helper threads with okhttp; slow robots.txt bounded by the deadline with okhttp.
- The existing `testThreadTimeout` passes unchanged through the okhttp path.

---

### For all changes

### For code changes


rzo1 (commented):
Thanks for picking this up. A few things from reading the diff:

**robots.txt timeout behaves differently per protocol**

With okhttp the deadline fires inside `http.getProtocolOutput()`, and `HttpRobotRulesParser` catches that (the `catch (Throwable t)` around line 218). It returns `EMPTY_RULES` and puts them in the error cache, so the page gets fetched as if there were no robots.txt. On the helper path the same timeout ends up as FETCH_ERROR for the URL. The okhttp behaviour isn't new, but the yaml comment and the PR description read as if both paths handle a hanging robots.txt the same way. I'd make them consistent.

**The deadline is per call, not per fetch, with okhttp**

The robots lookup can follow up to 5 redirects and each one gets the full timeout, then the page fetch gets another one. Worst case that's around 7x `fetcher.thread.timeout` for one URL. Fine by me if it's intended, but the docs should say so, because "hard timeout for a single fetch" suggests otherwise.

**`DelegatorProtocol.supportsFetchTimeout()`**

If any delegate returns false, every URL goes to the helper pool, including the ones routed to okhttp. The delegate is chosen per URL anyway, so could we resolve it first, e.g. `supportsFetchTimeout(url, metadata)`, or check `getProtocolFor(url, metadata)` in the bolt?

**`fetch.timeout` counter**

It counts any `InterruptedIOException`, and `SocketTimeoutException` is a subclass, so normal `http.timeout` socket timeouts are counted too. That makes it hard to see how often the new deadline actually fires. Could the deadline get its own counter?

**Clamping**

The clamp to `topology.message.timeout.secs` is only in `HttpProtocol`. Doing it in `FetchTimeoutHelpers` would cover both paths.

**Nit**

`StuckProtocol.HANG_ROBOTS` is static mutable state shared between tests. OK while tests run serially, but it will break with parallel surefire.

dpol1 (commented):
Two small things on top of rzo1's points.

*Findings flagged by Claude*

core/src/main/java/org/apache/stormcrawler/Constants.java
```diff
@@ -60,4 +60,13 @@ public class Constants {
     public static final int MAX_ARRAY_SIZE = Integer.MAX_VALUE - 8;
 
     private Constants() {}
+
+    /** Hard deadline in seconds for a single fetch, -1 to disable. */
+    public static final String FETCH_TIMEOUT_PARAM_KEY = "fetcher.thread.timeout";
```
dpol1, line 65: `FetcherBolt.FETCH_TIMEOUT_PARAM_KEY` shipped in 3.6.0 and 3.7.0, so code that used it stops compiling. Keep a deprecated alias, or note it for 4.0?

core/src/main/java/org/apache/stormcrawler/bolt/FetcherBolt.java
```diff
@@ -1002,10 +954,19 @@ public void run() {
...
                         LOG.info("Socket timeout fetching {}", fit.url);
                         message = "Socket timeout fetching";
+                        eventCounter.scope("fetch.timeout").incrBy(1);
+                    } else if (exece instanceof FetchTimeoutHelpers.SaturatedException) {
+                        eventCounter.scope("fetch.helper.rejected").incrBy(1);
+                        LOG.warn(
+                                "{}: all {} fetch helpers are busy",
+                                message,
+                                fetchHelpers.maxHelpers());
+                        message = "No fetch helper available";
```
dpol1: This goes out as FETCH_ERROR, so a URL that never reached the network still takes a strike towards max.fetch.errors. Would the queue-timeout treatment above (ack, no status, spout retries) fit better?

core/src/main/java/org/apache/stormcrawler/Constants.java
GGraziadei: Kept `FetcherBolt.FETCH_TIMEOUT_PARAM_KEY` as a deprecated alias of `Constants.FETCH_TIMEOUT_PARAM_KEY`, marked for removal, so existing code keeps compiling.

dpol1: Grazie Gianluca, the alias does it. Only since should read 4.0, 3.8 will not exist. Necessary since this is a public API

core/src/main/java/org/apache/stormcrawler/bolt/FetcherBolt.java
```diff
@@ -1017,17 +993,33 @@ public void run() {
...
+                    // the URL never reached the network: like a URL which waited too long in
+                    // the queue, it is acked without a status so that the spout retries it
+                    // later, rather than taking a strike towards max.fetch.errors
+                    eventCounter.scope("fetch.helper.rejected").incrBy(1);
+                    LOG.warn(
+                            "[Fetcher #{}] {}: all {} fetch helpers are busy",
+                            taskId,
+                            e.getMessage(),
+                            fetchHelpers.maxHelpers());
+                    asap = true;
```
dpol1, line 1006: With asap = true a queue with more URLs is eligible again immediately while the helper pool is still full, so it collects repeated rejections and WARNs. Could saturation get some backoff?

core/src/main/java/org/apache/stormcrawler/protocol/okhttp/HttpProtocol.java
```diff
@@ -731,6 +748,16 @@ public ProtocolResponse getProtocolOutput(String url, final Metadata metadata)
                 }
                 call = fetchClient.newCall(currentRequest);
                 hopCalls.add(call);
+                if (deadlineNanos != 0) {
+                    // hard deadline for the whole chain, enforced by okio's watchdog: on
```
dpol1, line 752: DNS can still hold the fetcher thread past this deadline: cancellation cannot release the synchronous lookup until the resolver returns. Could we say so next to the hard-deadline description?

dpol1 (commented):
Second round looks good; two questions inline.

*Findings flagged by Claude*

rzo1 (changes requested):
Thanks, the cancellation approach looks good. Two things:

- **Pool rejects while a helper is idle** (`FetchTimeoutHelpers`): with `ThreadPoolExecutor(0, max, SynchronousQueue)`, `future.get()` returns before the worker is back in `poll()`. The page fetch submitted right after the robots lookup can then get `SaturatedException` when the pool is at max. A standalone repro with max=2 and one stuck task rejected ~1% of URLs, and `abandonedFetchesUseABoundedSharedPool` fails intermittently because of it. A `Semaphore` (`tryAcquire` before submit, release in `finally`) in front of a queueing executor would avoid this.
- **Timed-out robots lookups aren't cached** (`FetcherBolt:734`, `SimpleFetcherBolt:302`): the helper path only uses `EMPTY_RULES` locally. Every later URL for the host starts another robots.txt request on a new helper and waits a full deadline. Could the timeout result go into the error cache like other failed lookups?

core/src/main/java/org/apache/stormcrawler/Constants.java
GGraziadei: Thanks Davide. I am fixing it :-) ;

core/src/main/java/org/apache/stormcrawler/bolt/FetcherBolt.java
GGraziadei: Good catch, thanks! I am checking something like exponential backoff with equal jitter implementation.

core/src/main/java/org/apache/stormcrawler/protocol/okhttp/HttpProtocol.java
GGraziadei: Hi Davide, thanks. I have just updated the comment.

GGraziadei:
Hi @rzo1, @dpol1 
I have just fixed the PR according to your comments.

rzo1 (commented):
Thanks, the semaphore does it. I ran the fetcher bolt tests four times here and `abandonedFetchesUseABoundedSharedPool` was stable, and the robots caching on the helper path reads right. Four things left.

**The robots error cache shadows rules which did arrive**

`getRobotRulesSet` looks at `ERRORCACHE` before `CACHE`, and the lookup abandoned at the deadline keeps running on its helper and writes the real rules to `CACHE` when it finishes. The guard in `cacheLookupFailure` only covers the other order. So one timed-out lookup makes the host allow all for the error cache TTL (1h by default) although its robots.txt was fetched a moment later. With a robots.txt served after 3s and carrying `Disallow: /restricted/`:

```
rules obtained on the helper:             allowAll=false, /restricted/ allowed=false
rules served to the next URL of the host: allowAll=true,  /restricted/ allowed=true
```

Until now both caches were written by the same lookup and could not disagree. Invalidating the error cache entry where the rules are put in `CACHE` would be enough.

**The permit can be released twice** (`FetchTimeoutHelpers`)

`started.set(true)` is the first statement of the submitted task, but `FutureTask.run()` tests `state == NEW` before that. If `cancel(true)` falls between the two, the caller sees `started == false` and releases, then the helper runs the call and releases again in its `finally`. Every occurrence leaves an extra permit for good, so the bound erodes over the life of the bolt and `busy()` can go negative. One atomic handoff instead of the two checks would do it, e.g. a state `compareAndSet(PENDING, RUNNING)` in the task against `compareAndSet(PENDING, ABANDONED)` in the caller, releasing only if it wins.

**Deprecation**

`@Deprecated(since = "3.8")` in `FetcherBolt` still reads 3.8, see Davide's comment above: it should be 4.0.

**Nit**

`StuckProtocol` is back to static mutable state with `ROBOTS_HUNG` and `ROBOTS_TIMED_OUT`, just after `HANG_ROBOTS` became per instance.

core/src/main/java/org/apache/stormcrawler/Constants.java
GGraziadei: Hi @dpol1 I have just fixed in [9b23ff7](https://github.com/apache/stormcrawler/pull/2135/commits/9b23ff78499d2e249200f8507d1ff99a9b654009)

GGraziadei:
Hi @rzo1, 
Thanks for your comment. 

I fixed in: 
- https://github.com/apache/stormcrawler/pull/2135/commits/b16ce2390324de62f9f88e5c26b5731e56103c90 access to HTTPRobot Cache for the sequence FetcherBolts puts EMPTY the Helper thread ends after
- https://github.com/apache/stormcrawler/pull/2135/commits/92a127d74e13cc2934e669c6b125c1cf8d5abf2e managed the `private final AtomicInteger` variables in `StuckProtocol` including STARTED in addition to your previous comment
- https://github.com/apache/stormcrawler/pull/2135/commits/9b23ff78499d2e249200f8507d1ff99a9b654009 managed the deprecation message with the right version for FETCH_TIMEOUT_PARAM_KEY
- https://github.com/apache/stormcrawler/pull/2135/commits/7a2dbfaf1d4cb187a6e6f52fe373e8d5175767ab handled the race condition in FetchTimeoutHelpers. Added an Handoff based on AtomicInteger.

I conducted an additional check on the management of ERRORCACHE in 
[a575b47](https://github.com/apache/stormcrawler/pull/2135/commits/a575b472989a2ebd3948132cfd56ac3c100e8b22)

rzo1 (approved):
Thanks @GGraziadei. I pushed a small commit on top fixing a few things (saturation backoff with a negative `fetcher.max.crawl.delay`, error cache vs. a 5xx result from the helper, restored `permitsAreNeverReleasedTwice`). I think it is fine now.

GGraziadei:
Hi @rzo1 thanks !

dpol1 (approved):
Thank you @GGraziadei  - this is a huge improvement and a sharp bug fix, will merge shortly!

Issue #2162, opened 2026-09-15
Scope credential headers in http.custom.headers to hosts


rzo1:
Follow-up to #2089.

The PR for #2089 limits `http.basicauth.*` credentials to the hosts listed in `http.basicauth.hosts`. Credential headers configured through `http.custom.headers` (`Authorization`, `Cookie`, `X-Api-Key`, anything in `http.credentials.headers`) are still sent to every host; only a startup WARN was added.

Scoping them needs its own design, e.g. a per-host form of `http.custom.headers`, and is a separate breaking change. Today the only per-site alternative is `protocol.set-header` metadata, which is awkward for seed-based crawls.

PR #2163, opened 2026-09-15
#2089 Send http.basicauth credentials only to the hosts in http.basicauth.hosts

rzo1:
Part of #2089. Scoping credential headers in `http.custom.headers` is tracked in #2162.

Adds `http.basicauth.hosts`. The basic auth header is sent only to listed hosts (page fetches and robots.txt); credential headers in `http.custom.headers` now log a WARN.

**Release note (breaking):** `http.basicauth.user` without `http.basicauth.hosts` no longer sends credentials; a WARN is logged at startup.


core/src/main/java/org/apache/stormcrawler/protocol/okhttp/HttpProtocol.java
```diff
@@ -499,6 +532,50 @@ private URL getCookieOrigin(Metadata md, String url) {
...
+                host.startsWith("[") ? StringUtils.substringAfter(host, "]") : host;
+        final HttpUrl parsed =
+                host.contains("*") || afterAddress.contains(":")
+                        ? null
+                        : HttpUrl.parse("http://" + host + "/");
+        if (parsed == null
+                || !parsed.equals(
+                        new HttpUrl.Builder().scheme("http").host(parsed.host()).build())) {
+            LOG.warn(
+                    "Ignoring '{}' in http.basicauth.hosts, expected a host name or IP address "
```
dpol1: could we not log the value here? a pasted URL with user:pw in it would end up in the logs, like the proxy string in #2151

dpol1 (approved):
LGTM, one question inline

PR #2165, opened 2026-09-15
#2095 Apply the IP address filter to the requests made by the Playwright browser

### LINKED ISSUE #2095: Playwright protocol applies neither the IP filter nor the URL filters to the requests it makes
## What happens
`IPFilterRules` is only used by the okhttp protocol, where it runs as a network interceptor. The playwright protocol never constructs it, so `http.filter.ipaddress.include` and `http.filter.ipaddress.exclude` have no effect on a topology that fetches with playwright. The module already installs a route handler that sees every request the browser makes, but that handler only aborts resource types listed in `playwright.skip.resource.types` and otherwise resumes. This covers the navigation itself, its redirect hops, and every subresource, XHR, iframe and script request the rendered page issues.

## Where
`external/playwright/src/main/java/org/apache/stormcrawler/protocol/playwright/HttpProtocol.java:243-256`.

```java
page.route( lambdaUrl -> true, route -> { // abort if we know the main page is a redirection if (status.get() != -1) { ... } else if (resourceTypesToSkip.contains(route.request().resourceType())) { route.abort(); } else { route.resume(); } });
```

The IP filter itself lives in `core/src/main/java/org/apache/stormcrawler/protocol/IPFilterRules.java` and is wired up only in `core/src/main/java/org/apache/stormcrawler/protocol/okhttp/HttpProtocol.java`.

## Why it matters
This is documented: `crawler-default.yaml:161-162` and `docs/src/main/asciidoc/configuration.adoc:259-260` both say the IP filter is okhttp only, and both keys ship commented out, so nobody was told something untrue. The gap is still worth closing. A crawler that renders pages in a browser makes far more requests than one that fetches bytes, and every one of them leaves the worker with no address check, including requests the fetched page decides to make. An operator who moves a topology from okhttp to playwright loses the control silently, because the keys stay in the configuration and stop doing anything.

## Reproduction
No automated test. `configure()` launches or connects to a real Chromium, so exercising the route handler needs a browser plus an HTTP server bound to a filtered address.

Manual steps:

1. Configure a topology with the playwright protocol and set `http.filter.ipaddress.exclude: "localhost,sitelocal"`.
2. Run a small HTTP server on the loopback interface and fetch its URL through the protocol. It is fetched, whereas the okhttp protocol fails the fetch with an `IOException`.
3. Serve a page whose body requests a subresource from a loopback URL. The request is issued by the browser and the route handler resumes it.

## Suggested fix
Build an `IPFilterRules` instance in `HttpProtocol.configure()` from the same configuration keys, and apply it inside the existing `page.route` handler: resolve the request host and call `route.abort()` when the address is rejected. Apply `URLFilters` in the same place so redirect hops and subresource requests are checked against the crawl scope. Update `docs/src/main/asciidoc/configuration.adoc` so the "OkHttp only" note reflects the new state. Resolution in the handler adds a lookup per request, so cache results per host. A separate DNS lookup is also not the address the browser ends up connecting to, which is worth stating in the documentation.

rzo1:
Fixes #2095.

The Playwright protocol now checks the requests of the page, its subresources and popups against `http.filter.ipaddress.*` and aborts the rejected ones. The check sits on a context route, so popups are covered as well, and service workers are blocked while the filter is active since their requests cannot be intercepted. Not applied through `http.proxy`, same as OkHttp.

Not checked: redirect hops (only the first URL of a chain goes through the route handler) and WebSockets. The address is also resolved separately from the browser, so short-lived DNS records or a remote browser with another resolver can end up on a different address. These limits are documented in `crawler-default.yaml` and `configuration.adoc`.

A blocked page fails the fetch, a blocked subresource is missing from the rendered page.

URL filters are not applied here: main page redirects are already aborted and go through the regular pipeline, and crawl-scope filters on subresources would break rendering.

**Release note:** with the default exclude list, Playwright topologies crawling loopback or private addresses need to override `http.filter.ipaddress.exclude`.


external/playwright/src/main/java/org/apache/stormcrawler/protocol/playwright/HttpProtocol.java
```diff
@@ -188,6 +172,111 @@ public void configure(final Config conf) {
...
+     *
+     * @return false if any address of the host is rejected or the host cannot be resolved
+     */
+    boolean isAllowedAddress(final String url) {
+        if (ipFilterRules == null) {
+            return true;
+        }
+        final URI uri;
+        try {
+            uri = new URI(url);
```
dpol1: `new URI` fails on `|`, `[`, `]` (and `{ } ^` in a query), which Chromium sends as is, and gives no host for a name with `_`: with the default exclude list those get aborted as forbidden. `HttpUrl.parse` takes them all, and a test with one of them?

core/src/main/resources/crawler-default.yaml
```diff
@@ -220,6 +220,9 @@ config:
   # Fetches through a proxy are not filtered: the proxy resolves and connects
   # to the target, so only its address is known here and a proxy on a private
   # address is fine. The proxy's own egress rules decide what it may reach.
+  # Applies to the okhttp and playwright protocols. Playwright checks each
+  # request of the browser (page, redirects, subresources) against an address
```
dpol1: redirect hops never go through the handler, only the first url does (the note above `page.route` says it, and a 302 on an image and on the main page confirmed it), so this should list redirects with websockets as not checked. same block still says okhttp only, and the adoc line on failing with an `IOException` doesn't hold for a blocked subresource.

external/playwright/src/main/java/org/apache/stormcrawler/protocol/playwright/HttpProtocol.java
```diff
@@ -251,6 +340,11 @@ public ProtocolResponse getProtocolOutput(String url, Metadata md) throws Except
                                 } else if (resourceTypesToSkip.contains(
                                         route.request().resourceType())) {
                                     route.abort();
+                                } else if (!isAllowedAddress(route.request().url())) {
```
dpol1: the handler also misses what a service worker fetches (playwright suggests `setServiceWorkers(BLOCK)` when routing) and popups, which open unblocked with no handler at all unless the route sits on the context. block them or document them, your call

dpol1 (changes requested):
the URI parsing breaks with the default exclude list and the docs promise more than the handler sees, see inline

external/playwright/src/main/java/org/apache/stormcrawler/protocol/playwright/HttpProtocol.java
rzo1: Switched to `HttpUrl.parse`, test added with `|[]{}^`.

core/src/main/resources/crawler-default.yaml
rzo1: Right, fixed: the yaml and adoc now list redirect hops and WebSockets as not checked, the "okhttp only" header is gone, and the adoc says what happens on each side (IOException for OkHttp, aborted request for Playwright).

external/playwright/src/main/java/org/apache/stormcrawler/protocol/playwright/HttpProtocol.java
rzo1: Blocked both: the IP check moved to a context route (page route falls back to it), so popups are covered, and service workers are blocked while the filter is active. Documented as well.

dpol1 (approved):
LGTM, all three points addressed. Two things before merging: a rebase on main now that #2158 is in, and the description, which still says redirects are checked and would become the squash message.

PR #2167, opened 2026-09-18
#2101 HostURLFilter: drop the source host memoization shared by all fetcher threads

rzo1:
`HostURLFilter` kept the last source URL, host and domain in three plain instance fields and compared by identity. `StatusEmitterBolt` builds one `URLFilters` per bolt and `FetcherBolt` runs `fetcher.threads.number` threads through it, so those fields were written and read concurrently with no synchronisation, and the three writes were not atomic with each other. A thread could see its own URL in `previousSourceUrl` next to another thread's host in `previousSourceHost`, and compare the candidate against the wrong origin — dropping a same-host URL or admitting one outside the source host.

The cache only saved one `PaidLevelDomain.getPLD` call, so it is removed rather than synchronised.


dpol1 (approved):
LGTM.

PR #2168, opened 2026-09-18
#2090 Fetcher bolts: drop a set-header value coming from a response

rzo1:
The protocols read the headers to add to an outgoing request from `<protocol.md.prefix>set-header`. The fetcher bolts merge every response header into the tuple metadata under that same prefix, lowercased, so a server returning a header named `set-header` writes exactly that key. Where the operator persists or transfers `protocol.` keys — internals.adoc recommends it for `protocol.etag` and `protocol.set-cookie`, and the lists accept wildcards — the forged value is carried into the next fetch, and the crawled site chooses headers sent afterwards, including on requests to other hosts.

The merge now drops the key the way it already drops `robots.crawl.delay`, then puts back the value configured for the URL. `SET_HEADER_BY_REQUEST` becomes public so the bolts can name it.


dpol1 (approved):
LGTM - needs rebase

Issue #2166, opened 2026-09-18
[Improvement] Add a cap on the number of outlinks emitted per sitemap


rzo1:
## What is missing

`SiteMapParserBolt` emits one DISCOVERED tuple per URL found in a sitemap, at `core/src/main/java/org/apache/stormcrawler/bolt/SiteMapParserBolt.java:203-206`, with no upper bound. `JSoupParserBolt` has had `parser.emitOutlinks.max.per.page` for the equivalent case since a while, read at `JSoupParserBolt.java:169` and applied at `:604`, but there is no counterpart for sitemaps.

## Why it matters

A single sitemap can enrol an arbitrary number of URLs in one tuple. crawler-commons applies its `MAX_URLS = 50000` only to plain-text sitemaps (`SiteMapParser.java:528`); the XML path has no cap at all, which is crawler-commons/crawler-commons#611. Until that is settled upstream, a gzipped sitemap is a cheap way for a site to push tens of thousands of URLs into the status stream from one fetch.

A cap here is only a partial mitigation and should be described as one: the parse has already allocated the `SiteMapURL` objects by the time the bolt emits, so this bounds what reaches the downstream queues and the status index, not the parser bolt's own heap. That is still the part an operator can act on.

## Suggested shape

A `sitemap.emitOutlinks.max` key, default -1 (no limit) to keep current behaviour, applied in the emit loop and logged when it trims, following what `JSoupParserBolt` already does. Worth deciding whether the truncated sitemap is still marked FETCHED, which is what happens today.

Related: #2100 for the unbounded expansion itself.

PR #2169, opened 2026-09-18
#2100 SiteMapParserBolt: read the content type under the protocol prefix

rzo1:
The fetcher stores response headers prefixed with `protocol.md.prefix`, so reading `Content-Type` without it never found a value in a shipped topology. The declared type was ignored and every sitemap was identified by guessing. It also left the media type check in the content sniffing dead: a page served as `text/html` carrying the sitemaps namespace was promoted to a sitemap.

Reads the key with the configured prefix, as `JSoupParserBolt` does.

The other half of #2100 is not addressed here. `processGzippedXML` streams into SAX, so there is no decompressed buffer to bound — what grows is the accumulated `SiteMapURL` set, and neither `MAX_BYTES_ALLOWED` nor `MAX_URLS` is applied on the XML path. Raised upstream as crawler-commons/crawler-commons#611, with #2166 for a cap on our side in the meantime.


core/src/main/java/org/apache/stormcrawler/bolt/SiteMapParserBolt.java
```diff
@@ -111,13 +112,16 @@ public class SiteMapParserBolt extends StatusEmitterBolt {
...
+    private String protocolMetadataPrefix = "";
+
     @Override
     public void execute(Tuple tuple) {
         Metadata metadata = (Metadata) tuple.getValueByField("metadata");
         byte[] content = tuple.getBinaryByField("content");
         String url = tuple.getStringByField("url");
 
-        String ct = metadata.getFirstValue(HttpHeaders.CONTENT_TYPE);
+        String ct = metadata.getFirstValue(HttpHeaders.CONTENT_TYPE, protocolMetadataPrefix);
```
dpol1, line 124: the type now also reaches crawler-commons, which trusts a declared `application/xml` instead of detecting it: a gzipped sitemap served as xml with `Content-Encoding: x-gzip` (okhttp only decodes `gzip`) parsed before and fails now. keep passing null here and use the header for the sniff check only?

dpol1 (changes requested):
the prefix fix is right, one regression inline

core/src/main/java/org/apache/stormcrawler/bolt/SiteMapParserBolt.java
rzo1: Good catch. crawler-commons gets null again and detects the format itself, the header is only used for the sniff check. Added a test with a gzipped sitemap declared as `application/xml` with `Content-Encoding: x-gzip`.

dpol1 (approved):

PR #2170, opened 2026-09-18
#2099 Fetcher bolts: skip a metrics.* value which is not a number

rzo1:
The protocols copy every response header into the response metadata, so a header named `metrics.something` lands in the namespace the bolts read their protocol timings from. `Long.parseLong` was called unguarded, so a non-numeric value threw inside the fetch block and a response that had arrived intact was recorded as `FETCH_ERROR`, with a bare `NumberFormatException` as the reason.

The parsing moves into a helper shared by both bolts, which skips a value it cannot parse and logs it.

This is the bolt-side half. Stopping the server from creating those keys at all is in #2091, which strips `metrics.*` in the protocols; between the two, #2099 can close. Once both are in, `ProtocolMetrics` should use `ProtocolResponse.METRICS_PREFIX` rather than its own copy of the string.


core/src/test/java/org/apache/stormcrawler/bolt/AbstractFetcherBoltTest.java
```diff
@@ -192,6 +193,30 @@ void invalidProxyMetadataEmitsFetchError(WireMockRuntimeInfo wmRuntimeInfo)
         Assertions.assertEquals(0, output.getEmitted(Utils.DEFAULT_STREAM_ID).size());
     }
 
+    /**
+     * The protocols copy every response header into the response metadata, so a header named
+     * metrics.something reaches the loop which feeds the fetcher metric registry. A value which is
+     * not a number must not turn a complete fetch into a FETCH_ERROR, see issue #2099.
+     */
+    @Test
+    void nonNumericMetricsHeaderDoesNotFailTheFetch(WireMockRuntimeInfo wmRuntimeInfo)
```
dpol1, line 202: once #2171 drops `metrics.*` at the protocol, this passes without the catch in `ProtocolMetrics`. a direct test of `update` with a valid and an invalid value would keep the guard covered?

dpol1 (approved):
LGTM, one question on the test

core/src/test/java/org/apache/stormcrawler/bolt/AbstractFetcherBoltTest.java
rzo1: Added `ProtocolMetricsTest`, which calls `update` directly with a valid and an invalid value, so the guard stays covered after #2171.

PR #2171, opened 2026-09-18
#2091 Protocols: reject metadata keys arriving in a response header

rzo1:
The protocols copy every response header into the response metadata, and a header name is not restricted to anything, so a fetched server could write the keys the crawler uses to record its own fetch. `WARCRecordFormat` promotes a record from `resource` to `response` whenever `_response.headers_` is non-blank and embeds the block verbatim, and `WARCRequestRecordFormat` writes `_request.headers_` as what the crawler supposedly sent — so the archive could be made to contain a request and response of the server's choosing, with its own `WARC-IP-Address` and `WARC-Date`. Values under `metrics.*` reached the fetcher metric registry the same way, which is the other half of #2099.

`ProtocolResponse.isReservedMetadataKey` names the keys the crawler writes itself, and the okhttp and Playwright protocols skip a header which matches.

Two details worth a look in review:

- In okhttp the keys `HTTPHeadersInterceptor` passes through the response headers stay readable, but only while that interceptor is installed, which is what `http.store.headers` does. Without that condition the change would break `http.store.headers: true`.
- The interceptor sets the cipher suite only when there is a handshake, so over `http://` a value from the server survived even with headers stored. It now drops that header in the else branch.

This also removes an unguarded base64 decode: a value under `_request.headers_` which was not valid base64 threw `IllegalArgumentException` out of `getProtocolOutput` and failed the fetch.


core/src/main/java/org/apache/stormcrawler/protocol/ProtocolResponse.java
```diff
@@ -78,6 +82,36 @@ public class ProtocolResponse {
...
+            Stream.of(
+                            REQUEST_HEADERS_KEY,
+                            RESPONSE_HEADERS_KEY,
+                            RESPONSE_IP_KEY,
+                            REQUEST_TIME_KEY,
+                            PROTOCOL_VERSIONS_KEY,
+                            CIPHER_SUITE_KEY,
+                            TRIMMED_RESPONSE_KEY,
+                            TRIMMED_RESPONSE_REASON_KEY,
+                            REDIRECTED_TO_KEY)
```
dpol1, line 99: `playwright.protocol.end` is another key a protocol writes and the js rendering guards read through `skipIfMetadataPresent`: with an empty `protocol.md.prefix`, or the key configured with it, a server can set it through okhttp. worth reserving too?

dpol1 (approved):
LGTM, one question inline

core/src/main/java/org/apache/stormcrawler/protocol/ProtocolResponse.java
rzo1: Yes, reserved `playwright.protocol.start` and `playwright.protocol.end` as well, with a test.

PR #2174, opened 2026-09-19
#2097 Tika ParserBolt: limit the extracted text with parser.tika.text.maxlength

rzo1:
Part of #2097.

Adds `parser.tika.text.maxlength`. Default -1 (no limit), and any negative value means no limit. When the limit is reached, the document is emitted with the text extracted so far and `parse.text.trimmed=true` instead of failing, and the trim is counted under `text_trimmed`.

The parse timeout from the issue is left for a follow-up.


jnioche (approved):

jnioche:
thanks @rzo1

PR #2175, opened 2026-09-19
#2098 URLFrontier: optional TLS for the channels to the frontier

### LINKED ISSUE #2098: URLFrontier channels are always plaintext and the module has no TLS option
## What happens
`ManagedChannelUtil.createChannel()` builds every gRPC channel with `usePlaintext()`. It is the only channel factory in the module: `Spout`, `StatusUpdaterBolt` and `QueueRegulatorBolt` all go through it. There is no configuration key for TLS, mutual TLS or channel credentials anywhere in `external/urlfrontier`, so an operator who wants the link encrypted cannot get it, even deliberately.

## Where
`external/urlfrontier/src/main/java/org/apache/stormcrawler/urlfrontier/ManagedChannelUtil.java:46-53`. Related config keys: `urlfrontier.address`, `urlfrontier.host`, `urlfrontier.port`.

```java
static ManagedChannel createChannel(@NotNull String address) {
    ...
    return ManagedChannelBuilder.forTarget(address).usePlaintext().build();
}
```

## Why it matters
URLFrontier is typically a separate service, and the module supports several `urlfrontier.address` entries for exactly that layout, so the channel usually crosses a host boundary. Everything the crawler knows about a URL travels on it: the URL itself and its metadata, which can include cookies when cookie support is enabled. Anyone who can read that segment reads the crawl state, and anyone who can write to it can change what the workers fetch, because per-URL metadata influences request behaviour. Operators who need the link protected today have to tunnel it themselves.

## Reproduction
No automated test. Demonstrating the transport would need a running frontier service and a second gRPC endpoint with TLS, which is not something to put in a unit test. Manual steps:

1. `grep -rn "usePlaintext\|TlsChannelCredentials" external/urlfrontier/src/main` returns only line 52 of `ManagedChannelUtil.java` and no TLS builder.
2. `grep -rn "urlfrontier\." external/urlfrontier/src/main/java/org/apache/stormcrawler/urlfrontier/Constants.java` lists every configuration key the module reads. None of them concerns transport security.
3. Start a URLFrontier service configured to require TLS and point a topology at it. The channel fails to connect, and no setting changes that.

## Suggested fix
Add configuration to `ManagedChannelUtil.createChannel` for TLS channel credentials, built with `Grpc.newChannelBuilder(address, TlsChannelCredentials...)`, with optional client certificate and trust roots. Keep plaintext as the default for now so existing deployments keep working, and log a warning at channel creation when plaintext is used. Flipping the default to TLS is a breaking change and belongs in a major release, not in a patch.

rzo1:
Fixes #2098.

The gRPC channels built in `ManagedChannelUtil` (used by the Spout, StatusUpdaterBolt and QueueRegulatorBolt) can now use TLS. Plaintext stays the default.

- `urlfrontier.tls.enabled`: TLS on or off, default false
- `urlfrontier.tls.trust.cert.collection`: PEM file of trusted certificates; the JVM trust store is used if not set
- `urlfrontier.tls.client.cert.chain`, `urlfrontier.tls.client.private.key` and the optional `urlfrontier.tls.client.private.key.password`: client certificate for mutual TLS

A half-configured client certificate or an unreadable file fails the component at startup. No new dependencies. The tests cover TLS and mutual TLS round trips against a local gRPC server using certificates generated during the test.


jnioche:
@rzo1 is it even possible to run URLFrontier with TLS at the moment? if so, do you have any instructions on how to do so?

rzo1:
No, the frontier server doesn't do TLS at the moment. It only listens in plaintext, so today you'd need a TLS terminating proxy in front of it. I've opened crawler-commons/url-frontier#219 to add TLS to the server, the channels between nodes and the CLI client. Once that's released, the settings in this PR connect to it directly, without a proxy.

external/urlfrontier/README.md
```diff
@@ -24,6 +24,30 @@ urlfrontier.max.buckets: 10
 urlfrontier.max.urls.per.bucket:10
 ```
 
+## Transport security
+
+The gRPC channels to the frontier are plaintext unless TLS is enabled. Plaintext is kept as the
+default so that existing deployments keep working. The channel carries the URLs and their metadata, so enable TLS whenever the
```
dpol1, line 30: can you add what you said above? URLFrontier 2.6 needs a TLS proxy in front of it, until crawler-commons/url-frontier#219 is released

external/urlfrontier/README.md
```diff
@@ -24,6 +24,30 @@ urlfrontier.max.buckets: 10
...
+urlfrontier.tls.client.cert.chain: /etc/stormcrawler/crawler.pem
+urlfrontier.tls.client.private.key: /etc/stormcrawler/crawler-key.pem
+# only needed if the private key is encrypted
+urlfrontier.tls.client.private.key.password: changeit
+```
+
+The server certificate must be valid for the host name in `urlfrontier.address` or
+`urlfrontier.host`. Setting only one of `urlfrontier.tls.client.cert.chain` and
+`urlfrontier.tls.client.private.key`, or pointing a key at a file which cannot be read, fails
+the component at startup. The settings apply to `Spout`, `StatusUpdaterBolt` and
```
dpol1: a failed handshake doesn't fail the startup: `Spout` and `StatusUpdaterBolt` use `withWaitForReady()` without a deadline and just wait. worth a line here

dpol1 (approved):
LGTM. Two notes on the README, the first follows from the question above.

